PR #44, opened 2020-08-27
Update README.md

l123456789jy:
Modify the maven dependency error report,
Could not set unknown property'url' for repository container of type org.gradle.api.internal.artifacts.dsl.DefaultRepositoryHandler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,7 +256,9 @@ The library depends on the Kotlin Standard Library not lower than `1.4.0`.
 
 ```kotlin
 repositories {
-    maven(url = "https://kotlin.bintray.com/kotlinx/") // soon will be just jcenter()
+    maven {
+            url 'https://kotlin.bintray.com/kotlinx/'
+        }    // soon will be just jcenter()
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Jadarma:
The current version is correct, since it is written for Gradle with Kotlin DSL (notice the kotlin marker for the markdown code block).
The fix you submitted is for a Groovy script. It would not be a valid `build.gradle.kts` file.

l123456789jy:
> The current version is correct, since it is written for Gradle with Kotlin DSL (notice the kotlin marker for the markdown code block).
> The fix you submitted is for a Groovy script. It would not be a valid `build.gradle.kts` file.

But there is no place to indicate that it is KotlinKotlin DSL, it will be misunderstood early, should it be indicated?

nuhkoca:
It would be better if you implement collapsable blocks for Groovy and make Kotlin DSL expanded by default. You can refer to my [PR](https://github.com/JLLeitschuh/ktlint-gradle/pull/305 ) for [ktlint-gradle](https://github.com/JLLeitschuh/ktlint-gradle) repository to see how it better fits.

ilya-g:
Thanks for the PR, but we now the library artifacts can be found in `mavenCentral()`, which looks the same in Kotlin and in Groovy scripts, see https://github.com/Kotlin/kotlinx-datetime/blob/d3c63fb/README.md#gradle

PR #46, opened 2020-09-04
Common: add Instant.fromEpochSeconds(Long, Int)

dkhalanskyjb:
This overload allows to pass an Int as the number of nanoseconds.

### DIFF
--- a/core/jsMain/src/mathJs.kt
+++ b/core/jsMain/src/mathJs.kt
@@ -61,13 +61,13 @@ internal actual fun safeMultiply(a: Int, b: Int): Int {
 
 /**
  * Returns the floor division.
- * <p>
- * This returns {@code 0} for {@code floorDiv(0, 4)}.<br />
- * This returns {@code -1} for {@code floorDiv(-1, 4)}.<br />
- * This returns {@code -1} for {@code floorDiv(-2, 4)}.<br />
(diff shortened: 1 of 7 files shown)


core/jsMain/src/mathJs.kt
```diff
@@ -58,3 +58,35 @@ internal actual fun safeMultiply(a: Int, b: Int): Int {
...
+
+/**
+ * Returns the floor division.
+ * <p>
+ * This returns {@code 0} for {@code floorDiv(0, 4)}.<br />
+ * This returns {@code -1} for {@code floorDiv(-1, 4)}.<br />
+ * This returns {@code -1} for {@code floorDiv(-2, 4)}.<br />
+ * This returns {@code -1} for {@code floorDiv(-3, 4)}.<br />
+ * This returns {@code -1} for {@code floorDiv(-4, 4)}.<br />
+ * This returns {@code -2} for {@code floorDiv(-5, 4)}.<br />
```
ilya-g, line 70: Do not use `{@code}` and `<br>` tags in kdocs.

dkhalanskyjb: Fixed, and also removed such constructs where they were copied from.

ilya-g (approved):

Issue #42, opened 2020-08-21
Instant.minus functions missing


brewin:
I expected these to be available, since the plus functions are:
```
Instant.minus(period: DateTimePeriod, timeZone: TimeZone): Instant
Instant.minus(unit: DateTimeUnit, timeZone: TimeZone): Instant
Instant.minus(value: Long, unit: DateTimeUnit, timeZone: TimeZone): Instant
```

MairwunNx:
I'm also feeling this problem. Hope it will implemented in next minor version.

MairwunNx:
@elizarov this issue can be closed now. :)

PR #49, opened 2020-09-07
Match 'plus' with the corresponding 'minus' functions

dkhalanskyjb:
Solves https://github.com/Kotlin/kotlinx-datetime/issues/42

### DIFF
--- a/core/commonMain/src/Instant.kt
+++ b/core/commonMain/src/Instant.kt
@@ -412,7 +412,7 @@ public fun Instant.minus(value: Long, unit: DateTimeUnit, timeZone: TimeZone) =
     if (value != Long.MIN_VALUE) {
         plus(-value, unit, timeZone)
     } else {
-        plus(-(value+1), unit, timeZone).plus(unit, timeZone)
+        plus(-(value + 1), unit, timeZone).plus(unit, timeZone)
     }
 
 /**
@@ -433,7 +433,12 @@ public expect fun Instant.plus(value: Long, unit: DateTimeUnit.TimeBased): Insta
  *
  * The return value is clamped to the platform-specific boundaries for [Instant] if the result exceeds them.
  */
-public fun Instant.minus(value: Long, unit: DateTimeUnit.TimeBased): Instant = plus(-value, unit)
+public fun Instant.minus(value: Long, unit: DateTimeUnit.TimeBased): Instant =
+    if (value != Long.MIN_VALUE) {
+        plus(-value, unit)
+    } else {
+        plus(-(value + 1), unit).plus(unit)
+    }
 
 /**
  * Returns the whole number of the specified date or time [units][unit] between [other] and `this` instants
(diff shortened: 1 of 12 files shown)


core/commonMain/src/Instant.kt
```diff
@@ -314,6 +356,17 @@ public expect fun Instant.plus(value: Int, unit: DateTimeUnit, timeZone: TimeZon
...
+/**
+ * Returns an instant that is the result of subtracting the [value] number of the specified [unit] from this instant.
+ *
+ * If the [value] is positive, the returned instant is earlier than this instant.
+ * If the [value] is negative, the returned instant is later than this instant.
+ *
+ * The return value is clamped to the platform-specific boundaries for [Instant] if the result exceeds them.
+ */
+public fun Instant.minus(value: Int, unit: DateTimeUnit.TimeBased): Instant =
+    plus(-value.toLong(), unit)
```
ilya-g, line 368: Better `-(value.toLong())`

core/commonMain/src/Instant.kt
```diff
@@ -303,6 +333,18 @@ public fun Instant.plus(unit: DateTimeUnit.TimeBased): Instant =
...
+ * Returns an instant that is the result of subtracting the [value] number of the specified [unit] from this instant
+ * in the specified [timeZone].
+ *
+ * If the [value] is positive, the returned instant is earlier than this instant.
+ * If the [value] is negative, the returned instant is later than this instant.
+ *
+ * @throws DateTimeArithmeticException if this value or the result is too large to fit in [LocalDateTime].
+ */
+public fun Instant.minus(value: Int, unit: DateTimeUnit, timeZone: TimeZone): Instant =
+    plus(-value, unit, timeZone)
```
ilya-g, line 346: Is the negation always overflow-safe? I.e. it certainly can overflow at MIN_VALUE, but would this overflow be compensated later?

core/commonMain/src/Instant.kt
ilya-g: Though it's likely the same, but the order of operations would be more explicit.

core/commonMain/src/Instant.kt
dkhalanskyjb: You're right, this would be silently wrong. I pushed a fix for this and a similar problem with `LocalDate.minus(Int, ...)` on JVM.

core/commonMain/src/Instant.kt
dkhalanskyjb: Done.

core/commonMain/src/Instant.kt
ilya-g: Could you also ensure that `instant.minus(Long.MIN_VALUE, DateTimeUnit.NANOSECOND) > instant`?

dkhalanskyjb: Done. Since this edge case is a requirement, consistency also required solving a whole new can of worms with `DateTimePeriod`, which I also did.

ilya-g (approved):

PR #55, opened 2020-09-21
Native (iOS): fix crash when getting the current time

dkhalanskyjb:
The root cause is unclear as of yet, but this workaround should do.


ilya-g (approved):

dkhalanskyjb:
The cause: while `clock_gettime` is present in the headers for iOS, the symbol itself only became available on iOS 10: https://github.com/jasonacox/Build-OpenSSL-cURL/issues/17

PR #80, opened 2020-12-11
Orthogonal DateTimePeriod implementation

dkhalanskyjb:


core/common/src/DateTimePeriod.kt
```diff
@@ -45,83 +52,269 @@ sealed class DateTimePeriod {
...
 
-class DatePeriod(
-        override val years: Int = 0,
-        override val months: Int = 0,
-        override val days: Int = 0
+public fun String.toDateTimePeriod(): DateTimePeriod = DateTimePeriod.parse(this)
+
+class DatePeriod internal constructor(
+    internal override val totalMonths: Int,
+    override val days: Int = 0,
```
ilya-g, line 238: The default value here looks unnecessary.

core/common/src/DateTimePeriod.kt
```diff
@@ -45,83 +52,269 @@ sealed class DateTimePeriod {
...
 ) : DateTimePeriod() {
+    constructor(years: Int = 0, months: Int = 0, days: Int = 0): this(totalMonths(years, months), days)
+    // avoiding excessive computations
     override val hours: Int get() = 0
     override val minutes: Int get() = 0
-    override val seconds: Long get() = 0
-    override val nanoseconds: Long get() = 0
+    override val seconds: Int get() = 0
+    override val nanoseconds: Int get() = 0
+    override val totalNanoseconds: Long get() = 0
```
ilya-g, line 246: Is it internal? Better to specify the visibility to avoid an impression that it is public.

core/common/src/DateTimePeriod.kt
```diff
@@ -45,83 +52,269 @@ sealed class DateTimePeriod {
...
-    DateTimePeriodImpl(years, months, days, hours, minutes, seconds, nanoseconds)
-else
-    DatePeriod(years, months, days)
+): DateTimePeriod = buildDateTimePeriod(totalMonths(years, months), days,
+    totalNanoseconds(hours, minutes, seconds, nanoseconds))
 
 @OptIn(ExperimentalTime::class)
 fun Duration.toDateTimePeriod(): DateTimePeriod = toComponents { hours, minutes, seconds, nanoseconds ->
-    DateTimePeriod(hours = hours, minutes = minutes, seconds = seconds.toLong(), nanoseconds = nanoseconds.toLong())
+    buildDateTimePeriod(totalNanoseconds = totalNanoseconds(hours, minutes, seconds, nanoseconds.toLong()))
```
ilya-g, line 307: Perhaps we could now avoid splitting duration to components and just convert its nanosecond value?

core/common/test/DateTimePeriodTest.kt
```diff
@@ -17,7 +17,7 @@ class DateTimePeriodTest {
         assertEquals("P1Y", DateTimePeriod(years = 1).toString())
         assertEquals("P1Y1M", DatePeriod(years = 1, months = 1).toString())
         assertEquals("P11M", DateTimePeriod(months = 11).toString())
-        assertEquals("P14M", DateTimePeriod(months = 14).toString()) // TODO: normalize or not
+        assertEquals("P1Y2M", DateTimePeriod(months = 14).toString()) // TODO: normalize or not
```
ilya-g, line 20: // TODO comment is now resolved

core/js/src/Instant.kt
```diff
@@ -111,13 +111,12 @@ public actual fun Instant.plus(period: DateTimePeriod, timeZone: TimeZone): Inst
     val thisZdt = this.value.atZone(timeZone.zoneId)
     with(period) {
         thisZdt
-                .run { if (years != 0 && months == 0) plusYears(years) else this }
-                .run { if (months != 0) plusMonths(years * 12.0 + months) else this }
+                .run { if (totalMonths != 0) plusMonths(totalMonths) else this }
```
ilya-g, line 114: There could have been a reason why years were not fused with months before when months were 0.

core/jvm/test/ConvertersTest.kt
```diff
@@ -85,15 +85,21 @@ class ConvertersTest {
 
     @Test
     fun datePeriod() {
+
+        fun assertJtPeriodEqual(a: JTPeriod, b: JTPeriod) {
```
ilya-g, line 89:
```suggestion
        fun assertJtPeriodNormalizedEquals(a: JTPeriod, b: JTPeriod) {
```

core/js/src/Instant.kt
dkhalanskyjb: I'm assuming we did it because it's slightly more efficient due to not using division: https://github.com/ThreeTen/threetenbp/blob/master/src/main/java/org/threeten/bp/LocalDate.java#L1105-L1142 (the js-joda implementation is virtually the same). With the orthogonal implementation, getting the number of years and months performs this division anyway.

core/common/src/DateTimePeriod.kt
dkhalanskyjb: Fixed.

core/common/src/DateTimePeriod.kt
dkhalanskyjb: These fields are now internal everywhere.

core/common/src/DateTimePeriod.kt
dkhalanskyjb: Thanks!

core/common/test/DateTimePeriodTest.kt
dkhalanskyjb: Fixed.

core/common/src/DateTimePeriod.kt
```diff
@@ -45,83 +52,267 @@ sealed class DateTimePeriod {
...
-        result = 31 * result + minutes
-        result = 31 * result + seconds.hashCode()
-        result = 31 * result + nanoseconds.hashCode()
+        result = 31 * result + totalNanoseconds.hashCode()
         return result
     }
 
-    // TODO: parsing from iso string
+    companion object {
+        fun parse(text: String): DateTimePeriod {
```
ilya-g, line 70: I believe we need `DatePeriod.parse` as well, that prevents parsing (non-zero?) components after `T`

core/common/test/DateTimePeriodTest.kt
```diff
@@ -29,8 +29,43 @@ class DateTimePeriodTest {
...
+        assertEquals("PT0.999999999S", DateTimePeriod(seconds = 1, nanoseconds = -1L).toString())
+        assertEquals("-PT0.000000001S", DateTimePeriod(nanoseconds = -1L).toString())
+        assertEquals("P1DT-0.000000001S", DateTimePeriod(days = 1, nanoseconds = -1L).toString())
+        assertEquals("-PT0.999999999S", DateTimePeriod(seconds = -1, nanoseconds = 1L).toString())
+        assertEquals("P1DT-0.999999999S", DateTimePeriod(days = 1, seconds = -1, nanoseconds = 1L).toString())
+    }
+
+    @Test
+    fun parseIsoString() {
+        assertEquals(DateTimePeriod(years = 1), DateTimePeriod.parse("P1Y"))
```
ilya-g, line 45: Better to assert the expected property values rather than equality as a whole in this test. Because otherwise it's unclear what values these properties would get as a result of parsing.

core/common/src/LocalDate.kt
```diff
@@ -137,6 +137,8 @@ public operator fun LocalDate.minus(period: DatePeriod): LocalDate =
  * - negative or zero if this date is later than the other,
  * - exactly zero if this date is equal to the other.
  *
+ * @throws DateTimeArithmeticException if the number of months between the two dates exceeds an Int (JVM only)
```
ilya-g, line 140: Terminate with a dot.

core/js/src/Instant.kt
ilya-g: ok, makes sense.

core/common/test/DateTimePeriodTest.kt
```diff
@@ -29,8 +29,43 @@ class DateTimePeriodTest {
...
+
+        assertEquals(DateTimePeriod(years = -1, months = -2, days = -3, hours = -4, minutes = -5, seconds = 0, nanoseconds = 500_000_000),
+            DateTimePeriod.parse("P-1Y-2M-3DT-4H-5M0.500000000S"))
+
+        assertEquals(DateTimePeriod(nanoseconds = 999_999_999_999_999L), DateTimePeriod.parse("PT277H46M39.999999999S"))
+        assertEquals(DateTimePeriod(seconds = 1, nanoseconds = -1L), DateTimePeriod.parse("PT0.999999999S"))
+        assertEquals(DateTimePeriod(nanoseconds = -1L), DateTimePeriod.parse("-PT0.000000001S"))
+        assertEquals(DateTimePeriod(days = 1, nanoseconds = -1L), DateTimePeriod.parse("P1DT-0.000000001S"))
+        assertEquals(DateTimePeriod(seconds = -1, nanoseconds = 1L), DateTimePeriod.parse("-PT0.999999999S"))
+        assertEquals(DateTimePeriod(days = 1, seconds = -1, nanoseconds = 1L), DateTimePeriod.parse("P1DT-0.999999999S"))
```
ilya-g, line 68: A couple of negative tests won't hurt. For example that DatePeriod.parse fails on non-zero time components.

ilya-g:
Please refer #81 in (one of) the final commit message.

core/common/test/DateTimePeriodTest.kt
dkhalanskyjb: Added some.

core/common/src/LocalDate.kt
dkhalanskyjb: Fixed.

core/common/test/DateTimePeriodTest.kt
dkhalanskyjb: You raise a good point, though I think that the values of properties and the results of parsing are separate concerns, so I added another test to check normalization.

core/common/test/DateTimePeriodTest.kt
```diff
@@ -12,6 +12,39 @@ import kotlin.time.*
 
 class DateTimePeriodTest {
 
+    @Test
+    fun normalization() {
+        assertPeriodComponents(DateTimePeriod(years = 1) as DatePeriod, years = 1)
```
ilya-g, line 17: Can you use this assertion function in tests for `parse` as well?

dkhalanskyjb: Sure, but, in my opinion, it would make those tests less clear. For example,
```kotlin
         assertEquals(DateTimePeriod(months = 14), DateTimePeriod.parse("P14M"))
```
clearly implies that `DateTimePeriod(months = m) == DateTimePeriod.parse("P${m}M")`, which is a property that we want. Seeing the result in terms of values of class properties would obscure this. Moreover, the values class properties are now being tested with `normalization()`.

What benefits do you see in using this function for tests of parsing?

ilya-g: By mixing parsing and normalized equality this test makes an impression that the resulting properties of the parsed `DateTimePeriod` would be the same as those passed to the constructor-like function of expected value. This is misleading.

dkhalanskyjb: For some unfathomable reason, I didn't see a notification about your response. I am sorry for the delay.

I still disagree and consider these concerns orthogonal: we want parsing to behave the same way construction does, and we want construction to produce specific values of properties. Parsing is not semantically connected to the values of properties, and mixing these rubs me the wrong way. However, I don't consider this issue important enough to die on this hill, so I added a commit to accommodate your preferences, though I would be glad to revert it were you to change your mind about this.

ilya-g: As a compromise, we could leave the equality assertions in the parsing test, but add `assertPeriodComponents` for those values that cause normalization to happen. This would be done mostly for explanatory reasons.

dkhalanskyjb: I added checks for actual values of properties to places where normalization does affect them.

ilya-g (approved):
Please refer #81 in (one of) the final commit message.

PR #93, opened 2021-01-12
Support x86_64-based watchOS simulator target

sbogolepov:


goncalossilva:
Can this be merged and released? 🙏 Our adoption of KMP is blocked on this...

goncalossilva:
[Swapping `compileOnly` with `api`](https://github.com/Kotlin/kotlinx-datetime/blob/master/core/build.gradle.kts#L155) fixes the issue. It is strangely related with #110.

Unfortunately, that issue does not provide a lot of context nor do I have much, so I am unsure if this is an acceptable compromise.

goncalossilva:
There are also version mismatches, e.g., the setup using 1.4.30 and serializable pulling in 1.5.0, whose APIs are incompatible.

dkhalanskyjb:
`watchosX64` support added in the latest release (0.2.1).

PR #103, opened 2021-03-18
Add Dokka 1.4.30 for documentation

kamildoleglo:
Currently blocked by https://github.com/Kotlin/kotlinx.html/issues/81

There are several unresolved classes on the native platform unfortunately, this may happen until Kotlin/Native has a stable API

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -9,6 +9,7 @@ buildscript {
 
 plugins {
     id("kotlinx.team.infra") version "0.3.0-dev-64"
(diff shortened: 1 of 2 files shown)


dkhalanskyjb:
Just FYI: it looks like https://github.com/Kotlin/kotlinx.html/issues/81 is fixed.

hfhbd:
I think, you could close this PR, see https://github.com/Kotlin/kotlinx-datetime/pull/231

Issue #58, opened 2020-10-13
Support custom formats (for parsing as well as toString operations)


fkrauthan:
I think one of the biggest missing items is a formatting system to parse custom date formats as well as to serialize a date object back to a string with custom format. When working with dates both directions are very important and common.

tadfisher:
A replacement (or delegate) for the ICU RelativeDateTimeFormatter would be a useful addition. The use case is to format localized absolute times, e.g. "today", "yesterday", and "last week".

tylerwilson:
Hope this is being worked on. I would love to use this package cross-platform, but not having a generic formatting/parsing system is holding us back. Thanks.

Artrosis:
I needed formatting for logs

Issue #39, opened 2020-08-19
Feature Request: Parse Instant from complex pattern


Brutus5000:
**Use case:**
I am working with timeseries data coming from 
- sensors of all sorts of types and vendors
- business files (e.g. Excel exports in all kind of localizations and formats)
- output from other systems in all sorts of formats.

From my experience the amount of variations is almost unlimited. Be it the separators (or sometimes the lack of), the order of the elements and also the potential absence of some parts (optionally millisecond fractions, optionally offsets minutes, ...). The requirement goes way beyond what you can usually represent as a string pattern with describing string like ("yyyy-MM-dd") which are fairly unreadable as well if you think about the amount of variables for non-standard-stuff (fractions, timezone hour, timezone minute, ...)

**Reference solution:**
Coming from Java, all of these formats can be easily represented using a `DateTimeFormatter` and described using a `DateTimeFormatterBuilder`. The DateTimeFormatter offers a lot of predefined ISO formats as well.

Side note: The DateTimeFormatter can be used for printing instants as well.

**Summary:**
A comprehensive datetime library needs utilities to allow import and export of date/time strings in all formats for compatibility with the rest of the world.

tylerwilson:
Agreed. I deal with an old REST API that has something like 8 different date and time formats I need to parse. So a generic formatting option, to and from String, is needed in my case.

PR #106, opened 2021-04-03
Implement Instant parsing in common module

### LINKED ISSUE #56: Allow to parse instants with UTC offsets other than Z
The README states that e.g. `Instant` only has ISO-8601 support but i think it only has partial support.

`Instant.parse("2020-09-21T09:12:47Z")` works as expected but `Instant.parse("2020-09-21T09:12:47+00:00")` throws a DateTimeParseException

okarmazin:
Resolves #56 

Instant.parse now supports:
- Time zone offsets in the form of `+-hh:mm` and `+-hhmm`
- Time part allows the omission of colons (allows hhmmss)
- Time part allows the omission of seconds

Implementation notes:
Input string is split by the time delimiter (T|t) into date and time parts.
Date parsing is delegated to `LocalDate.parse`
Time parsing employs the well known algorithm from `Iso8601Utils.java`
originally implemented in Jackson. This commit is based on Moshi's
version of that file.

##### Motivation
Any datetime library worth its salt *must* be able to parse numeric time zone offsets. I'm afraid here's no way around it.
I initially implemented this for myself since I need the ability to parse numeric time zone offsets. In fact I think that most of the `OffsetDateTime`-type strings floating around the Internet come with the `+-hh:mm` offset representation rather than the `Z` designator. Migrating from `ThreeTenBP` I was surprised that only the `Z` designator was supported.

This implementation may hold fort until a format builder is added (#58 #39)

###### Discussion points:
The leniency of the time parser comes free with the algorithm.
1. Should the omission of colons in time be allowed?
2. Should the omission of seconds be allowed?

I vote that the leniency be kept in the spirit of the "Be permissive on input, strict on output" rule of thumb.

### DIFF
--- a/core/common/src/InstantParser.kt
+++ b/core/common/src/InstantParser.kt
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2019-2021 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.datetime
+
+import kotlin.math.min
+import kotlin.math.pow
+
+internal fun parseInstantCommon(string: String): Instant = parseIsoString(string)
+
+/*
+ * The algorithm for parsing time and zone offset was adapted from
+ * https://github.com/square/moshi/blob/aea17e09bc6a3f9015d3de0e951923f1033d299e/adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.java
+ */
+private fun parseIsoString(isoString: String): Instant {
+    try {
+        val dateTimeSplit = isoString.split('T', ignoreCase = true)
+        if (dateTimeSplit.size != 2) {
+            throw DateTimeFormatException("ISO 8601 datetime must contain exactly one (T|t) delimiter.")
+        }
+        val localDate = LocalDate.parse(dateTimeSplit[0])
+
+        // Iso8601Utils.parse
+        val timePart = dateTimeSplit[1]
+        var offset = 0
+        val hour = parseInt(timePart, offset, offset + 2).also { offset += 2 }
+        if (checkOffset(timePart, offset, ':')) {
+            offset += 1
+        }
+        val minutes = parseInt(timePart, offset, offset + 2).also { offset += 2 }
+        if (checkOffset(timePart, offset, ':')) {
+            offset += 1
+        }
+
+        var seconds = 0
+        var nanosecond = 0
+        // seconds and fraction can be optional
+        if (timePart.length > offset) {
+            val c = timePart[offset]
+            if (c != 'Z' && c != 'z' && c != '+' && c != '-') {
+                seconds = parseInt(timePart, offset, offset + 2).also { offset += 2 }
+                if (seconds > 59 && seconds < 63) { // https://github.com/Kotlin/kotlinx-datetime/issues/5
+                    seconds = 59 // truncate up to 3 leap seconds
+                }
+                if (checkOffset(timePart, offset, '.')) {
+                    offset += 1
+                    val endOffset =
+                        indexOfNonDigit(timePart, offset + 1) // assume at least one digit
+                    val parseEndOffset =
+                        min(endOffset, offset + 9) // parse up to 9 digits
+                    val fraction = parseInt(timePart, offset, parseEndOffset)
+                    nanosecond = (10.0.pow(9 - (parseEndOffset - offset)) * fraction).toInt()
+                    offset = endOffset
+                }
+            }
+        }
+
+        // extract timezone
+        if (timePart.length <= offset) {
+            throw IllegalArgumentException("No time zone indicator in '$timePart'")
+        }
+        val timezone: TimeZone
+        val timezoneIndicator = timePart[offset]
+        if (timezoneIndicator == 'Z' || timezoneIndicator == 'z') {
+            timezone = TimeZone.UTC
+        } else if (timezoneIndicator == '+' || timezoneIndicator == '-') {
+            val timezoneOffset = timePart.substring(offset)
+            // 18-Jun-2015, tatu: Minor simplification, skip offset of "+0000"/"+00:00"
+            if ("+0000" == timezoneOffset || "+00:00" == timezoneOffset) {
+                timezone = TimeZone.UTC
+            } else {
+                val timezoneId = "UTC$timezoneOffset"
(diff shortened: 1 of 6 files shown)


core/common/src/InstantParser.kt
```diff
@@ -0,0 +1,142 @@
...
+package kotlinx.datetime
+
+import kotlin.math.min
+import kotlin.math.pow
+
+internal fun parseInstantCommon(string: String): Instant = parseIsoString(string)
+
+/*
+ * The algorithm for parsing time and zone offset was adapted from
+ * https://github.com/square/moshi/blob/aea17e09bc6a3f9015d3de0e951923f1033d299e/adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.java
```
okarmazin, line 15: Please advise on the proper formulation of attribution that shall be added.

core/common/src/InstantParser.kt
```diff
@@ -0,0 +1,142 @@
...
+
+        var seconds = 0
+        var nanosecond = 0
+        // seconds and fraction can be optional
+        if (timePart.length > offset) {
+            val c = timePart[offset]
+            if (c != 'Z' && c != 'z' && c != '+' && c != '-') {
+                seconds = parseInt(timePart, offset, offset + 2)
+                offset += 2
+                if (seconds > 59 && seconds < 63) { // https://github.com/Kotlin/kotlinx-datetime/issues/5
```
okarmazin, line 45: Is #5 applicable here, or should this be converted to idiomatic range check?

core/common/src/InstantParser.kt
```diff
@@ -0,0 +1,142 @@
...
+ * Copyright 2019-2021 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.datetime
+
+import kotlin.math.min
+import kotlin.math.pow
+
+internal fun parseInstantCommon(string: String): Instant = parseIsoString(string)
```
okarmazin, line 11: The parseInstantCommon facade may try multiple formats before giving up such as #83 if that gets implemented.

core/common/src/InstantParser.kt
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        var seconds = 0
+        var nanosecond = 0
+        // seconds and fraction can be optional
+        if (timePart.length > offset) {
+            val c = timePart[offset]
+            if (c != 'Z' && c != 'z' && c != '+' && c != '-') {
+                seconds = parseInt(timePart, offset, offset + 2).also { offset += 2 }
+                if (seconds > 59 && seconds < 63) { // https://github.com/Kotlin/kotlinx-datetime/issues/5
```
okarmazin, line 44: Is #5 applicable here, or should this be converted to idiomatic range check?

core/common/src/InstantParser.kt
```diff
@@ -0,0 +1,141 @@
...
+        if (timePart.length <= offset) {
+            throw DateTimeFormatException("No time zone indicator in '$timePart'")
+        }
+        val timezone: TimeZone
+        val timezoneIndicator = timePart[offset]
+        if (timezoneIndicator == 'Z' || timezoneIndicator == 'z') {
+            timezone = TimeZone.UTC
+        } else if (timezoneIndicator == '+' || timezoneIndicator == '-') {
+            val timezoneOffset = timePart.substring(offset)
+            // 18-Jun-2015, tatu: Minor simplification, skip offset of "+0000"/"+00:00"
```
dkhalanskyjb, line 70: These comments have little meaning without the commit history, so they probably shouldn't be included here.

core/js/src/Instant.kt
```diff
@@ -75,12 +75,7 @@ public actual class Instant internal constructor(internal val value: jtInstant)
             if (epochMilliseconds > 0) MAX else MIN
         }
 
-        actual fun parse(isoString: String): Instant = try {
-            Instant(jtInstant.parse(isoString))
-        } catch (e: Throwable) {
-            if (e.isJodaDateTimeParseException()) throw DateTimeFormatException(e)
-            throw e
-        }
+        actual fun parse(isoString: String): Instant = parseInstantCommon(isoString)
```
dkhalanskyjb, line 78: This could be implemented easily—and consistently with the other parsers—by parsing a `ZonedDateTime` and then converting that to `Instant`.

The same for the Java implementation.

core/native/src/Instant.kt
```diff
@@ -23,54 +23,6 @@ public actual enum class DayOfWeek {
...
-        .chain(intParser(2, 2)) // hour
-        .chainIgnoring(concreteCharParser(':'))
-        .chain(intParser(2, 2)) // minute
-        .chainIgnoring(concreteCharParser(':'))
-        .chain(intParser(2, 2)) // second
-        .chain(optional(
-            concreteCharParser('.')
-                .chainSkipping(fractionParser(0, 9, 9)) // nanos
-        ))
-        .chainIgnoring(concreteCharParser('Z').or(concreteCharParser('z')))
```
dkhalanskyjb, line 40: If the other parsers are implemented via `ZonedDateTime`, there is no need to have the parser for `Instant` in the common code, and the only place that requires an explicit implementation is the native part. Then, it's probably easier to modify this parser at this line so that it accepts any `ZoneOffset` and not just `Z` (and propagate this change correspondingly).

dkhalanskyjb (changes requested):
Hi! Thanks for the PR. We are actually interested in making this change and wanted to do it before the library's next release.

We can't accept this PR as is, though, given that the proposed changes are inconsistent with how the other parsers behave (`LocalDateTime` in particular). Moreover, it initially looks like this change could be implemented in a much more simple manner by parsing `ZonedDateTime` (EDIT: or `OffsetDateTime`?) on JVM and JS and slightly editing the existing parser on Native, and it's unclear why it would be worth doing anything other than that.

okarmazin:
Thank you for your review. Since this code is highly inconsistent with existing logic I'll close this PR down.

As you're _already_ planning on making the change yourselves and have it done by the next release (this month is the target according to [Ilya Gorbunov on Slack](https://kotlinlang.slack.com/archives/C01923PC6A0/p1617187585007200?thread_ts=1617187416.006900&cid=C01923PC6A0) I see no reason for a third party to be a part of that process. I hope this bootleg PR reinforces the fact that accepting time zone offsets is critically important. I'm looking forward to the next release!

PR #107, opened 2021-04-07
Implement parsing Instant with offset

dkhalanskyjb:
Fixes https://github.com/Kotlin/kotlinx-datetime/issues/56

### DIFF
--- a/core/common/src/Instant.kt
+++ b/core/common/src/Instant.kt
@@ -478,4 +478,9 @@ public fun Instant.minus(other: Instant, unit: DateTimeUnit.TimeBased): Long =
 internal const val DISTANT_PAST_SECONDS = -3217862419201
 internal const val DISTANT_FUTURE_SECONDS = 3093527980800
 
+/**
+ * Displays the given Instant in the given [offset].
+ *
+ * Be careful: this function may throw for some values of the [Instant].
+ */
 internal expect fun Instant.toStringWithOffset(offset: ZoneOffset): String
--- a/core/common/src/Instant.kt
+++ b/core/common/src/Instant.kt
@@ -125,9 +125,10 @@ public expect class Instant : Comparable<Instant> {
          * Parses a string that represents an instant in ISO-8601 format including date and time components and
          * the mandatory time zone offset and returns the parsed [Instant] value.
          *
-         * In addition to allowing the `Z` designator of the UTC+0 time zone to be passed as the time zone offset, which
-         * is required by the ISO-8601, this parser also accepts the other possible offsets, but does not provide a way
-         * to query the result for which offset was specified in the string.
+         * Supports the following ways of specifying the time zone offset:
+         * - the `Z` designator for the UTC+0 time zone,
+         * - a custom time zone offset specified with `+hh`, or `+hh:mm`, or `+hh:mm:ss`
+         *   (with `+` being replaced with `-` for the negative offsets)
          *
          * Examples of instants in the ISO-8601 format:
          * - `2020-08-30T18:43:00Z`
@@ -476,3 +477,5 @@ public fun Instant.minus(other: Instant, unit: DateTimeUnit.TimeBased): Long =
 
 internal const val DISTANT_PAST_SECONDS = -3217862419201
 internal const val DISTANT_FUTURE_SECONDS = 3093527980800
+
+internal expect fun Instant.toStringWithOffset(offset: ZoneOffset): String
--- a/core/common/src/Instant.kt
+++ b/core/common/src/Instant.kt
@@ -123,12 +123,18 @@ public expect class Instant : Comparable<Instant> {
 
         /**
          * Parses a string that represents an instant in ISO-8601 format including date and time components and
-         * the mandatory `Z` designator of the UTC+0 time zone and returns the parsed [Instant] value.
+         * the mandatory time zone offset and returns the parsed [Instant] value.
          *
-         * Examples of instants in ISO-8601 format:
+         * In addition to allowing the `Z` designator of the UTC+0 time zone to be passed as the time zone offset, which
+         * is required by the ISO-8601, this parser also accepts the other possible offsets, but does not provide a way
+         * to query the result for which offset was specified in the string.
+         *
+         * Examples of instants in the ISO-8601 format:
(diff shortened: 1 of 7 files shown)


core/common/src/Instant.kt
```diff
@@ -123,12 +123,18 @@ public expect class Instant : Comparable<Instant> {
 
         /**
          * Parses a string that represents an instant in ISO-8601 format including date and time components and
-         * the mandatory `Z` designator of the UTC+0 time zone and returns the parsed [Instant] value.
+         * the mandatory time zone offset and returns the parsed [Instant] value.
          *
-         * Examples of instants in ISO-8601 format:
+         * In addition to allowing the `Z` designator of the UTC+0 time zone to be passed as the time zone offset, which
+         * is required by the ISO-8601, this parser also accepts the other possible offsets, but does not provide a way
+         * to query the result for which offset was specified in the string.
```
ilya-g, line 130: I propose to rephrase this simpler, something like:
> Supports the following ways of specifying the time zone offset:
> - `Z` designator for the UTC+0 time zone,
> - a custom time zone offset specified with `+hh`, or `+hh:mm`, or `+hh:mm:ss`, where negative offsets are prefixed with `-` instead of `+`.

core/common/test/InstantTest.kt
```diff
@@ -85,6 +85,27 @@ class InstantTest {
...
     }
 
+    @Test
+    fun parseStringsWithOffsets() {
+        val instants = arrayOf(
+            Triple("2020-01-01T00:01:01.02+18:00", 1577772061L, 20_000_000),
+            Triple("2020-01-01T00:01:01.123456789-17:59:59", 1577901660L, 123456789),
+            Triple("2020-01-01T00:01:01.010203040+17:59:59", 1577772062L, 10203040),
+            Triple("2020-01-01T00:01:01.010203040+17:59", 1577772121L, 10203040),
+            // Triple("2020-01-01T00:01:01+00", 1577836861L, 0), // fails on JS, passes everywhere else
```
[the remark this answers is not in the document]
dkhalanskyjb: Yeah, this is the proper way to write this test. Fixed.

core/common/test/InstantTest.kt
```diff
@@ -85,6 +85,27 @@ class InstantTest {
         assertInvalidFormat { Instant.parse("+1000000001-12-31T23:59:59.000000000Z") }
     }
 
+    @Test
+    fun parseStringsWithOffsets() {
```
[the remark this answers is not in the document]
dkhalanskyjb: If by distant dates you mean `DISTANT_PAST` and `DISTANT_FUTURE`, then ok, added them. If you mean something larger, then parsing fails when the parsed dates are outside `LocalDateTime.MIN/MAX`, as `OffsetDateTime` uses `LocalDateTime` internally (as does `ZonedDateTime`, for that matter, so both of them fail to parse `Instant.MAX.toString()`).

core/common/src/Instant.kt
dkhalanskyjb: Agreed, I was clever by half there. However, now that we are so explicit about the accepted offsets, we can't just ignore that parsing `+hh` fails on JS, so I added a workaround for that.

dkhalanskyjb: Turns out that JDK8 also fails to parse `+hh`, as evidenced by the build failing. Added a workaround for that, too.

core/jvm/src/Instant.kt
```diff
@@ -64,11 +64,19 @@ public actual class Instant internal constructor(internal val value: jtInstant)
...
         public actual fun parse(isoString: String): Instant = try {
-            Instant(jtOffsetDateTime.parse(isoString).toInstant())
+            Instant(jtOffsetDateTime.parse(fixOffsetRepresentation(isoString)).toInstant())
         } catch (e: DateTimeParseException) {
             throw DateTimeFormatException(e)
         }
 
+        /** A workaround for a quirk of the JDKs older than 11 where the string representations of Instant that have an
+         * offset of the form "+XX" are not recognized by [jtOffsetDateTime.parse], while "+XX:XX" work fine. */
+        private fun fixOffsetRepresentation(isoString: String): String {
```
ilya-g, line 74: Can we make it non-allocating in common case?

core/common/test/InstantTest.kt
```diff
@@ -87,23 +87,47 @@ class InstantTest {
...
+            TimeZone.of("+03:12:14") as ZoneOffset,
+            TimeZone.of("-03:12:14") as ZoneOffset,
+            TimeZone.of("+02:35") as ZoneOffset,
+            TimeZone.of("-02:35") as ZoneOffset,
+            TimeZone.of("+04") as ZoneOffset,
+            TimeZone.of("-04") as ZoneOffset,
+        )
+
+        for (instant in instants) {
+            for (offset in offsets) {
```
ilya-g, line 126: `offset` loop variable seems to be unused.

core/common/test/InstantTest.kt
```diff
@@ -87,23 +87,47 @@ class InstantTest {
...
+            Pair("2020-01-01T00:01:01.02+18:00", "2019-12-31T06:01:01.020Z"),
+            Pair("2020-01-01T00:01:01.123456789-17:59:59", "2020-01-01T18:01:00.123456789Z"),
+            Pair("2020-01-01T00:01:01.010203040+17:59:59", "2019-12-31T06:01:02.010203040Z"),
+            Pair("2020-01-01T00:01:01.010203040+17:59", "2019-12-31T06:02:01.010203040Z"),
+            Pair("2020-01-01T00:01:01+00", "2020-01-01T00:01:01Z"),
         )
-        instants.forEach {
-            val (str, seconds, nanos) = it
+        strings.forEach {
+            val (str, strInZ) = it
```
ilya-g, line 98: Minor, style: use lambda or loop destructuring, e.g. `strings.forEach { (str, strInZ) -> ...`

core/jvm/src/Instant.kt
dkhalanskyjb: Of course, but the implementation feels more elaborate.

ilya-g (approved):

PR #109, opened 2021-04-21
Fix breaking builds on Native due to depending on serialization

dkhalanskyjb:
Looks like `compileOnly` doesn't, in fact, ensure that there is no
runtime dependency, as Kotlin/Native does not support that. Thus,
it looks like `compileOnly` transforms into something like `api` in
a contrived manner, which leads to the tests implicitly acquiring
a runtime dependency.

Therefore, the changes:
* There is now a separate artifact for serialization.
* The datetime classes are no longer marked as `@Serializable` and
  don't have default serializers.
* The package name for serializers is now
  "kotlinx.datetime.serialization", as it makes sense for it to be
  similar to the name of the artifact, and an artifact with the
  name "serializers" would look odd.


DRSchlaubi:
But that also means that you now always have to put a @Contextual above those dependencies

dkhalanskyjb:
Closed in favor of https://github.com/Kotlin/kotlinx-datetime/pull/110

PR #110, opened 2021-04-22
Add api dependencies for serialization to JS and Native

dkhalanskyjb:
An alternative attempt at fixing the build on Native (and, it turns out, on JS, too).


dkhalanskyjb:
I checked this PR on https://github.com/dkhalanskyjb/testRcDatetime, whose build used to fail but now passes, on MacOS and Linux.

ilya-g (approved):

PR #113, opened 2021-05-04
Humanize LocalDateTime

mitchtabian:
Add extension function for humanizing LocalDateTime.


DRSchlaubi:
Wouldn't it be better to have java.time like formatters instead of one specific format

mitchtabian:
Yes I was hoping this would give someone the idea to do it :)

Issue #118, opened 2021-05-18
uses wrong name for kotlinx-serialization module with legacy compiler


ethanmdavidson:
repo demonstrating issue: https://github.com/ethanmdavidson/serializationNameBug

When using the legacy compiler, kotlinx-datetime uses an old name when referencing kotlinx-serialization: 'kotlinx-serialization-kotlinx-serialization-core-jsLegacy' instead of 'kotlinx-serialization-kotlinx-serialization-core-js-legacy'.

May be related to https://youtrack.jetbrains.com/issue/KT-46598

I think it might be fixed by simply upgrading kotlinx.serialization to 1.2.1 (latest version)

dkhalanskyjb:
Thanks for raising this, your reproducer is great! Yes, either rebuilding the datetime library with dependency on `kotlinx.serialization` 1.2.1 or using `kotlinx.serialization` 1.1.0 does fix the issue. This problem is unexpected, we will be looking into it.

set321go:
It looks like this is the cause https://youtrack.jetbrains.com/issue/KT-43237 so my interpretation is everything that there is now an incompatability between modules published with 1.4 and 1.5

PR #119, opened 2021-05-20
Enable support for watchosX64

goncalossilva:
Supersedes #93, with the changes:
- Enables `watchosX64` in both modules, not just core.
- Rebased from `master`, fixing inconsistencies with the stdlib (1.4.30 and 1.5.0 are source-incompatible, hence the CI failure in #93).
- The above enables switching to kotlinx.serialization 1.2.1 (watchosX64 is only supported on 1.2.0 and up).
- Simplified setup, with kotlinx.serialization being declared only once.
- Test suite should be green, but I can't trigger TeamCity. 😄

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -36,6 +36,7 @@ kotlin {
             target("watchosArm32")
             target("watchosArm64")
             target("watchosX86")
+            target("watchosX64")
             target("tvosArm64")
             target("tvosX64")
         }
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -152,7 +152,7 @@ kotlin {
         commonMain {
             dependencies {
                 api("org.jetbrains.kotlin:kotlin-stdlib-common")
-                compileOnly("org.jetbrains.kotlinx:kotlinx-serialization-core:$serializationVersion")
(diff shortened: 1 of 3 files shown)


goncalossilva:
Is there any chance this might be reviewed soon?

We have an internal milestone by the end of the month that needs this... We can package and publish this internally for demonstration purposes, but if we have a chance to work with the official build, we'd take it. Thanks!

dkhalanskyjb:
> Simplified setup, with kotlinx.serialization being declared only once.

The complex setup that we have is intentional. Its purpose is so that the JVM version doesn't introduce a dependency on the serialization library. So, we have to make it work with a `compileOnly` dependency in the common code.

We will need to make a new release in order to fix https://github.com/Kotlin/kotlinx-datetime/issues/118, and `watchosX64` support would likely fit there nicely, so, if everything goes well, we'll have a release by the end of the week that introduces it.

goncalossilva:
> > Simplified setup, with kotlinx.serialization being declared only once.
>
> The complex setup that we have is intentional. Its purpose is so that the JVM version doesn't introduce a dependency on the serialization library. So, we have to make it work with a compileOnly dependency in the common code.

That should be an easy change, but it's probably not worth it if you're planniing to fix it as part of #118? Should we close this PR?

goncalossilva:
Changes in #120.

Issue #75, opened 2020-11-29
No support for linuxArm32Hfp as a target for Kotlin/Native


v79:
I notice that there's no support for ARM linux platforms in kotlinx-datetime (a common problem across K/N projects, sadly!).

Specifically, I am writing for the Raspberry Pi (linuxArm32Hfp) and it's not listed as a valid target. All I've found are:

kotlinx-datetime-iosarm32/
kotlinx-datetime-iosarm64/
kotlinx-datetime-iosx64/
kotlinx-datetime-js/
kotlinx-datetime-jvm/
kotlinx-datetime-linuxx64/
kotlinx-datetime-macosx64/
kotlinx-datetime-metadata/
kotlinx-datetime-mingwx64/
kotlinx-datetime-tvosarm64/
kotlinx-datetime-tvosx64/
kotlinx-datetime-watchosarm32/
kotlinx-datetime-watchosarm64/
kotlinx-datetime-watchosx86/
kotlinx-datetime/

Are there likely to be any major challenges in just duplicating the codebase from kotlinx-datetime-linuxx64 and recompiling? Is that something I could help with?

v79:
I prepared a local build for my Raspberry Pi 4, and it just worked. I added:

```kotlin
kotlin {
    infra {
        target("linuxArm32Hfp")
  }
}
```

To `build.gradle.kts` and executed `gradle clean publishLinuxArm32HfpPublicationToMavenLocal`. I could then add it to my project, compile it (in Windows) and deploy it to the Pi. The code works, so version 0.2.0-SNAPSHOT looks good to add `linuxArm32Hfp` as a supported target.

dkhalanskyjb:
Yeah, it would make sense to provide builds for some other targets. The main problem is infrastructural: we don't have automated testing for linuxArm yet. The code *probably* works as is, but we are not going to publish it until we have a working CI. Not having CI (especially for one platforms where other platforms do have it!) can lead to nightmarish subtle problems due to human errors, and as an official library, we have additional responsibility. We are just not going to risk it.

That said, there are some good news: an active discussion of this problem is underway. A public-facing issue where you can track progress can be found here: https://youtrack.jetbrains.com/issue/KT-43996

dkhalanskyjb:
Alas, the `linuxArm32Hfp` target is deprecated: https://kotlinlang.org/docs/native-target-support.html#deprecated-targets There's nothing we can do in this library to work around that.

PR #122, opened 2021-05-27
Adding the linuxArm64 target support

KirillObukhov:


dkhalanskyjb:
Thanks for the PR! Sorry, but we'll close it for now. See https://github.com/Kotlin/kotlinx-datetime/issues/75#issuecomment-748888306 for an explanation that is applicable here as well. We'll add all the `linuxArm` targets once either 1) the testing infrastructure is in place or 2) Kotlin/Native establishes a notion of support tiers that permits the official libraries to be released without being thoroughly tested for `linuxArm`.

rossdanderson:
Given the support tiers have now been set up, could we reopen this one?

https://youtrack.jetbrains.com/issue/KT-43996

Thanks :)

dkhalanskyjb:
@rossdanderson, no need to reopen the PR, the `linuxArm64` support is already merged into `master` and will be available as part of the next release: https://github.com/Kotlin/kotlinx-datetime/blob/master/core/build.gradle.kts#L44

PR #124, opened 2021-06-03
Fix deprecations and minor cleanup in build scripts

ilya-g:


core/jvm/src/Instant.kt
```diff
@@ -43,8 +43,8 @@ public actual class Instant internal constructor(internal val value: jtInstant)
     public actual operator fun minus(duration: Duration): Instant = plus(-duration)
 
     public actual operator fun minus(other: Instant): Duration =
-            (this.value.epochSecond - other.value.epochSecond).seconds + // won't overflow given the instant bounds
-            (this.value.nano - other.value.nano).nanoseconds
+            Duration.seconds((this.value.epochSecond - other.value.epochSecond)) + // won't overflow given the instant bounds
+            Duration.nanoseconds((this.value.nano - other.value.nano))
```
dkhalanskyjb, line 47:
```suggestion
            Duration.seconds(this.value.epochSecond - other.value.epochSecond) + // won't overflow given the instant bounds
            Duration.nanoseconds(this.value.nano - other.value.nano)
```

dkhalanskyjb (approved):

core/jvm/src/Instant.kt
ilya-g: Thanks for spotting, removed a couple more unnecessary parentheses.

PR #131, opened 2021-07-08
Remove extra hierarchy level in DateTimeUnit sealed class inheritors

ilya-g:


core/common/src/DateTimeUnit.kt
```diff
@@ -66,44 +66,51 @@ public sealed class DateTimeUnit {
...
     public sealed class DateBased : DateTimeUnit() {
-        // TODO: investigate how to move subclasses up to DateTimeUnit scope
-        @Serializable(with = DayBasedDateTimeUnitSerializer::class)
-        public class DayBased(public val days: Int) : DateBased() {
-            init {
-                require(days > 0) { "Unit duration must be positive, but was $days days." }
-            }
+        @Suppress("TOPLEVEL_TYPEALIASES_ONLY")
+        @Deprecated("Use DateTimeUnit.DayBased", ReplaceWith("DateTimeUnit.DayBased", "kotlinx.datetime.DateTimeUnit"))
+        public typealias DayBased = DateTimeUnit.DayBased
```
dkhalanskyjb, line 71: I seem to remember that these type aliases didn't work in some regard, right? I tried to use them and everything worked fine. Does this mean that you found a solution to that issue? If not, `Deprecated` should probably be with the `DeprecationLevel.ERROR`.

dkhalanskyjb (approved):

core/common/src/DateTimeUnit.kt
ilya-g: As I recall, they caused problems when used from inside of the `DateTimeUnit` class. That's why I had to prefix the affected types with `DateTimeUnit.` — to avoid resolving to type aliases there.

PR #132, opened 2021-07-09
Add String.toInstantOrNull

hfhbd:
Many converter functions of the standard library contains a `orNull` function instead throwing, like `String.toIntOrNull`.

This PR add this function to `String.toInstant`.

### DIFF
--- a/core/common/src/Instant.kt
+++ b/core/common/src/Instant.kt
@@ -182,6 +182,16 @@ public val Instant.isDistantFuture: Boolean
 public fun String.toInstant(): Instant = Instant.parse(this)
 
 /**
+ * Converts this string representing an instant in ISO-8601 format including date and time components and
+ * the time zone offset to an [Instant] value.
+ *
+ * See [Instant.parse] for examples of instant string representations.
+ *
+ * @throws IllegalArgumentException if the text cannot be parsed or the boundaries of [Instant] are exceeded.
+ */
+public fun String.toInstantOrNull(): Instant? = try { Instant.parse(this) } catch(cause: IllegalArgumentException) { null }
+
+/**
  * Returns an instant that is the result of adding components of [DateTimePeriod] to this instant. The components are
  * added in the order from the largest units to the smallest, i.e. from years to nanoseconds.
  *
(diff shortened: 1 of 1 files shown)


core/common/src/Instant.kt
```diff
@@ -181,6 +181,16 @@ public val Instant.isDistantFuture: Boolean
...
 
+/**
+ * Converts this string representing an instant in ISO-8601 format including date and time components and
+ * the time zone offset to an [Instant] value.
+ * 
+ * Returns null, if this string does not match the ISO-8601 format or the boundaries of [Instant] are exceeded.
+ *
+ * See [Instant.parse] for examples of instant string representations.
+ */
+public fun String.toInstantOrNull(): Instant? = try { Instant.parse(this) } catch(cause: IllegalArgumentException) { null }
```
svenjacobs, line 192: You could use [runCatching](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/run-catching.html) here unless you only want to explicitly catch `IllegalArgumentException`? It is available for multiplatform since Kotlin 1.3.

```suggestion
public fun String.toInstantOrNull(): Instant? = runCatching { Instant.parse(this) }.getOrNull()
```

hfhbd: I do want to only catch `IllegalArgumentException` because this exception is thrown by `java.time.Instant` when passing an invalid instant. Also runCatching creates a new class, `Result`, just to do the same thing. But I guess, it is not merged, because ideally it would be better to not use catch at all because it is still inefficient. Maybe checking the format with a regex first would be better.

svenjacobs: In that case let's keep the `try ... catch`. I doubt creating an instance of `Regex` and executing that regex is any more performant than `try ... catch` though.

dkhalanskyjb:
We decided to remove functions like `String.toInstant` at all: https://github.com/Kotlin/kotlinx-datetime/issues/339

Also, there are now `parseOrNull` functions directly on parsers. For example, `LocalDate.Formats.ISO.parseOrNull("2023-04-1") == null`.

PR #150, opened 2021-10-04
Work around a crash on older Android versions

### LINKED ISSUE #149: ArrayIndexOutOfBoundsException ZoneRules.isFixedOffset
After updating to `kotlinx-datetime:0.3.0` I am seeing crashes on Android with the following stacktrace:

>java.lang.ArrayIndexOutOfBoundsException: length=0; index=-1
    at j$.time.zone.ZoneRules.isFixedOffset
    at kotlinx.datetime.TimeZone$Companion.ofZone$kotlinx_datetime(SourceFile:47)
    at kotlinx.datetime.TimeZone$Companion.currentSystemDefault(SourceFile:34)

I am not totally sure where to report this, the crash is introduced due to `0.3.0` making calls to `ZoneRules.isFixedOffset`.
Though by looking at the source code of `java.time.zone.ZoneRules` [here](https://github.com/google/desugar_jdk_libs/blob/master/src/share/classes/java/time/zone/ZoneRules.java#L506), I can't see how an `ArrayIndexOutOfBoundsException` can happen.

I have `coreLibraryDesugaring` enabled, using `com.android.tools:desugar_jdk_libs:1.1.5`

dkhalanskyjb:
Fixes #149

### DIFF
--- a/core/jvm/src/TimeZoneJvm.kt
+++ b/core/jvm/src/TimeZoneJvm.kt
@@ -44,7 +44,7 @@ public actual open class TimeZone internal constructor(internal val zoneId: Zone
         internal fun ofZone(zoneId: ZoneId): TimeZone = when {
             zoneId is jtZoneOffset ->
                 FixedOffsetTimeZone(UtcOffset(zoneId))
-            zoneId.rules.isFixedOffset ->
+            zoneId.isFixedOffset ->
                 FixedOffsetTimeZone(UtcOffset(zoneId.normalized() as jtZoneOffset), zoneId)
             else ->
                 TimeZone(zoneId)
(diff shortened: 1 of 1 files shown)


core/jvm/src/TimeZoneJvm.kt
```diff
@@ -54,6 +54,15 @@ public actual open class TimeZone internal constructor(internal val zoneId: Zone
     }
 }
 
+// Workaround for https://github.com/Kotlin/kotlinx-datetime/issues/149
+private val ZoneId.isFixedOffset: Boolean
+    get() = try {
+        // On older Android versions, this can throw even though it shouldn't
```
ilya-g, line 60: Let's reference here a bug in the Android issue tracker so that we know when to remove the workaround.

PR #157, opened 2021-11-22
Improve the documentation

dkhalanskyjb:
Add the missing documentation.

Other changes:
* Update the Gradle version (the IDE refused to operate until I
  did this);
* Reformat some of the files;
* Remove some entities from the expect declarations; in particular,
  `MIN`/`MAX` values that weren't used in the common code other
  than in tests. For tests, these values are defined as
  expect/actual in the test module.
* Fix some warnings.

### DIFF
--- a/core/common/src/Clock.kt
+++ b/core/common/src/Clock.kt
@@ -7,10 +7,22 @@ package kotlinx.datetime
 
 import kotlin.time.*
 
+/**
+ * A source of [Instant] values.
+ *
+ * @see Clock.System for the instance that queries the operating system.
+ */
 public interface Clock {
+    /**
+     * The current time, according to this clock.
+     */
     public fun now(): Instant
 
+    /**
+     * The [Clock] instance that queries the operating system as its source of knowledge of time.
+     */
     public object System : Clock {
+        /** @suppress */
         override fun now(): Instant = @Suppress("DEPRECATION_ERROR") Instant.now()
     }
 
@@ -19,9 +31,15 @@ public interface Clock {
     }
 }
 
+/**
+ * Returns the current date at a given [time zone][timeZone], according to [this Clock][this].
+ */
 public fun Clock.todayAt(timeZone: TimeZone): LocalDate =
         now().toLocalDateTime(timeZone).date
 
+/**
+ * Returns the [TimeSource] that wraps the [Instant] values from [Clock.now] intro [TimeMark] instances.
+ */
 @ExperimentalTime
 public fun Clock.asTimeSource(): TimeSource = object : TimeSource {
     override fun markNow(): TimeMark = InstantTimeMark(now(), this@asTimeSource)
--- a/core/common/src/Clock.kt
+++ b/core/common/src/Clock.kt
@@ -10,11 +10,11 @@ import kotlin.time.*
 /**
  * A source of [Instant] values.
  *
- * @see Clock.System for the instance that queries the operating system.
+ * See [Clock.System][Clock.System] for the clock instance that queries the operating system.
  */
 public interface Clock {
     /**
-     * The current time, according to this clock.
+     * Returns the [Instant] corresponding to the current time, according to this clock.
      */
     public fun now(): Instant
 
@@ -22,7 +22,6 @@ public interface Clock {
      * The [Clock] instance that queries the operating system as its source of knowledge of time.
      */
     public object System : Clock {
-        /** @suppress */
         override fun now(): Instant = @Suppress("DEPRECATION_ERROR") Instant.now()
     }
 
@@ -32,13 +31,13 @@ public interface Clock {
 }
 
 /**
- * Returns the current date at a given [time zone][timeZone], according to [this Clock][this].
+ * Returns the current date at the given [time zone][timeZone], according to [this Clock][this].
  */
 public fun Clock.todayAt(timeZone: TimeZone): LocalDate =
         now().toLocalDateTime(timeZone).date
 
 /**
- * Returns the [TimeSource] that wraps the [Instant] values from [Clock.now] intro [TimeMark] instances.
+ * Returns a [TimeSource] that uses this [Clock] to mark a time instant and to find the amount of time elapsed since that mark.
  */
 @ExperimentalTime
 public fun Clock.asTimeSource(): TimeSource = object : TimeSource {
--- a/core/common/src/DateTimePeriod.kt
+++ b/core/common/src/DateTimePeriod.kt
@@ -12,11 +12,16 @@ import kotlin.time.Duration
 import kotlinx.serialization.Serializable
 
 /**
- * A difference between two [instants][Instant], decomposed into date and time units.
+ * A difference between two [instants][Instant], decomposed into date and time components.
  *
- * See [DatePeriod] for a special case that only stores date units.
- * A [DatePeriod] is automatically returned from all constructor functions for [DateTimePeriod] if it turns out that
- * the time components are zero.
+ * The date components are: [years], [months], [days].
+ *
+ * The time components are: [hours], [minutes], [seconds], [nanoseconds].
+ *
+ * A `DateTimePeriod` can be constructed using the same-named constructor function,
+ * [parsed][DateTimePeriod.parse] from a string, or returned as the result of instant arithmetic operations (see [Instant.periodUntil]).
+ * All these functions can return a [DatePeriod] value, which is a subtype of `DateTimePeriod`,
+ * a special case that only stores date components, if all time components of the result happen to be zero.
  */
 @Serializable(with = DateTimePeriodIso8601Serializer::class)
 // TODO: could be error-prone without explicitly named params
@@ -24,9 +29,9 @@ public sealed class DateTimePeriod {
     internal abstract val totalMonths: Int
 
     /**
-     * The number of days.
+     * The number of calendar days.
      *
-     * We do not consider a day to be equal to 24 hours. Please see [DateTimeUnit.DAY] for a discussion.
+     * Note that a calendar day is not identical to 24 hours, see [DateTimeUnit.DayBased] for details.
      */
     public abstract val days: Int
     internal abstract val totalNanoseconds: Long
@@ -302,6 +307,15 @@ public sealed class DateTimePeriod {
  */
 public fun String.toDateTimePeriod(): DateTimePeriod = DateTimePeriod.parse(this)
 
+/**
+ * A special case of [DateTimePeriod] that only stores date components and has all time components equal to zero.
+ *
+ * A `DatePeriod` is automatically returned from all constructor functions for [DateTimePeriod] if it turns out that
+ * the time components are zero.
+ *
+ * `DatePeriod` values are used in operations on [LocalDates][LocalDate] and are returned from operations on [LocalDates][LocalDate],
+ * but they also can be passed anywhere where a [DateTimePeriod] is expected.
+ */
 @Serializable(with = DatePeriodIso8601Serializer::class)
 public class DatePeriod internal constructor(
     internal override val totalMonths: Int,
@@ -309,16 +323,16 @@ public class DatePeriod internal constructor(
 ) : DateTimePeriod() {
     public constructor(years: Int = 0, months: Int = 0, days: Int = 0): this(totalMonths(years, months), days)
     // avoiding excessive computations
-    /** The constant 0. */
+    /** The number of whole hours in this period. Always equal to zero. */
     override val hours: Int get() = 0
 
-    /** The constant 0. */
+    /** The number of whole minutes in this period. Always equal to zero. */
     override val minutes: Int get() = 0
 
-    /** The constant 0. */
+    /** The number of whole seconds in this period. Always equal to zero. */
     override val seconds: Int get() = 0
 
-    /** The constant 0. */
+    /** The number of nanoseconds in this period. Always equal to zero. */
     override val nanoseconds: Int get() = 0
     internal override val totalNanoseconds: Long get() = 0
 
@@ -417,7 +431,7 @@ public fun DateTimePeriod(
 /**
  * Constructs a [DateTimePeriod] from a [Duration].
  *
- * If the duration is too long to be represented as a [Long] number of nanoseconds,
+ * If the duration value is too big to be represented as a [Long] number of nanoseconds,
  * the result will be [Long.MAX_VALUE] nanoseconds.
  */
 // TODO: maybe it's more consistent to throw here on overflow?
--- a/core/common/src/DateTimePeriod.kt
+++ b/core/common/src/DateTimePeriod.kt
@@ -314,7 +314,7 @@ public fun String.toDateTimePeriod(): DateTimePeriod = DateTimePeriod.parse(this
  * the time components are zero.
  *
  * `DatePeriod` values are used in operations on [LocalDates][LocalDate] and are returned from operations on [LocalDates][LocalDate],
- * but they also can be passed anywhere where a [DateTimePeriod] is expected.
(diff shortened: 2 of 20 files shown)


ilya-g:
Could you please extract changes not related to documentation, such as refactorings or build tooling upgrade to separate PRs? That would ease reviewing.

dkhalanskyjb:
Ok

core/common/src/serializers/MonthSerializers.kt
```diff
@@ -12,7 +12,7 @@ import kotlinx.serialization.encoding.*
 import kotlinx.serialization.internal.*
 
 /**
- * A serializer for [Month] that represents the values as strings.
+ * A serializer for [Month] that represents the month value as a string.
```
dkhalanskyjb, line 15: It is correct, but I don't see the need for redundancy. What other values could the serializer operate on? So, I'd suggest either "a value", or "the values", which I still prefer.

core/common/src/serializers/LocalDateTimeSerializers.kt
```diff
@@ -33,7 +33,7 @@ public object LocalDateTimeIso8601Serializer: KSerializer<LocalDateTime> {
 }
 
 /**
- * A serializer for [LocalDateTime] that represents the values as their components.
+ * A serializer for [LocalDateTime] that represents the value as its components.
```
dkhalanskyjb, line 36: > the value

I believe it's incorrect as there's no implied value here. In "the values", `the` represents the implied `LocalDateTime`, so is equivalent to "the LocalDateTime values". It would be better to either keep it as it were, or use "a value".

core/common/src/serializers/InstantSerializers.kt
```diff
@@ -33,7 +33,7 @@ public object InstantIso8601Serializer: KSerializer<Instant> {
 }
 
 /**
- * A serializer for [Instant] that represents the values as second and nanosecond component of the Unix time.
+ * A serializer for [Instant] that represents an `Instant` value as second and nanosecond components of the Unix time.
```
dkhalanskyjb, line 36: This is fine, though I don't see anything wrong with the initial proposal.

core/common/src/serializers/DateTimePeriodSerializers.kt
```diff
@@ -95,7 +95,7 @@ public object DateTimePeriodIso8601Serializer: KSerializer<DateTimePeriod> {
 /**
  * A serializer for [DatePeriod] that uses a different field for each component, only listing non-zero components.
  *
- * Deserializes the time components as well, as long as they are zero.
+ * Deserializes the time components as well ensuring they are zero.
```
dkhalanskyjb, line 98: I think this obfuscates the fact that the time components are optional. Instead, it looks like the serializer makes sure that they are present and are zero.

core/common/src/Clock.kt
```diff
@@ -10,19 +10,18 @@ import kotlin.time.*
 /**
  * A source of [Instant] values.
  *
- * @see Clock.System for the instance that queries the operating system.
+ * See [Clock.System][Clock.System] for the clock instance that queries the operating system.
```
dkhalanskyjb, line 13: How is `[Clock.System][Clock.System]` different from `[Clock.System]`?

core/common/src/DateTimePeriod.kt
```diff
@@ -302,23 +307,32 @@ public sealed class DateTimePeriod {
...
 public fun String.toDateTimePeriod(): DateTimePeriod = DateTimePeriod.parse(this)
 
+/**
+ * A special case of [DateTimePeriod] that only stores date components and has all time components equal to zero.
+ *
+ * A `DatePeriod` is automatically returned from all constructor functions for [DateTimePeriod] if it turns out that
+ * the time components are zero.
+ *
+ * `DatePeriod` values are used in operations on [LocalDates][LocalDate] and are returned from operations on [LocalDates][LocalDate],
+ * but they also can be passed anywhere where a [DateTimePeriod] is expected.
```
dkhalanskyjb, line 317: > `anywhere where`

Sounds a bit redundant compared to just `anywhere`.

core/common/src/DateTimeUnit.kt
```diff
@@ -14,24 +14,26 @@ import kotlin.time.Duration.Companion.nanoseconds
...
  * - By constructing an instance manually with [TimeBased], [DayBased], or [MonthBased]: for example,
  *   `TimeBased(nanoseconds = 10)`.
  *
- * Note that a day is not considered equal to 24 hours. See [DateTimeUnit.DAY] for a discussion.
+ * Note that a calendar day is not considered identical to 24 hours. See [DateTimeUnit.DayBased] for a discussion.
  */
 @Serializable(with = DateTimeUnitSerializer::class)
 public sealed class DateTimeUnit {
 
+    /** Produces a date-time unit that is a multiple of this unit by the specified integer [scalar] value. */
```
dkhalanskyjb, line 29: How about
```
Produces a date-time unit that is [scalar] times longer than this unit.
```

core/common/src/DateTimeUnit.kt
```diff
@@ -91,10 +93,11 @@ public sealed class DateTimeUnit {
     }
 
     /**
-     * The date-time unit equal to some number of days or months.
+     * A date-time unit equal to some number of days or months.
      *
-     * Unless used for operations on dates, measuring with such units requires knowledge of the time zone, as it is
-     * otherwise unknown when one date ends and another one begins.
+     * Operations involving `DateBased` units are performed with dates. The same operations on [Instants][Instant]
```
dkhalanskyjb, line 98: > `with dates`

`on dates`?

core/common/src/DateTimeUnit.kt
```diff
@@ -91,10 +93,11 @@ public sealed class DateTimeUnit {
...
+     * require a [TimeZone] to find the corresponding [LocalDateTimes][LocalDateTime] first and then to perform
```
dkhalanskyjb, line 99: > `require a [TimeZone] to find the corresponding [LocalDateTimes][LocalDateTime] first and then to perform`
`require a [TimeZone] to find the corresponding [LocalDateTimes][LocalDateTime] first to perform`

Otherwise, it looks like the time zone is also needed for the latter.

core/common/src/DateTimeUnit.kt
```diff
@@ -199,23 +209,14 @@ public sealed class DateTimeUnit {
...
          *
-         * We follow ISO-8601 in that we do not consider a day to be the same as 24 hours.
-         * The reason lies in time zone transitions, because of which some days can be 23 or 25 hours.
-         * For example, we say that exactly a whole day has passed between `2019-10-27T02:59` and `2019-10-28T02:59`
-         * in Berlin, despite the fact that the clocks were turned back one hour, so there are, in fact, 25 hours
-         * between the two date-times.
-         *
-         * It is for this reason that most operations with [DateBased] units require a [TimeZone], whereas those with
-         * [TimeBased] ones never do.
+         * Note that a calendar day is not the same as 24 hours.
```
dkhalanskyjb, line 214: The reason why I placed the documentation here and not on `DayBased` is that I'd expect most people to never encounter `DayBased` and instead use `DAY` or its multiples. So, I think that the description of the calendar day should be at least referenced here.

core/common/src/DateTimeUnit.kt
```diff
@@ -199,23 +209,14 @@ public sealed class DateTimeUnit {
...
          */
         public val DAY: DayBased = DayBased(days = 1)
 
         /**
-         * 7 days.
-         *
-         * It can not represented as some number of hours. Please see [DAY] for a discussion.
+         * A week, which is 7 calendar days.
```
dkhalanskyjb, line 219: Ditto.

core/common/src/DayOfWeek.kt
```diff
@@ -29,7 +29,7 @@ public val DayOfWeek.isoDayNumber: Int get() = ordinal + 1
 private val allDaysOfWeek = DayOfWeek.values().asList()
 
 /**
- * Returns the [DayOfWeek] instance for the given ISO-8601 week number. January is 1, December is 12.
+ * Returns the [DayOfWeek] instance for the given ISO-8601 week day number. Monday is 1, Sunday is 7.
```
dkhalanskyjb, line 32: :+1:

core/common/src/Instant.kt
```diff
@@ -17,7 +17,11 @@ import kotlin.time.*
  * in different time zones: someone in Tokyo would think its already `1970-01-01` several hours earlier than someone in
  * Berlin would. To represent such entities, use [LocalDateTime].
  * In contrast, "the moment the clocks in London first showed 00:00 on Jan 1, 2000" is a specific moment
- * in time, as is "1970-01-01, 00:00:00 UTC", and so can be represented as an [Instant].
+ * in time, as is "1970-01-01, 00:00:00 UTC+0", and so it can be represented as an [Instant].
+ *
+ * `Instant` uses UTC-SLS (smeared leap second) time scale. This time scale doesn't contain instants
```
dkhalanskyjb, line 22: > uses the UTC-SLS

core/common/src/Instant.kt
```diff
@@ -17,7 +17,11 @@ import kotlin.time.*
...
+ * corresponding to leap seconds, but instead "smears" positive and negative leap seconds among the last 1000 seconds
+ * of the day when a leap second happens.
```
dkhalanskyjb, line 24: This whole section, I think, is out-of-place here, as the implementation of `Instant` does not reflect the SLS aspect.

Additionally, it should be qualified with "on the JVM, if the implementation permits it", or, better yet, simply removed, both from here and from the README.
On the JS, just `Date` is used: https://github.com/js-joda/js-joda/blob/0702890538ab460a446aba89622878fb70fc94e2/packages/core/src/Clock.js#L218
Searching for the JVM implementation of UTC-SLS, I only find mentions of it being specified, but not supported:  https://stackoverflow.com/a/35465689 https://stackoverflow.com/a/31162603
On the Native, well, there's nothing in the vein of UTC-SLS either.

core/common/src/TimeZone.kt
```diff
@@ -12,7 +12,7 @@ import kotlinx.datetime.serializers.*
 import kotlinx.serialization.Serializable
 
 /**
- * A time zone, which is a collection of rules specifying which [LocalDateTime] values correspond to each [Instant].
+ * A time zone, which is a collection of rules specifying how to convert an [Instant] to a [LocalDateTime] and vice versa.
```
dkhalanskyjb, line 15: "Vice versa" is not accurate if we assume that convertation is a function, which seems intuitive.

core/common/src/UtcOffset.kt
```diff
@@ -44,26 +51,26 @@ public expect class UtcOffset {
...
  *
  * All components must have the same sign.
  *
  * The bounds are checked: it is invalid to pass something other than `±[0; 59]` as the number of seconds or minutes.
- * For example, `UtcOffset(hours = 03, minutes = 61)` is invalid.
+ * For example, `UtcOffset(hours = 3, minutes = 61)` is invalid.
  *
- * However, if a component is the first non-null one, it can exceed these bounds.
- * So, for example, `UtcOffset(minutes = 241)` is valid.
+ * However, the first non-null component of highest order can exceed these bounds,
```
dkhalanskyjb, line 61: > `the first non-null component of highest order`

Maybe just `the non-null component of the highest order` or `the first non-null component`?

core/common/src/Clock.kt
ilya-g: Can be rendered as just `System` by dokka. So it is to ensure the exact text of the link label.

core/common/src/Instant.kt
ilya-g: It's the time scale we have chosen for Instants. The section answers the question "where are my leap seconds". The fact that the platform clocks don't provide this behavior can be explained by their inaccuracy.

core/common/src/DateTimeUnit.kt
ilya-g: Applied

core/common/src/DateTimeUnit.kt
ilya-g: I've added a link from DAY to DayBased.

core/common/src/UtcOffset.kt
ilya-g: Chosen the former.

core/common/src/TimeZone.kt
dkhalanskyjb: My initial proposal:
> `A time zone, which is a collection of rules specifying which [LocalDateTime] values correspond to each [Instant].`

this doesn't say anything about which conversions are available, it only gives the intuition behind the data structure. We can also make a separate statement, like
```
[Instant.toLocalDateTime] is an unambiguous conversion according to these rules.
[LocalDateTime.toInstant] is a best-effort attempt at reversing that conversion.
```

ilya-g: Currently, all that TimeZone exposes is the pair of conversion functions, not the collection of rules.

"which [LocalDateTime] values correspond to each [Instant]" implies that there could be several LDT values for a single Instant in the given TZ, which was my concern with it originally.

dkhalanskyjb: Ah, ok. How about this then? `which [LocalDateTime] value correspond to each [Instant]`

core/common/src/Instant.kt
dkhalanskyjb: > length of the ISO-8601 calendar day (3.1.2.11) depends on the leap seconds, so we can't even claim that `DateTimePeriod` corresponds to some ISO-8601 entity or a combination of them.

Now that I think of it, it seems fine. So, only the `Instant` parsing/formatting is problematic.

ilya-g: I don't think ISO mandates the use of UTC scale with leap seconds. As usual this is handled by an agreement between the interchange parties.

dkhalanskyjb: Seems fairly conclusive to me.

3.1.1.13. **UTC of day** is the time of day *in UTC*.
3.2.5. "Z" indicates that a time of day is represented as *UTC of day*.

ilya-g: Again, I believe it is possible to use ISO format to interchange UTC-SLS instants without converting them first to UTC. It just has to be agreed upon, and the appropriate place to state that agreement would indeed be where we parse and format values from and to ISO string representations.

dkhalanskyjb: Could you point to a particular section where this is permitted? I traced `"Z"` meaning exactly UTC to the base definitions, but maybe there are some rules overriding this that I missed.

ilya-g: No, that's my interpretation of it.

dkhalanskyjb: Sure, but what part of the standard is open to interpretation?

dkhalanskyjb: Historical note: after an internal discussion, we decided that this is fine, being the de-facto way the ISO-8601 strings are typically used.

core/common/src/TimeZone.kt
ilya-g: Ok, that would do. Rephrased to a middle ground between these two variants.

ilya-g:
Since the last commit introduces behavior changes let's leave it out of this PR and discuss separately. Actually, I expected `Instant.parse` to handle leap second values as it does in Java.

dkhalanskyjb:
I amended the commit not to change the behavior. I should note though that the behavior change only affected Native, the other platforms already fail on leap seconds in `Instant.parse`.

ilya-g (approved):

PR #159, opened 2021-12-05
Bump Kotlin to 1.6.0 and remove ExperimentalTime

hfhbd:
Fixes #156

PR #161, opened 2021-12-10
Bump Kotlin to 1.6.0 and remove ExperimentalTime

### LINKED ISSUE #156: Remove experimental opt in for durations (for kotlin 1.6)
Hello,
since kotlin 1.6 has been released duration is no longer experimental. Are there plans to release a version for this version of kotlin without the optional annotation ?

dkhalanskyjb:
Fixes #156
Supersedes https://github.com/Kotlin/kotlinx-datetime/pull/159

### DIFF
--- a/core/jvm/src/Instant.kt
+++ b/core/jvm/src/Instant.kt
@@ -12,12 +12,13 @@ import java.time.DateTimeException
 import java.time.format.DateTimeParseException
 import java.time.temporal.ChronoUnit
 import kotlin.time.*
+import kotlin.time.Duration.Companion.nanoseconds
+import kotlin.time.Duration.Companion.seconds
 import java.time.Instant as jtInstant
 import java.time.OffsetDateTime as jtOffsetDateTime
 import java.time.Clock as jtClock
 
 @Serializable(with = InstantIso8601Serializer::class)
-@OptIn(ExperimentalTime::class)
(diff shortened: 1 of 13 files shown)


core/jvm/src/Instant.kt
```diff
@@ -43,8 +44,8 @@ public actual class Instant internal constructor(internal val value: jtInstant)
     public actual operator fun minus(duration: Duration): Instant = plus(-duration)
 
     public actual operator fun minus(other: Instant): Duration =
-            Duration.seconds(this.value.epochSecond - other.value.epochSecond) + // won't overflow given the instant bounds
-            Duration.nanoseconds(this.value.nano - other.value.nano)
+        (this.value.epochSecond - other.value.epochSecond).seconds + // won't overflow given the instant bounds
```
ilya-g, line 47: Minor: I'd probably use in such cases a more verbose `.toDuration(DurationUnit.SECONDS)` to sound less tautological.

ilya-g (approved):

core/jvm/src/Instant.kt
dkhalanskyjb: Tried this and didn't like the result, looked to me as much more cumbersome than both `Duration.seconds(...)` and `.seconds`, so the intent is obscured. `.seconds` is not pretty, but very obvious.

PR #166, opened 2021-12-16
Gradle upgrade 7.3.3

SampathKumarAmex:
https://docs.gradle.org/7.3.3/release-notes.html

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.2-bin.zip
-distributionSha256Sum=23b89f8eac363f5f4b8336e0530c7295c55b728a9caa5268fdd4a532610d5392
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
+distributionSha256Sum=b586e04868a22fd817c8971330fec37e298f3242eb85c374181b12d637f80302
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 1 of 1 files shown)


ilya-g:
Hi, do you have any specific reasons why we should update to that particular version of Gradle?

JakeWharton:
7.3.3 is latest, FYI

SampathKumarAmex:
> 7.3.3 is latest, FYI

Yes, agreed. The new version 7.3.3 released on 22nd Dec which has the fix for log4j security issues.

https://docs.gradle.org/7.3.3/release-notes.html

SampathKumarAmex:
> Hi, do you have any specific reasons why we should update to that particular version of Gradle?

No specific reason. When I raised the PR, `7.3.2` was the latest version. I've updated to latest version. Thanks 👍🏻.

svenjacobs:
I'm not able to build this project on an MacBook with Apple Silicon chip:

```
Unable to download toolchain matching these requirements: {languageVersion=8, vendor=any, implementation=vendor-specific}
> Unable to download toolchain. This might indicate that the combination (version, architecture, release/early access, ...) for the requested JDK is not available.
   > Could not read 'https://api.adoptopenjdk.net/v3/binary/latest/8/ga/mac/aarch64/jdk/hotspot/normal/adoptopenjdk' as it does not exist.
```

Also see this similar [OkHttp issue](https://github.com/square/okhttp/issues/6943). AdoptOpenJDK ist deprecated. Upgrading this project to the current Gradle version (which is 7.5.1) would be a huge step to fixing this.

Goooler:
> Also see this similar https://github.com/square/okhttp/issues/6943. AdoptOpenJDK ist deprecated. Upgrading this project to the current Gradle version (which is 7.5.1) would be a huge step to fixing this.

We can wait for Gradle 7.6 which supports Zulu vendors.

dkhalanskyjb:
Superseded by https://github.com/Kotlin/kotlinx-datetime/pull/234

PR #171, opened 2021-12-28
Add js.Date converter

### LINKED ISSUE #170: JS: Add Converter to Date
On JVM you can use `Instant.toJavaInstant()` etc, while on JS there are no converters at all

Usecase:
1. Display localized string of an `Instant`
2. Use the `Date` with other JS libraries

hfhbd:
Fixes #170


core/js/src/Converters.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import kotlin.js.*
+
+/**
+ * Converts the [Instant] to an instance of [Date].
+ *
+ * The conversion is lossy: JS uses millisecond precision to represent dates, and [Instant] allows for nanosecond
+ * resolution.
+ */
+public fun Instant.toJSDate(): Date = Date(milliseconds = toEpochMilliseconds())
```
ilya-g, line 16: Better to convert to `Double` (JS Number) explicitly to be safe from potential problems like https://youtrack.jetbrains.com/issue/KT-50657

core/js/src/Converters.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright 2019-2021 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.datetime
+
+import kotlin.js.*
+
+/**
+ * Converts the [Instant] to an instance of [Date].
```
ilya-g, line 11:
```suggestion
 * Converts the [Instant] to an instance of JS [Date].
```

core/js/src/Converters.kt
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * The conversion is lossy: JS uses millisecond precision to represent dates, and [Instant] allows for nanosecond
+ * resolution.
+ */
+public fun Instant.toJSDate(): Date = Date(milliseconds = toEpochMilliseconds())
+
+/**
+ * Converts the [Date] to the corresponding [Instant].
```
ilya-g, line 19:
```suggestion
 * Converts the JS [Date] to the corresponding [Instant].
```

ilya-g (changes requested):

core/js/src/Converters.kt
hfhbd: Fixed by using `toEpochMilliseconds().toDouble()`.

ilya-g (approved):

PR #172, opened 2021-12-29
Ignore local.properties

wkornewald:
That file is auto-created by Android Studio and should never be committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,3 @@
 target
 build
 /local.properties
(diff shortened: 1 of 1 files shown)


ilya-g (approved):

ilya-g:
Could you also cleanup the last three entries? They seem to get here by mistake 😅

wkornewald:
Sure. Does that look right now? :)

ilya-g (approved):

Issue #96, opened 2021-01-31
Don't typealias Month (and others)


fluidsonic:
I suggest to not typealias anything in a multiplatform date/time library.

- It mixes platform-provided API into the library's API which causes confusion and is platform-dependent.
- It doesn't allow for adding a companion and thus "static" functionality - not even through extensions.

For example:
I had code in the back-end that used `Month.of(someInt)`. The code was relatively simple and there were no JVM-specific  imports. So I've moved it to a shared multiplatform dependency and was surprised that `Month.of()` was no longer available. `Month.<autocomplete>` didn't provide me with any useful alternative.

After Cmd-clicking on `Month` I've learned that it is merely a typealias on JVM. It took a while to figure out that there's an additional `Month(Int)` function in kotlinx-datetime that I should've used instead. Scenarios like that can easily lead to confusion and inconsistency.

september669:
Another example for Android 

```
import android.content.Context
import kotlinx.datetime.DayOfWeek

fun DayOfWeek.toString(context: Context): String{
    return when(this){
        java.time.DayOfWeek.MONDAY -> TODO() <---- Field requires API Level 26
        java.time.DayOfWeek.TUESDAY -> TODO()
        java.time.DayOfWeek.WEDNESDAY -> TODO()
        java.time.DayOfWeek.THURSDAY -> TODO()
        java.time.DayOfWeek.FRIDAY -> TODO()
        java.time.DayOfWeek.SATURDAY -> TODO()
        java.time.DayOfWeek.SUNDAY -> TODO()
    }
}
```

JakeWharton:
@september669 the whole library requires coreLibraryDesugaring to be enabled enabling the use of java.time on previous API levels.

september669:
Thanks a lot! 

> @september669 the whole library requires coreLibraryDesugaring to be enabled enabling the use of java.time on previous API levels.

Thanks a lot!

PaulWoitaschek:
I actually like the typealiasing as when we further process the data in the android project, we'd need need to create a wrapper that does the mapping.
We actually use a `expect typealias LocalDate` function to expose public properties coming from our library surface and `actual` implement it as javas LocalDate on kotlin and `NSDateComponents` for ios.

The argument with the factories will go away when the namespace feature arrives.

fluidsonic:
With Kotlin 1.5 it's gotten super annoying and really needs to be changed.

[...]

PR #174, opened 2021-12-29
WIP: Don't typealias Month

wkornewald:
This is just one example from the codebase. Why are so many APIs `expect` and why do we even try to use `typealias` at all? This makes it more difficult to define a nice API for Kotlin and it causes lots of code repetition. It's even impossible to use `sealed interface/class` with `expect/actual` because the multiplatform code lives in different modules (see #173, #177  and #175 for how this can be useful).

I think it would be much better to separate the `expect` APIs and make them `internal` and expose a nice and clear non-expect based `public` API (which can call the `internal expect` API inside).

What do you think?

### DIFF
--- a/core/common/src/Month.kt
+++ b/core/common/src/Month.kt
@@ -7,7 +7,7 @@ package kotlinx.datetime
 
 import kotlin.native.concurrent.*
 
-public expect enum class Month {
+public enum class Month {
     JANUARY,
     FEBRUARY,
     MARCH,
@@ -21,10 +21,8 @@ public expect enum class Month {
     NOVEMBER,
     DECEMBER;
 
-//    val value: Int // member missing in java.time.Month has to be an extension
+    public val number: Int get() = ordinal + 1
 }
-
-public val Month.number: Int get() = ordinal + 1
 
 @SharedImmutable
 private val allMonths = Month.values().asList()
--- a/core/js/src/Month.kt
+++ b/core/js/src/Month.kt
@@ -7,19 +7,4 @@ package kotlinx.datetime
 
 import kotlinx.datetime.internal.JSJoda.Month as jsMonth
 
-public actual enum class Month {
-    JANUARY,
-    FEBRUARY,
-    MARCH,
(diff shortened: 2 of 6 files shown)


wkornewald:
@dkhalanskyjb Closing this PR. Is it even worth discussing in a separate issue? This could be a general architectural discussion and the Month typealias is just one aspect of the way expect/actual is used.

dkhalanskyjb:
There's an issue for it already: https://github.com/Kotlin/kotlinx-datetime/issues/96

wkornewald:
Not quite. That only focuses on the Month typealias. My suggestion is to generally apply the rule: avoid `expect` + `public` because leads to problems like `sealed` not working  (maybe the only exception is `expect public fun` which might be always ok?). Try to use `expect` only with `private` and `internal`.

dkhalanskyjb:
The library is fairly compact, so its internal architecture can be changed on a whim once we need it and does not matter much. What matters is the API that we provide to the programmers, and they typically shouldn't care whether something is `expect` or not.

`actual typealias` has actual issues on that front: https://github.com/Kotlin/kotlinx-datetime/issues/96#issuecomment-847921926. Do you see any potential issues for the programmers because of, say, `LocalDate` being `expect`?

wkornewald:
> The library is fairly compact, so its internal architecture can be changed on a whim once we need it and does not matter much. What matters is the API that we provide to the programmers, and they typically shouldn't care whether something is `expect` or not.
> 
> `actual typealias` has actual issues on that front: [#96 (comment)](https://github.com/Kotlin/kotlinx-datetime/issues/96#issuecomment-847921926). Do you see any potential issues for the programmers because of, say, `LocalDate` being `expect`?

I don't think there's any similarly fatal issue. The other issue is that this whole library can't use `sealed` which limits the API design in general and indirectly affects how comfortably programmers can work with this lib. At least one can't easily provide a `FlexibleDate`/`FlexibleDateTime` in this lib and I find that unfortunate because we keep reinventing the wheel.

Issue #163, opened 2021-12-15
Add a ZonedDateTime type


cromefire:
Add a type that aligns with the normal calendar and combines TumeZone information with a date.

cromefire:
@dkhalanskyjb 
> > The same argument can be made for date + time.
> 
> It can't because 1) waiting for an hour at 23:31 will start the next day 2) the set of valid times depends on the date.

1. It's just a PITA to have APIs using `Pair`.
2. How would you e.g. fix the offset (`toOffsetDateTime()`) or get an `Instant` from that pair?
3. How would you handle parsing and serialization?

dkhalanskyjb:
> It's just a PITA to have APIs using `Pair`.

True. However, there's a difference between introducing an `OffsetDateTime` or a `ZonedDateTime` with their whole semantics and behavior, and adding a small wrapper that's almost like a pair, but with a couple of helper functions. Therefore, we need to decide which path to take. A small wrapper seems like a good approach, but if there are other use cases for `ZonedDateTime`, we may just introduce it instead.

> How would you e.g. fix the offset (toOffsetDateTime()) or get an Instant from that pair?

I'm not sure what "fixing an offset" means, but to get an `Instant` from a `Pair<LocalDateTime, UtcOffset>`, you can do the following:
```kotlin
    fun ldtAndOffsetToInstant(pair: Pair<LocalDateTime, UtcOffset>): Instant {
        val (ldt, offset) = pair
        return ldt.toInstant(offset)
    }
```

> How would you handle parsing and serialization?

At the moment, we don't provide means to handle this.

cromefire:
> True. However, there's a difference between introducing an OffsetDateTime or a ZonedDateTime with their whole semantics and behavior, and adding a small wrapper that's almost like a pair, but with a couple of helper functions.

The helper functions definite the semantics of the class, so I don't see a difference here.

> I'm not sure what "fixing an offset" means

Convert from a TimeZone (dynamic-ish offset) to a fixed offset.

> At the moment, we don't provide means to handle this.

I'd define this as a goal though.

dkhalanskyjb:
> The helper functions definite the semantics of the class, so I don't see a difference here.

[...]

PR #175, opened 2021-12-29
WIP: Added ZonedDateTime

wkornewald:
Initial sketch of

* `ZoneDateTime`: common base class
* `RegionDateTime`: works for past and future dates and mimics how humans see dates
* `OffsetDateTime`: how many official specifications represent past/present dates

Also see discussion in #163 

Feedback welcome.

### DIFF
--- a/core/common/src/ZonedDateTime.kt
+++ b/core/common/src/ZonedDateTime.kt
@@ -1,0 +1,99 @@
+package kotlinx.datetime
+
+/** A timezone-aware date-time object. */
+public sealed class ZonedDateTime(
+    protected val localDateTime: LocalDateTime,
+) : Comparable<ZonedDateTime> {
+
+    public abstract val timeZone: TimeZone
+
+    // XXX: If the underlying time zone database can change while the current process is running
+    // this value could become incorrect. Maybe don't cache at all? Or detect time zone db changes?
+    private val instant: Instant by lazy { localDateTime.toInstant(timeZone) }
+
+    public val year: Int get() = localDateTime.year
+    public val monthNumber: Int get() = localDateTime.monthNumber
+    public val month: Month get() = localDateTime.month
+    public val dayOfMonth: Int get() = localDateTime.dayOfMonth
+    public val dayOfWeek: DayOfWeek get() = localDateTime.dayOfWeek
+    public val dayOfYear: Int get() = localDateTime.dayOfYear
+    public val hour: Int get() = localDateTime.hour
+    public val minute: Int get() = localDateTime.minute
+    public val second: Int get() = localDateTime.second
+    public val nanosecond: Int get() = localDateTime.nanosecond
+
+    public fun toInstant(): Instant = instant
+
+    public fun toLocalDateTime(): LocalDateTime = localDateTime
+
+    public fun toLocalDateTime(timeZone: TimeZone): LocalDateTime =
+        toInstant().toLocalDateTime(timeZone)
+
+    public fun toLocalDate(): LocalDate = toLocalDateTime().date
+
+    public fun toLocalDate(timeZone: TimeZone): LocalDate = toLocalDateTime(timeZone).date
+
+    override fun compareTo(other: ZonedDateTime): Int = toInstant().compareTo(other.toInstant())
+
+    override fun equals(other: Any?): Boolean =
+        this === other || (other is ZonedDateTime && compareTo(other) == 0)
+
+    override fun hashCode(): Int = localDateTime.hashCode() xor timeZone.hashCode()
+
+    public companion object {
+
+        public fun parse(isoString: String): ZonedDateTime {
+            TODO()
+        }
+    }
+}
+
+/** Constructs a new [ZonedDateTime] from the given [localDateTime] and [timeZone]. */
+public fun ZonedDateTime(localDateTime: LocalDateTime, timeZone: TimeZone): ZonedDateTime =
+    when (timeZone) {
+        is FixedOffsetTimeZone -> OffsetDateTime(localDateTime, timeZone)
+        // TODO: Define a common RegionTimeZone and make TimeZone a sealed class/interface
+        else -> RegionDateTime(localDateTime, timeZone)
+    }
+
+public fun String.toZonedDateTime(): ZonedDateTime = ZonedDateTime.parse(this)
+
+/**
+ * A [ZonedDateTime] describing a region-based [TimeZone].
+ *
+ * This class tries to represent how humans think in terms of dates.
+ * For example, adding one day will result in the same local time even if a DST change happens
+ * within that day.
+ * Also, you can safely represent future dates because time zone database changes are taken into
+ * account.
+ */
+public class RegionDateTime(
+    localDateTime: LocalDateTime,
+    // TODO: this should be a RegionTimeZone
+    override val timeZone: TimeZone,
+    // TODO: Add optional DST offset or at least a UTC offset (should it be part of RegionTimeZone?)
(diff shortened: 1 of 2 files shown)


dkhalanskyjb:
* As defined, `ZonedDateTime` is just a collection of functions that are very easy to implement and doesn't offer much over just storing `LocalDateTime` and `TimeZone` separately. If something is this easy to implement and doesn't meaningfully interact with the other entities in the library, there's little point in adding it to the base library.
* It's unclear how `OffsetDateTime` and `RegionDateTime` could be useful. When is it important to know what exactly does the `ZonedDateTime` store? If you need it for checking whether the time zone is fixed-offset, why not check the time zones themselves?
* `OffsetDateTime` is probably not the best name for this, as it stores a `FixedOffsetTimeZone` and not an offset.
* As you yourself note, `instant` is ill-defined, though for more problematic reasons: `LocalDateTime` to `TimeZone` conversion is ambiguous, which `ZonedDateTime` hides.

wkornewald:
> * As defined, `ZonedDateTime` is just a collection of functions that are very easy to implement and doesn't offer much over just storing `LocalDateTime` and `TimeZone` separately. If something is this easy to implement and doesn't meaningfully interact with the other entities in the library, there's little point in adding it to the base library.
> 

1. We can serialize/deserialize, format/parse, sort and compare dates very easily. It's impossible if we store those separately.
2. This is a draft/WIP and I wanted to add support for adding/subtracting durations and other operations which map to how humans think about dates.

>     * It's unclear how `OffsetDateTime` and `RegionDateTime` could be useful. When is it important to know what exactly does the `ZonedDateTime` store? If you need it for checking whether the time zone is fixed-offset, why not check the time zones themselves?
> 

When you want to use static types to enforce minimum requirements. We might want to ensure we store non-ambiguous `RegionDateTime` in the backend and using static types prevents mistakes.

>     * `OffsetDateTime` is probably not the best name for this, as it stores a `FixedOffsetTimeZone` and not an offset.
> 

That's just the name other APIs (e.g. java.time) use, but we can rename it to `FixedOffsetDateTime` (though, IMHO `OffsetTimeZone` would've been a nicer name without the `Fixed` prefix).

>     * As you yourself note, `instant` is ill-defined, though for more problematic reasons: `LocalDateTime` to `TimeZone` conversion is ambiguous, which `ZonedDateTime` hides.

1. It's not ambiguous with `OffsetDateTime`. It's also not ambiguous with `RegionDateTime` when adding the optional offset (see the TODO) like java.time and the new JavaScript Temporal API do.
2. It's still the way humans think about time, so even if it's ambiguous, we need an API for it.

cromefire:
> 1. It's also not ambiguous with `RegionDateTime` when adding the optional offset

In which case it would just be a OffsetDateTime right? And all the calendar usage traits would be gone.

core/common/src/ZonedDateTime.kt
```diff
@@ -0,0 +1,99 @@
...
+ * For example, adding one day will result in the same local time even if a DST change happens
+ * within that day.
+ * Also, you can safely represent future dates because time zone database changes are taken into
+ * account.
+ */
+public class RegionDateTime(
+    localDateTime: LocalDateTime,
+    // TODO: this should be a RegionTimeZone
+    override val timeZone: TimeZone,
+    // TODO: Add optional DST offset or at least a UTC offset (should it be part of RegionTimeZone?)
```
cromefire, line 74: No it should not, because it invalidates the reason for it's existence.

core/common/src/ZonedDateTime.kt
```diff
@@ -0,0 +1,99 @@
...
+ */
+public class OffsetDateTime(
+    localDateTime: LocalDateTime,
+    override val timeZone: FixedOffsetTimeZone,
+) : ZonedDateTime(localDateTime) {
+
+    public constructor(instant: Instant, timeZone: FixedOffsetTimeZone) :
+            this(instant.toLocalDateTime(timeZone), timeZone)
+
+    override fun toString(): String = "$localDateTime$timeZone"
```
cromefire, line 98: This should probably be in RFC3339 as that's the common serialization and representation everywhere.

wkornewald:
> > 1. It's also not ambiguous with `RegionDateTime` when adding the optional offset
> 
> In which case it would just be a OffsetDateTime right? And all the calendar usage traits would be gone.

No, the UTC offset is used for disambiguation. The region is still needed. We need to handle these cases:

* A country decides to get rid of DST or introduce DST or change the DST offset.
* A country moves its whole time zone offset.
* You want to refer to a moment within a winter<->summer time change.

For calendar applications the region would always take precedence and the offset would be used to disambiguate only. Example: Your wedding event in Italy in 2024 at 14:00 must keep its local time at 14:00 even if Europe gets rid of DST handling in the meantime. So, the offset must be ignored unless there's an ambiguity.

See also how others are doing it:

* https://tc39.es/proposal-temporal/docs/ambiguity.html#ambiguity-due-to-dst-or-other-time-zone-offset-changes
* https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html

The only open question is whether we should support a DST flag in addition to UTC based offsets. If a country moves its time zone offset and this overlaps with a DST switch then a UTC offset alone can't correctly disambiguate the hour during a DST change. Only a DST flag could handle that edge case. Maybe that's too much of an edge case though.

core/common/src/ZonedDateTime.kt
```diff
@@ -0,0 +1,99 @@
...
+    localDateTime: LocalDateTime,
+    // TODO: this should be a RegionTimeZone
+    override val timeZone: TimeZone,
+    // TODO: Add optional DST offset or at least a UTC offset (should it be part of RegionTimeZone?)
+) : ZonedDateTime(localDateTime) {
+
+    public constructor(instant: Instant, timeZone: TimeZone) :
+            this(instant.toLocalDateTime(timeZone), timeZone)
+
+    // TODO: Should RegionTimeZone.toString() print with surrounding `[]`?
```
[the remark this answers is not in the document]
wkornewald: PostgreSQL, Elasticsearch, java.time, JS Temporal and others seem to have settled on this full format: `2021-03-28T03:16:20+02:00[Europe/Berlin]`

core/common/src/ZonedDateTime.kt
wkornewald: Since this is a duplicate of your comment above, let's continue the discussion in the comments.

core/common/src/ZonedDateTime.kt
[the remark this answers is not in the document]
cromefire: In that case I'd settle with the established format

cromefire:
In that case I'd stick to the TC39 proposal, that looks quite good (there's also usually with at lot of peoples checking it).

kkocel:
I port some JVM-related code and this exact class is needed. Hope that it gets merged...

dkhalanskyjb:
@kkocel most probably the `Instant` arithmetic, which already exists, will suit you: https://github.com/Kotlin/kotlinx-datetime#instant-arithmetic

kkocel:
@dkhalanskyjb actually I need to have a class that can be deserialized from the `java.time.ZonedDateTime` that comes from JVM backend

dkhalanskyjb:
For serialization and deserialization, `java.time.ZonedDateTime` and `java.time.OffsetDateTime` default formats can be implemented using `kotlinx.datetime.format.DateTimeComponents`: https://github.com/Kotlin/kotlinx-datetime?tab=readme-ov-file#parsing-and-formatting-partial-compound-or-out-of-bounds-data

PR #182, opened 2022-01-18
Fix Instant parsing on Native being too lenient

dkhalanskyjb:


ilya-g:
First we need to discuss what behavior we want from `Instant.parse`. In java.time, for example, Instant.parse does support times like 23:59:60.

dkhalanskyjb:
Closing in favor of #251

ilya-g:
I believe we still should discuss how to address an inconsistency between platforms if there is any.

ilya-g:
There's no inconsistency between platforms in kotlinx.datetime, but there's inconsistency between kotlinx.datetime and java.time in `Instant.parse`. Let's reopen.

dkhalanskyjb:
Ok: #332

PR #187, opened 2022-02-28
Add local time

### LINKED ISSUE #57: Provide time only representation, e.g. LocalTime
Hey,

Currently there is support for `LocalDateTime`, but I would like to use some time representation without a date. Ideall would be something like `LocalTime` for my use case.

my use case:
represent daily events

my current workaround:
use `DateTimePeriod` with a max of 24h in total

bishiboosh:
Fixes #57

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,9 @@ val thisTime: LocalTime = now.toLocalDateTime(TimeZone.currentSystemDefault()).t
 
 `LocalTime` can be constructed from four components, hour, minute, second and nanosecond:
 ```kotlin
-val knownTime = LocalTime(23, 59, 12, 999)
+val knownTime = LocalTime(23, 59, 12)
+val timeWithMillis = LocalTime(23, 59, 12, 999)
+val hourMinute = LocalTime(12, 13)
 ```
 
 ### Converting instant to and from unix time
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ The library provides the basic set of types for working with date and time:
 - `Clock` to obtain the current instant;
 - `LocalDateTime` to represent date and time components without a reference to the particular time zone; 
 - `LocalDate` to represent the components of date only;
+- `LocalTime` to represent the components of time only;
 - `TimeZone` and `FixedOffsetTimeZone` provide time zone information to convert between `Instant` and `LocalDateTime`;
 - `Month` and `DayOfWeek` enums;
 - `DateTimePeriod` to represent a difference between two instants decomposed into date and time units;
@@ -61,6 +62,8 @@ Here is some basic advice on how to choose which of the date-carrying types to u
   Also, use `LocalDateTime` to decode an `Instant` to its local date-time components for display and UIs.
   
 - Use `LocalDate` to represent a date of the event that does not have a specific time associated with it (like a birth date).
+
+- Use `LocalTime` to represent a time of the event that does not have a specific date associated with it.
  
 ## Operations
 
@@ -143,6 +146,21 @@ Note, that today's date really depends on the time zone in which you're observin
 val knownDate = LocalDate(2020, 2, 21)
 ```
 
+### Getting local time components
+
+`LocalTime` type represents local time without date. You can obtain it from `Instant`
+by converting it to `LocalDateTime` and taking its `time` property.
+
+```kotlin
+val now: Instant = Clock.System.now()
+val thisTime: LocalTime = now.toLocalDateTime(TimeZone.currentSystemDefault()).time
+```
+
+`LocalTime` can be constructed from four components, hour, minute, second and nanosecond:
+```kotlin
+val knownTime = LocalTime(23, 59, 12, 999)
+```
+
 ### Converting instant to and from unix time
 
 An `Instant` can be converted to a number of milliseconds since the Unix/POSIX epoch with the `toEpochMilliseconds()` function.
@@ -150,7 +168,7 @@ To convert back, use `Instant.fromEpochMilliseconds(Long)` companion object func
 
 ### Converting instant and local date/time to and from string
 
-Currently, `Instant`, `LocalDateTime`, and `LocalDate` only support ISO-8601 format.
+Currently, `Instant`, `LocalDateTime`, `LocalDate` and `LocalTime` only support ISO-8601 format.
 The `toString()` function is used to convert the value to a string in that format, and 
 the `parse` function in companion object is used to parse a string representation back. 
 
@@ -168,10 +186,13 @@ where it feels more convenient:
 
 `LocalDate` uses format with just year, month, and date components, e.g. `2010-06-01`.
 
+`LocalTime` uses format with just hour, minute, second and nanosecond components, e.g. `12:01:03`.
+
 ```kotlin
 "2010-06-01T22:19:44.475Z".toInstant()
 "2010-06-01T22:19:44".toLocalDateTime()
 "2010-06-01".toLocalDate()
+"12:01:03".toLocalTime()
 ```
 
 ### Instant arithmetic
--- a/README.md
+++ b/README.md
@@ -158,9 +158,9 @@ val thisTime: LocalTime = now.toLocalDateTime(TimeZone.currentSystemDefault()).t
 
 `LocalTime` can be constructed from four components, hour, minute, second and nanosecond:
 ```kotlin
-val knownTime = LocalTime(23, 59, 12)
-val timeWithMillis = LocalTime(23, 59, 12, 999)
-val hourMinute = LocalTime(12, 13)
+val knownTime = LocalTime(hour = 23, minute = 59, second = 12)
+val timeWithNanos = LocalTime(hour = 23, minute = 59, second = 12, nanosecond = 999)
+val hourMinute = LocalTime(hour = 12, minute = 13)
 ```
 
 ### Converting instant to and from unix time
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -143,14 +146,29 @@ Note, that today's date really depends on the time zone in which you're observin
...
+
+`LocalTime` type represents local time without date. You can obtain it from `Instant`
+by converting it to `LocalDateTime` and taking its `time` property.
+
+```kotlin
+val now: Instant = Clock.System.now()
+val thisTime: LocalTime = now.toLocalDateTime(TimeZone.currentSystemDefault()).time
+```
+
+`LocalTime` can be constructed from four components, hour, minute, second and nanosecond:
```
kluever, line 159: Might be worth showing an example with just H/M/S, and even another example with just H/M?

bishiboosh: sure, I'll try to see what I can add

vanniktech (approved):

core/common/src/LocalTime.kt
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public constructor(hour: Int, minute: Int, second: Int = 0, nanosecond: Int = 0)
+
+    /** Returns the hour-of-day time component of this date/time value. */
+    public val hour: Int
+    /** Returns the minute-of-hour time component of this date/time value. */
+    public val minute: Int
+    /** Returns the second-of-minute time component of this date/time value. */
+    public val second: Int
+    /** Returns the nanosecond-of-second time component of this date/time value. */
```
dkhalanskyjb, line 67:
```suggestion
    /** Returns the nanosecond-of-second time component of this time value. */
```
Ditto for everything else.

dkhalanskyjb (changes requested):
Good work! This is very similar to what we planned to implement.

core/darwin/src/Converters.kt
```diff
@@ -82,3 +82,17 @@ public fun LocalDateTime.toNSDateComponents(): NSDateComponents {
...
     return components
 }
+
+/**
+ * Converts the given [LocalTime] to [NSDateComponents].
+ *
+ * Of all the fields, only the bare minimum required for uniquely identifying the date and time are set.
+ */
+public fun LocalTime.toNSDateComponents(): NSDateComponents {
+    val components = date.toNSDateComponents()
```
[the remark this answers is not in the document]
bishiboosh: yup, this was a copy/paste done too hastily. I'll remove that and let someone with a MacOS add it if necessary

core/native/src/LocalTime.kt
```diff
@@ -36,11 +39,27 @@ internal val localTimeParser: Parser<LocalTime>
...
+    init {
+        fun check(value: Int, lower: Int, upper: Int, str: String) =
+            require(value >= lower && value <= upper) {
+                "Invalid time: $str must be a number between $lower and $upper, got $value"
+            }
+        check(hour, 0, 23, "hour")
+        check(minute, 0, 59, "minute")
+        check(second, 0, 59, "second")
+        check(nanosecond, 0, NANOS_PER_ONE - 1, "nanosecond")
+    }
```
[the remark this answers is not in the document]
bishiboosh: on the JVM and JS side, they are done on each object construction in https://github.com/openjdk/jdk/blob/d07f7c76c5df1473bffa41f10a89ca1e21e001ef/src/java.base/share/classes/java/time/LocalTime.java#L348 and https://github.com/js-joda/js-joda/blob/f55e329863e05a185c060b4b755dcb2ca58300a7/packages/core/src/LocalTime.js#L277, so that's why I thought it would be safe

core/js/src/LocalTime.kt
```diff
@@ -0,0 +1,50 @@
...
+            jtLocalTime.parse(isoString).let(::LocalTime)
+        } catch (e: Throwable) {
+            if (e.isJodaDateTimeParseException()) throw DateTimeFormatException(e)
+            throw e
+        }
+
+        internal actual val MIN: LocalTime = LocalTime(jtLocalTime.MIN)
+        internal actual val MAX: LocalTime = LocalTime(jtLocalTime.MAX)
+    }
+}
```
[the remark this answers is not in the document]
bishiboosh: I don't exactly see what I need to reformat here, I'm already using four spaces

bishiboosh:
@dkhalanskyjb I also added some converter methods between LocalDate/LocalTime

core/common/src/LocalTime.kt
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * Combines this time's components with the specified date components into a [LocalDateTime] value.
+ */
+public fun LocalTime.atDate(year: Int, monthNumber: Int, dayOfMonth: Int = 0): LocalDateTime =
+    LocalDateTime(year, monthNumber, dayOfMonth, hour, minute, second, nanosecond)
+
+/**
+ * Combines this time's components with the specified date components into a [LocalDateTime] value.
+ */
+public fun LocalTime.atDate(year: Int, month: Month, dayOfMonth: Int = 0): LocalDateTime =
```
dkhalanskyjb, line 105: I think we need a native speaker to determine whether `atDate` or `onDate` is the more idiomatic name in this occasion. https://forum.wordreference.com/threads/at-on-the-date.639568/ seems to imply that `atDate` is more formal and less common, but, frankly, I'm out of my depth here.

core/native/src/LocalTime.kt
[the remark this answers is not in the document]
dkhalanskyjb: Sure, it's safe, just a bit more extra work for the computer to do. Maybe we'll merge this as is and rework it later if benchmarks show that the impact is non-negligible.

core/jvm/src/LocalTime.kt
```diff
@@ -0,0 +1,48 @@
...
+                throw IllegalArgumentException(e)
+            })
+
+    public actual val hour: Int get() = value.hour
+    public actual val minute: Int get() = value.minute
+    public actual val second: Int get() = value.second
+    public actual val nanosecond: Int get() = value.nano
+
+    override fun equals(other: Any?): Boolean =
+            (this === other) || (other is LocalTime && this.value == other.value)
```
dkhalanskyjb, line 30: The indentation is incorrect here. If you're using IDEA to write this code, the whole file can be reformatted with the "reformat file" command.

core/common/test/LocalTimeTest.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    val fromComponents = LocalTime(hour, minute, second, nanosecond)
+    checkEquals(fromComponents, value)
+}
+
+fun checkEquals(expected: LocalTime, actual: LocalTime) {
+    assertEquals(expected, actual)
+    assertEquals(expected.hashCode(), actual.hashCode())
+    assertEquals(expected.toString(), actual.toString())
+}
```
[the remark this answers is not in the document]
dkhalanskyjb: Could you also remove the ported tests from `ThreeTenBpLocalTimeTest` so that they don't run twice?

core/common/src/LocalTime.kt
```diff
@@ -0,0 +1,111 @@
...
+import kotlinx.serialization.Serializable
+
+/**
+ * The time part of [LocalDateTime].
+ *
+ * This class does not describe specific *moments in time*, which are represented as [Instant] values.
+ * Instead, its instances can be thought of as clock readings, something that an observer in a particular time zone
+ * could witness.
+ * For example, `18:43` is not a *moment in time*, since someone in Berlin and someone in Tokyo would witness
+ * this on their clocks at different times.
```
dkhalanskyjb, line 18:
```suggestion
 * This class represents time-of-day without a referencing a specific date.
 * To reconstruct a full [LocalDateTime], representing civil date and time,
 * [LocalTime] needs to be combined with [LocalDate] via [LocalDate.atTime] or [LocalTime.atDate].
 *
 * Also, [LocalTime] does not reference a particular time zone.
 * Therefore, even on the same date, [LocalTime] denotes different moments of time.
 * For example, `18:43` happens at different moments in Berlin and in Tokyo.
```

core/common/src/LocalTime.kt
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * The arithmetic on [LocalTime] values is not provided, since without accounting for the time zone
+ * transitions or things like daylight saving or leap seconds it may give misleading results.
```
dkhalanskyjb, line 21: Daylight saving and (in some time scales) leap seconds are special cases of time zone transitions, so
```suggestion
 * transitions it may give misleading results.
```

dkhalanskyjb (changes requested):

core/common/src/LocalTime.kt
bishiboosh: I have to admit I'd be more partial to `atDate` because we already have `atTime`. Harmonizing those two names seems to me to be more important than being perfectly grammatically correct.

vanniktech: I'd also go for symmetry meaning `atDate`

vanniktech: I'm coming from Android/JVM and there's [`atDate`](https://docs.oracle.com/javase/8/docs/api/java/time/LocalTime.html#atDate-java.time.LocalDate-). Naturally, I'd always start by typing `.at` and see what's auto suggested.

Up until you raised the point, I never thought `onDate` does sound more natural.

dkhalanskyjb: Ok, after some searching, it does seem like `atDate` is fairly common as well.

dkhalanskyjb (approved):

ilya-g (approved):

wakaztahir:
When will this be available , I really need this...

dkhalanskyjb:
Thank you for the great work!

core/native/src/LocalTime.kt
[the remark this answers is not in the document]
PaulWoitaschek: For external creation I think it's very valuable to have these classes validate the correct instantiation. 
This will eventually prevent / surface real world bugs.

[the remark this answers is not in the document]
dkhalanskyjb: Certainly, I wasn't arguing for user-created instances not to be validated, the idea is that there are some *internal* places where we're absolutely sure that the provided data is correct. Doing the validation in the `init` block prevents us from bypassing the checks in these cases.

dkhalanskyjb:
We probably just forgot to remove them at some point. We didn't plan to make them public, at least with those names. Do you need them for something?

dkhalanskyjb:
Could you please show some code that uses these constants? I didn't quite grasp how they help with calculating the fasting streaks.

Also, if it's something non-trivial, please open a separate issue so that this rationale becomes easier to find.

Issue #34, opened 2020-08-14
Add Interval implementation


JWood48:
Add an Interval implementation like:
```
data class Interval(val from: Instant, val to: Instant)
```

imanushin:
I think think class can be extended:

```
data class Interval(override val start: Instant, override val endInclusive: Instant): ClosedRange<Instant> {
      init {
           require(start <= endInclusive) {
                 "Start should be before end, actual data: $this
           }
      }
}
```

Also I think we can also support sequences generation via implementing [feature like this](https://youtrack.jetbrains.com/issue/KT-40934).

ilya-g:
@JWood48, what are you going to represent with such a class? It's important to consider use cases to design the proposed type properly.

jasonCodesAway:
This is very useful! Modeling a single point is time great, modeling a time range/interval is _almost_ as common a use-case. Very handy for the [overlap](https://www.joda.org/joda-time/apidocs/org/joda/time/Interval.html#overlap-org.joda.time.ReadableInterval-)/[intersection](https://moment.github.io/luxon/docs/class/src/interval.js~Interval.html#instance-method-intersection) method in particular. Makes code readable/clean. Off the top of my head here are a couple popular libraries that do this:

1. [moment.js's luxon Interval](https://moment.github.io/luxon/docs/class/src/interval.js~Interval.html)
1. [joda time's Interval](https://www.joda.org/joda-time/key_interval.html)

Whathecode:
Intervals are not only useful for datetime, but _any_ interval [would make a great Kotlin language feature](https://github.com/Kotlin/KEEP/issues/138).

Therefore, I started a multiplatform generic interval project, [kotlinx.interval](https://github.com/Whathecode/kotlinx.interval). I use a similar approach as I did for C# intervals. The operations a generic interval implementation needs are decoupled from the implementation. As such, when you want to support new _types_ of intervals, all you need to provide is the set of operations.

Such virtual calls are probably not ideal when performance is concerned, but performance is not always an issue, and it is a good experiment to see what such a language-level API may look like.

Currently there is only the basis, and I haven't even introduced `Interval<Instant, Duration>` yet, but I'll start adding relevant operations ([similar to those I added for a C# interval library long ago](https://github.com/Whathecode/NET-Standard-Library-Extension/blob/master/src/Whathecode.System/AbstractInterval.cs)) over time.

JakeWharton:
[...]

PR #189, opened 2022-03-03
Local date range

PeterAttardo:
This pull request enables Kotlin range and progression semantics with the LocalDate and DatePeriod classes. Its implementation borrows from the patterns used for `IntRange`.

Example:

```
val start = LocalDate.parse("2020-01-01")
val end = LocalDate.parse("2020-12-31")

// calls function for each day of the year in 2020
(start..end).forEach {
    processDayOf2020(it)
}

// calls function for each day of the year in 2020, ordered in reverse
(end downTo start).forEach {
   processDayOf2020(it)
}

// calls function for every other day of the year in 2020
(start..end step DatePeriod(days=2)).forEach {
   processDayOf2020(it)
}
```


dkhalanskyjb:
We have a similar request, but for `Instant`: https://github.com/Kotlin/kotlinx-datetime/issues/34
Could you please describe why this is needed? Either in the issue I linked if your use cases are similar to the ones listed there, or in a new issue.

boswelja:
@dkhalanskyjb I've had a use case for this, though it's more of a convenience having a dedicated `LocalDateRange` compared to a `ClosedRange<LocalDate>`. I'm building a calendar library that allows consumers to update a particular range of dates on the screen, the intended use case for this is to allow changing the rendered content in chunks.

A few other use cases off the top of my head:

* Representing a selected date range
* Using plus/minus + forEach in conjunction to iterate over a large range in chunks (granted this could be achieved by having a list of dates and chunking that, but if you've only got a start and end date this could be more efficient)

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,118 @@
...
+}
+
+public open class LocalDateProgression
+internal constructor(internal val intProgression: IntProgression) : Iterable<LocalDate> {
+
+    internal constructor(
+        start: LocalDate,
+        endInclusive: LocalDate,
+        step: Int
+    ) : this(IntProgression.fromClosedRange(start.toEpochDays(), endInclusive.toEpochDays(), step))
```
dkhalanskyjb, line 23: This implementation is incorrect for dates whose `toEpochDays()` does not fit into an `Int`. See https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/-local-date/to-epoch-days.html

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,118 @@
...
+ * Copyright 2019-2022 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.datetime
+
+import kotlin.random.Random
+import kotlin.random.nextInt
+
+internal class LocalDateProgressionIterator(private val iterator: IntIterator) : Iterator<LocalDate> {
```
dkhalanskyjb, line 11: I think this can be a `private class`.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,118 @@
...
+        return intProgression == other.intProgression
+    }
+
+    override fun hashCode(): Int {
+        return intProgression.hashCode()
+    }
+
+
+    public companion object {
+        public fun fromClosedRange(
```
dkhalanskyjb, line 49: Until it's clear that this function is useful, it can be kept `internal`.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    override fun iterator(): Iterator<LocalDate> = LocalDateProgressionIterator(intProgression.iterator())
+
+    public open fun isEmpty(): Boolean = intProgression.isEmpty()
+
+    override fun toString(): String = if (intProgression.step > 0) "$first..$last step ${intProgression.step}" else "$first downTo $last step ${intProgression.step}"
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null || this::class != other::class) return false
```
dkhalanskyjb, line 36:
```suggestion
        if (other !is LocalDateProgression) return false
```
Then you won't need the following line, and in fact, this whole function can be a perfectly readable one-liner.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+
+    public companion object {
+        public fun fromClosedRange(
+            rangeStart: LocalDate,
+            rangeEnd: LocalDate,
+            stepValue: Int,
+            stepUnit: DateTimeUnit.DayBased
+        ): LocalDateProgression = LocalDateProgression(rangeStart, rangeEnd, stepValue * stepUnit.days)
```
dkhalanskyjb, line 54: Multiplication can overflow an `Int`. This must be accounted for.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+public class LocalDateRange(start: LocalDate, endInclusive: LocalDate) : LocalDateProgression(start, endInclusive, 1), ClosedRange<LocalDate>, OpenEndRange<LocalDate> {
+    override val start: LocalDate get() = first
+    override val endInclusive: LocalDate get() = last
+    override val endExclusive: LocalDate get() = endInclusive.plus(1, DateTimeUnit.DAY)
```
dkhalanskyjb, line 61: See the proposed API: this function should be deprecated, as it can throw an exception on overflow, which may be surprising. Also, this function should include the documentation and throw `IllegalStateException` if the overflow did happen. See https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-char-range/end-exclusive.html

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,118 @@
...
+    if (isEmpty())
+        throw NoSuchElementException("Progression $this is empty.")
+    return this.last
+}
+public fun LocalDateProgression.firstOrNull(): LocalDate? = if (isEmpty()) null else this.first
+public fun LocalDateProgression.lastOrNull(): LocalDate? = if (isEmpty()) null else this.last
+
+public fun LocalDateProgression.reversed(): LocalDateProgression = LocalDateProgression(intProgression.reversed())
+
+public fun LocalDateProgression.step(value: Int, unit: DateTimeUnit.DayBased) : LocalDateProgression = LocalDateProgression(intProgression.step(value * unit.days))
```
dkhalanskyjb, line 92: Arithmetic overflow.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,118 @@
...
+public fun LocalDateProgression.step(value: Long, unit: DateTimeUnit.DayBased) : LocalDateProgression = step(value.toInt(), unit)
+
+public infix fun LocalDate.downTo(that: LocalDate) : LocalDateProgression = LocalDateProgression.fromClosedRange(this, that, -1, DateTimeUnit.DAY)
+public infix fun LocalDate.downUntil(that: LocalDate) : LocalDateProgression = downTo(that.plus(1, DateTimeUnit.DAY))
+
+public operator fun LocalDate.rangeTo(that: LocalDate): LocalDateRange = LocalDateRange(this, that)
+public operator fun LocalDate.rangeUntil(that: LocalDate) : LocalDateRange = rangeTo(that.minus(1, DateTimeUnit.DAY))
+
+public fun LocalDateProgression.random(randomIntFunction: (IntRange) -> Int) : LocalDate = intProgression.random(randomIntFunction)
```
dkhalanskyjb, line 101: Are you certain we need these functions? I can't think of a use case for them.

core/common/test/DateTimePeriodTest.kt
```diff
@@ -149,9 +149,19 @@ class DateTimePeriodTest {
...
+        assertEquals(p2, DateTimePeriod(years = 1, months = 6, days = 3) - dp1)
+
         val dp2 = dp1 + p3 + p4
+        val dp3 = dp2 - p3 -p4
         assertEquals(dp1, dp2)
+        assertEquals(dp1, dp3)
         assertTrue(dp2 is DatePeriod)
+        assertTrue(dp3 is DatePeriod)
+        assertEquals(DateTimePeriod(years = -10, days=-3, hours = -2), -(p1 + p2 + p3))
+        assertEquals(DatePeriod(years = -11, months = -6), -(dp1 + p1))
```
dkhalanskyjb, line 164: It doesn't seem like these tests are still relevant to the PR.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,118 @@
...
+    ) : this(IntProgression.fromClosedRange(start.toEpochDays(), endInclusive.toEpochDays(), step))
+
+    public val first: LocalDate = LocalDate.fromEpochDays(intProgression.first)
+    public val last: LocalDate = LocalDate.fromEpochDays(intProgression.last)
+
+    override fun iterator(): Iterator<LocalDate> = LocalDateProgressionIterator(intProgression.iterator())
+
+    public open fun isEmpty(): Boolean = intProgression.isEmpty()
+
+    override fun toString(): String = if (intProgression.step > 0) "$first..$last step ${intProgression.step}" else "$first downTo $last step ${intProgression.step}"
```
dkhalanskyjb, line 32: If you call `toString` on other range implementations, the result can be copy-pasted into the code to obtain the same progression. This one can not, and I think it would be nice to preserve that property.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,118 @@
...
+public fun LocalDateProgression.reversed(): LocalDateProgression = LocalDateProgression(intProgression.reversed())
+
+public fun LocalDateProgression.step(value: Int, unit: DateTimeUnit.DayBased) : LocalDateProgression = LocalDateProgression(intProgression.step(value * unit.days))
+public fun LocalDateProgression.step(value: Long, unit: DateTimeUnit.DayBased) : LocalDateProgression = step(value.toInt(), unit)
+
+public infix fun LocalDate.downTo(that: LocalDate) : LocalDateProgression = LocalDateProgression.fromClosedRange(this, that, -1, DateTimeUnit.DAY)
+public infix fun LocalDate.downUntil(that: LocalDate) : LocalDateProgression = downTo(that.plus(1, DateTimeUnit.DAY))
+
+public operator fun LocalDate.rangeTo(that: LocalDate): LocalDateRange = LocalDateRange(this, that)
+public operator fun LocalDate.rangeUntil(that: LocalDate) : LocalDateRange = rangeTo(that.minus(1, DateTimeUnit.DAY))
```
dkhalanskyjb, line 99: This can overflow and throw, and this can be avoided.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,118 @@
...
+public fun LocalDateProgression.firstOrNull(): LocalDate? = if (isEmpty()) null else this.first
+public fun LocalDateProgression.lastOrNull(): LocalDate? = if (isEmpty()) null else this.last
+
+public fun LocalDateProgression.reversed(): LocalDateProgression = LocalDateProgression(intProgression.reversed())
+
+public fun LocalDateProgression.step(value: Int, unit: DateTimeUnit.DayBased) : LocalDateProgression = LocalDateProgression(intProgression.step(value * unit.days))
+public fun LocalDateProgression.step(value: Long, unit: DateTimeUnit.DayBased) : LocalDateProgression = step(value.toInt(), unit)
+
+public infix fun LocalDate.downTo(that: LocalDate) : LocalDateProgression = LocalDateProgression.fromClosedRange(this, that, -1, DateTimeUnit.DAY)
+public infix fun LocalDate.downUntil(that: LocalDate) : LocalDateProgression = downTo(that.plus(1, DateTimeUnit.DAY))
```
dkhalanskyjb, line 96: I don't think we have a `downUntil` function for any other ranges. Consistency of experience is extremely important to us: we want our users' knowledge to be completely transferable across the API entries we provide. "Aha, I need an `Int` range, and I already know how to do this for `LocalDate` ranges" should always work, unless there's a significant barrier to that.

Also, overflow.

dkhalanskyjb (changes requested):
* Could you also make the range an implementation of `Collection`? See https://youtrack.jetbrains.com/issue/KT-36932/Let-Progression-classes-implement-Collection-to-reduce-time-when-iterates-over-IntRange-when-cast-to-Iterable
* All API entries (classes, functions, properties) need separate tests, a KDoc, and a usage sample.

core/common/src/LocalDateRange.kt
PeterAttardo: This is the one place where `step` is semi-exposed publicly. Since step isn't really settled, I didn't want to commit too much in any direction.

Being able to copy-paste into code is a nice consequence of the code representation of primitives matching their string representation, but they diverge for more complex objects. Additionally, there is no infix step function for `LocalDateProgression` yet. I think we would lose a lot in readability if we were to make it copy-pasteable:

`LocalDate.parse("2000-01-01")..LocalDate.parse("2000-01-10").step(2, DateTimeUnit.DAY)` is a pretty ugly string representation. We could improve it slightly with:

`LocalDate(2000, 1, 1)..LocalDate(2000, 1, 10).step(2, DateTimeUnit.DAY)`, but it still has poor readability, and at that point we're abandoning the existing string representation of `LocalDate` entirely.

core/common/src/LocalDateRange.kt
PeterAttardo: Will add check and deprecation warning

core/common/src/LocalDateRange.kt
PeterAttardo: I added it for symmetry with `rangeUntil`, but I can remove it.

core/common/src/LocalDateRange.kt
PeterAttardo: Most use cases will be fine with selecting from a uniform distribution, but any time random selection is at play, I assume there are some use cases for selecting from a normal distribution, a Poisson distribution, etc. This function allows a user to implement whatever distribution logic they want.

core/common/test/DateTimePeriodTest.kt
PeterAttardo: They're from the changes to `DatePeriod` arithmetic, which I'll be reverting. I'll revert these as well.

core/common/src/LocalDateRange.kt
dkhalanskyjb: Why? Using `IntProgression` looks like just an implementation detail, not some fundamental issue. It would be better to just support all dates without relying on `toEpochDays()`.

core/common/src/LocalDateRange.kt
dkhalanskyjb: IIRC, we don't expect this function to be useful to the public (and it looks like we're right not to: https://grep.app/search?q=.fromClosedRange%28), and instead, it's there for the compiler to generate more optimized code when people write things using the functions we do expect them to use. So, `((1..11) step 2).reversed()` becomes something like `fromClosedRange(11, 1, -2)`. See https://github.com/JetBrains/kotlin/blob/658b9e9a4b11171bd09127113e152c12aaa5597d/libraries/tools/kotlin-stdlib-gen/src/templates/Ranges.kt `LocalDate`, being outside the standard library, can't benefit from this.

@ilya-g knows more about this and could maybe correct me.

core/common/src/LocalDateRange.kt
dkhalanskyjb: Ok, makes sense, thanks! Could you then include the information that the step is measured in days?

core/common/src/LocalDateRange.kt
PeterAttardo: It looks like the bounds of `LocalDate` vary by implementation (`native` goes from year `-999_999` to `999_999`, while `jvm` goes from year `-999_999_999` to `999_999_999`) but each internally convert to epoch days in order to do arithmetic. However those internal conversions to epoch days can be broader than the `common` conversion to epoch days. I do wonder if it would be better for the `common` `toEpochDays()` to return a `Long` instead of an `Int`, so that it is not more restricted than some implementations of `LocalDate` itself. 

An alternative implementation is to switch to a `LongProgression` between `0` and `size of progression` and use `fun next() = startDate.plus(iterator.next(), DateTimeUnit.DAY)`, which would bypass the `common` `toEpochDays()` and `fromEpochDays()`. However `daysUntil` _also_ returns an `Int`, and so the size of the range can't be reliably computed either. Without the ability to compute the size of the progression, we also lose the ability to generate random values from within the progression without doing something hideous like materializing the entire progression. 

I could attempt to reimplement it without leaning on any of the primitive progressions at all, but we would still lose the ability to calculate the last element (again excepting a hideous full iteration) for steps other than 1 day and -1 day, as we can't calculate the `differenceModulo` without an `Int`/`Long` representation of `LocalDate`. 

It seems like our options are:

1. Change the return type of several `common` methods from `Int` to `Long` to ensure that the full range of `LocalDate` can be expressed as days without overflowing.
2. Accept that the use of `Int` in `common` clamps the allowable ranges tighter than some implementations of `LocalDate` itself.
3. Push everything down and implement `LocalDateProgression` on a per-platform basis.

I'm inclined to leave it clamped for now, and then unclamp it in the future if/when the `common` day methods are upgraded from `Int` to `Long`.

core/common/src/LocalDateRange.kt
PeterAttardo: I'll switch it to internal

core/common/src/LocalDateRange.kt
PeterAttardo: I'll remove the function and let advanced users lean on a custom implementation of `Random`. 

Outside the scope of this library, but this does make me interested in formal implementations of `NormalRandom`, `PoissonRandom`, `BinomialRandom`, etc.

core/common/src/LocalDateRange.kt
dkhalanskyjb: > I do wonder if it would be better for the common toEpochDays() to return a Long instead of an Int, so that it is not more restricted than some implementations of LocalDate itself.

Maybe it would, but that's a separate problem. We're considering making the ranges the same, and it's possible that `toEpochDays()` will naturally have to return a `Long`: https://github.com/Kotlin/kotlinx-datetime/issues/432

> However `daysUntil` also returns an Int, and so the size of the range can't be reliably computed either.

Inside the library, we're not bound by the `public` API. If you need a `daysUntil` that returns a `Long`, you can implement it using something like

```
// common code
internal expect fun LocalDate.daysUntilLong(other: LocalDate): Long

// JVM
internal actual fun LocalDate.daysUntilLong(other: LocalDate): Long =
    this.value.until(other.value, ChronoUnit.DAYS)

// other platforms
internal actual fun LocalDate.daysUntilLong(other: LocalDate): Long =
    daysUntil(other).toLong()
```

PeterAttardo: I added platform-specific `toEpochDaysLong()` and `fromEpochDays(epochDays: Long)` functions, and switched from an `IntProgression` to a `LongProgression`. So far, so good. The one issue that remains is that, when implementing `Collection<LocalDate>`, it expects `size` to be an `Int`. Because we have expanded `LocalDateProgression` beyond the bounds of `Int`, it is now possible to create progressions of greater size than `Int.MAX_VALUE`.

In light of this, do we still want to implement `Collection`, and if so, do we want to throw in these cases, or clamp the size to `Int`?

dkhalanskyjb: I think we should clamp the size to `Int.MAX_VALUE` for consistency with Java (and, most likely, the other implementations of `Collection` we have with the same problem): https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Collection.html#size() @ilya-g, WDYT?

PeterAttardo:
Something I came across that's not exactly an issue in the library, but is a usability question when using the library in IntelliJ that I'm wondering if there's a good answer for.

Because Kotlin automatically imports `kotlin.ranges.*` into every file, that means that the extension function `fun <T : Comparable<T>> T.rangeTo(that: T): ClosedRange<T>` is by-default available for all code. `LocalDate` implements `Comparable<LocalDate>` and so if you type the following:

```Kotlin
(localDate1..localDate2).forEach {
    // some code in here
}
```

`localDate1..localDate2` will **not** display an `Unresolved reference` warning due to a failure to import `kotlinx.datetime.rangeTo`, but will happily use `kotlin.ranges.rangeTo` and return a `ClosedRange<LocalDate>` instead. Because `ClosedRange<LocalDate>` is not `LocalDateRange`, all of our nice progression methods are unavailable, and we get an `Unresolved reference` warning on `forEach` instead.

This is easily solved by importing `kotlinx.datetime.rangeTo`, but IntelliJ is unable to recognize this as the solution, because from its perspective there is no issue resolving the `rangeTo` operator, just an issue with calling non-existing methods on `ClosedRange<LocalDate>`. This means users will have to be aware of this pitfall, and manually type in the import in every file they want to use the range operator with `LocalDate`.

dkhalanskyjb:
@PeterAttardo, maybe this issue can be solved if `rangeTo` is a member method of `LocalDate`? Good job noticing this problem!

PeterAttardo:
@dkhalanskyjb I was hoping not to have to touch the `LocalDate` class, but that does seem like the cleanest solution. I have added the range operators to `LocalDate`, but because expect classes don't allow default implementations, I have moved the implementation logic into static methods in `LocalDateRange`. That can be cleaned up in the future if default implementations are ever allowed.

core/common/src/LocalDateRange.kt
PeterAttardo: Looking to understand how best to avoid the overflow. If I'm understanding correctly, this can only overflow if `that` is `LocalDate.MIN` (every other date has a valid `LocalDate` one day before it). By definition this would make the range empty, as every possible `this` that yields a non-empty range would itself be less than `LocalDate.MIN` and therefore not valid. This would seem to imply you can just return `LocalDateRange.EMPTY` any time `that` is `LocalDate.MIN`, however `EMPTY` would have incorrect values for `start` and `endInclusive`, which are part of the public API. Even if you preserved `start` by returning `this.rangeTo(LocalDate.MIN)`, your `endInclusive` would still be one day off, and if `this` happened to be `LocalDate.MIN` you'd accidentally create a non-empty range. I'm not sure how to avoid the throw while still maintaining correctness in the resulting range.

dkhalanskyjb: As a rule of thumb, when in doubt, just try to keep the behavior consistent with the stdlib:
```kotlin
fun main() {
    val range = 10 until (Int.MIN_VALUE)
    println(range) // prints "1..0"
}
```
So it's okay to return `EMPTY` if you notice you can't do anything meaningful and the range *is* empty. When it's possible to preserve the boundaries, then sure, it's worth doing:
```kotlin
fun main() {
    val range = 10 until (-1)
    println(range) // prints "10..-2"
}
```

core/common/src/LocalDateRange.kt
dkhalanskyjb: It turns out we (the Kotlin library team) have already decided to use the same contract as Java here, I just didn't know that.

dkhalanskyjb:
`@ilya-g` observed that the issue of forgetting to import methods actually affects the other extension functions in this PR as well: `last()` on a `LocalDateRange` shouldn't be the one used for generic collections. Still, making all these extension methods into members just for that wouldn't be correct: they *are* extensions semantically, and there are a lot of them. In the end, we decided this was a tooling problem: https://youtrack.jetbrains.com/issue/KTIJ-31281

Still, adding the range-creating methods directly to `LocalDate` makes sense semantically, so I think we should stick to that.

PeterAttardo:
@dkhalanskyjb Sounds like a good solution to address this with tooling. I focused on the range operators because, while you're right that imports can be overlooked for any more-specific extension method, at least with named methods you can type the first few letters, and then select the appropriate extension from the drop down, which will then auto-import the extension method. There's no such tooling for operators, so it's a manual type or copy-paste every time (or I suppose using the `rangeTo` and `rangeUntil` nomenclature to force the import, and then rewrite to the operator notation, which is equally arduous).

core/common/test/LocalDateRangeTest.kt
```diff
@@ -0,0 +1,274 @@
...
+                val range = (0L..23L step step)
+                assertEquals(
+                    expectedRand.nextLong(0..range.last / step).let { Jan_24_2000.minus(it * step, DateTimeUnit.DAY) },
+                    (Jan_24_2000 downTo Jan_01_2000).step(step, DateTimeUnit.DAY).random(actualRand)
+                )
+            }
+        }
+        repeat(20) {
+            assertTrue {
+                (Jan_01_2000..Jan_24_2000).step(5, DateTimeUnit.DAY).let { it.contains(it.random()) }
```
PeterAttardo, line 166: This test is not deterministic, which I'm not thrilled about. The other way to test it would be with a spy, but adding mockk just for this one test case seemed like overkill/overstepping.

dkhalanskyjb: In my opinion, this test is just fine the way it is, but if you prefer to test this deterministically, you can make a custom `Random` implementation and pass it to the function.

dkhalanskyjb: In my opinion, this test is fine the way it is (in fact, we have other randomly generated things in our tests), but if you prefer to test this deterministically, you can make a custom `Random` implementation and pass it to the function.

dkhalanskyjb:
@PeterAttardo, most discussions are marked as resolved; looks like most is this ready for a new round of reviews? If so, please click "re-request review" in the "Reviewers" panel.

core/commonJs/src/LocalDate.kt
```diff
@@ -42,6 +42,13 @@ public actual class LocalDate internal constructor(internal val value: jtLocalDa
             throw e
         }
 
+        internal actual fun fromEpochDays(epochDays: Long): LocalDate {
+            require(epochDays in Int.MIN_VALUE..Int.MAX_VALUE) {
+                "Invalid date: boundaries of LocalDate exceeded"
+            }
+            return fromEpochDays(epochDays.toInt())
+        }
```
dkhalanskyjb, line 50:
```suggestion
        internal actual fun fromEpochDays(epochDays: Long): LocalDate =
            fromEpochDays(epochDays.clampToInt())
```
Our internal `clampToInt` function clamps the `Long` to `Int.MAX_VALUE` or `Int.MIN_VALUE`, and both of these values don't fit into the range checked by `fromEpochDays(Int)`.

core/commonKotlin/src/LocalDate.kt
```diff
@@ -82,6 +82,13 @@ public actual class LocalDate actual constructor(public actual val year: Int, pu
             return LocalDate(yearEst, month, dom)
         }
 
+        public actual fun fromEpochDays(epochDays: Long): LocalDate {
+            require(epochDays in MIN_EPOCH_DAY..MAX_EPOCH_DAY) {
+                "Invalid date: boundaries of LocalDate exceeded"
+            }
+            return fromEpochDays(epochDays.toInt())
+        }
```
dkhalanskyjb, line 90:
```suggestion
        internal actual fun fromEpochDays(epochDays: Long): LocalDate =
            fromEpochDays(epochDays.clampToInt())
```
First, `clampToInt`, like above. Second: `public` seems incorrect.

core/commonKotlin/src/LocalDate.kt
```diff
@@ -121,6 +128,10 @@ public actual class LocalDate actual constructor(public actual val year: Int, pu
         return total - DAYS_0000_TO_1970
     }
 
+    internal actual fun toEpochDaysLong() : Long {
+        return toEpochDays().toLong()
+    }
```
dkhalanskyjb, line 133:
```suggestion
    internal actual fun toEpochDaysLong(): Long = toEpochDays().toLong()
```
Style: consistency with the other file, no space before `:`.

core/jvm/src/LocalDate.kt
```diff
@@ -39,6 +39,9 @@ public actual class LocalDate internal constructor(internal val value: jtLocalDa
         public actual fun fromEpochDays(epochDays: Int): LocalDate =
             LocalDate(jtLocalDate.ofEpochDay(epochDays.toLong()))
```
dkhalanskyjb, line 40: Minor: I'd write this as `fromEpochDays(epochDays.toLong())` to avoid the tiny bit of unnecessary code duplication. Maybe this goes for `toEpochDays()`, though there, the benefit is even more miniscule. Up to you.

core/common/test/samples/LocalDateRangeSamples.kt
```diff
@@ -0,0 +1,58 @@
...
+    val Jan_01_2000 = LocalDate(2000, 1, 1)
+    val Jan_05_2000 = LocalDate(2000, 1, 5)
+    val Jan_06_2000 = LocalDate(2000, 1, 6)
+
+    @Test
+    fun simpleRangeCreation() {
+        // Creating LocalDateRange from LocalDate
+        check(Jan_01_2000..Jan_05_2000 == LocalDateRange(Jan_01_2000, Jan_05_2000))
+        check(Jan_01_2000..<Jan_06_2000 == LocalDateRange(Jan_01_2000, Jan_05_2000))
+        check(Jan_05_2000 downTo Jan_01_2000 == LocalDateProgression(Jan_05_2000, Jan_01_2000, -1))
```
dkhalanskyjb, line 22: Samples shouldn't contain internal API (like `LocalDateProgression` constructor here). Suggestion: how about something like `check((LocalDate(1, 1, 2000)..LocalDate(1, 5, 2000)).toList() == listOf(...))` instead? It's verbose but explains in no uncertain terms what ranges contain, which is what we want from samples.

core/common/test/samples/LocalDateRangeSamples.kt
```diff
@@ -0,0 +1,58 @@
...
+package kotlinx.datetime.test.samples
+
+import kotlinx.datetime.*
+import kotlin.random.Random
+import kotlin.test.Test
+
+class LocalDateRangeSamples {
+    val Jan_01_2000 = LocalDate(2000, 1, 1)
+    val Jan_05_2000 = LocalDate(2000, 1, 5)
+    val Jan_06_2000 = LocalDate(2000, 1, 6)
```
dkhalanskyjb, line 15: The samples are broken, as they will not include these: 
image
How to check this: `./gradlew dokkaHtml`, then open `./core/build/dokka/html/kotlinx-datetime/kotlinx.datetime/-local-date/range-to.html`

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,263 @@
...
+
+/**
+ * Returns a random [LocalDate] within the bounds of the [LocalDateProgression] or null if the progression is empty.
+ *
+ * Takes the step into account; will not return any value within the range that would be skipped over by the progression.
+ *
+ * @sample kotlinx.datetime.test.samples.LocalDateRangeSamples.random
+ */
+public fun LocalDateProgression.randomOrNull(random: Random = Random) : LocalDate? = longProgression.randomOrNull(random)
+    ?.let(LocalDate.Companion::fromEpochDays)
```
dkhalanskyjb, line 250: These implementations can be made simpler (and more consistent with the other ranges) if they are limited to `LocalDateRange` instead of `LocalDateProgression`. Then, we know that `longProgression` is also a `LongRange`, so we can call `LongRange.random()`.

dkhalanskyjb (changes requested):
I like how this is coming together!

core/commonJs/src/LocalDate.kt
PeterAttardo: > both of these values don't fit into the range checked by fromEpochDays(Int)

We can assume this to be true because we know how `fromEpochDays(Int)` is implemented, but the signature itself does not require that to be the case. Do we mind that we're relying on that assumption?

core/common/src/LocalDateRange.kt
PeterAttardo: While `longProgression` for a `LocalDateRange` is functionally identical to a `LongRange` it is not, as implemented, technically an instance of that class.

`Range`s are essentially `Progression`s with a step size of `1`, but there is nothing in the constructors or companion creation methods that forces a `Progression` with a step size of `1` to actually _be_ a `Range`. `LongProgression.fromClosedRange(start, end, 1)` will return an instance of `LongProgression`, not a `LongRange` downcast to a `LongProgression`.

We can check if the `step` is `1` and create a `LongRange` if so, but that logic doesn't seem any simpler than what we have now, for reduced functionality.

core/commonJs/src/LocalDate.kt
dkhalanskyjb: `fromEpochDays(Long)` already has the knowledge that days outside `Int` won't work on this platform, and learning the same about `Int.MAX_VALUE` is just the next step.

core/common/src/LocalDateRange.kt
dkhalanskyjb: The idea was to cast to a `LongRange`, not to check the `step`; also, with inheritance, we could also specialize `longProgression` to have the `LongRange` type, so the cast wouldn't even be needed.

That said, now that I think about it, since `LocalDateProgression` is a `Collection` now, adding `random` to it seems like a good idea, because otherwise, `random` will still work, but only because `Collection` itself has such an extension function. So, I agree we should keep `LocalDateProgression.random`.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,263 @@
...
+
+internal fun LongProgression.random(random: Random = Random) : Long = random.nextLong(0L..(last - first) / step) * step + first
+
+internal fun LongProgression.randomOrNull(random: Random = Random) : Long? = if (isEmpty()) null else random(random)
```
dkhalanskyjb, line 254:
```suggestion
// incorrect in general; see `random` just above
private fun LongProgression.randomOrNull(random: Random = Random) : Long? = if (isEmpty()) null else random(random)
```

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,263 @@
...
+
+internal fun LongProgression.contains(element: Long) : Boolean = element in (if(step > 0) first..last else last..first) && (element - first) % step == 0L
```
dkhalanskyjb, line 256:
```suggestion
// this implementation is incorrect in general (for example, `(Long.MIN_VALUE..Long.MAX_VALUE).step(5).contains(2)` returns `false`
// incorrectly https://www.wolframalpha.com/input?i=-2%5E63+%2B+1844674407370955162+*+5),
// but for the range of epoch days in LocalDate it's good enough
private fun LongProgression.contains(element: Long) : Boolean = element in (if(step > 0) first..last else last..first) && (element - first) % step == 0L
```

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,263 @@
...
+ *
+ * @sample kotlinx.datetime.test.samples.LocalDateRangeSamples.progressionWithStep
+ */
+public fun LocalDateProgression.step(value: Int, unit: DateTimeUnit.DayBased) : LocalDateProgression = step(value.toLong(), unit)
+
+/**
+ * Returns a [LocalDateProgression] with the same start and end, but a changed step value.
+ *
+ * **Pitfall**: the value parameter represents the magnitude of the step, not the direction, and therefore must be positive.
+ * Its sign will be matched to the sign of the existing step, in order to maintain the directionality of the progression.
```
dkhalanskyjb, line 217: > directionality

I don't see a suitable meaning in https://www.merriam-webster.com/dictionary/directionality
Maybe just "direction"?

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,263 @@
...
+ *
+ * @sample kotlinx.datetime.test.samples.LocalDateRangeSamples.reversedProgression
+ */
+public fun LocalDateProgression.reversed(): LocalDateProgression = LocalDateProgression(longProgression.reversed())
+
+/**
+ * Returns a [LocalDateProgression] with the same start and end, but a changed step value.
+ *
+ * **Pitfall**: the value parameter represents the magnitude of the step, not the direction, and therefore must be positive.
+ * Its sign will be matched to the sign of the existing step, in order to maintain the directionality of the progression.
```
dkhalanskyjb, line 206: Ditto.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,263 @@
...
+/**
+ * Returns the last [LocalDate] of the [LocalDateProgression], or null if the progression is empty.
+ *
+ * @sample kotlinx.datetime.test.samples.LocalDateRangeSamples.firstAndLast
+ */
+public fun LocalDateProgression.lastOrNull(): LocalDate? = if (isEmpty()) null else this.last
+
+/**
+ * Returns a reversed [LocalDateProgression], i.e. one that goes from [last] to [first].
+ * The sign of the step is switched, in order to reverse the directionality of the progression.
```
dkhalanskyjb, line 196: Ditto.

core/common/test/LocalDateRangeTest.kt
```diff
@@ -0,0 +1,274 @@
...
+        assertFails { (Jan_02_2000..Jan_01_2000).first() }
+        assertFails { (Jan_01_2000 downTo Jan_02_2000).first() }
+    }
+
+    @Test
+    fun last() {
+        assertEquals((Jan_01_2000..Jan_01_2000).last(), Jan_01_2000)
+        assertEquals((Jan_01_2000 downTo Jan_01_2000).last(), Jan_01_2000)
+        assertEquals((Jan_01_2000..Jan_05_2000).last(), Jan_05_2000)
+        assertEquals((Jan_05_2000 downTo Jan_01_2000).last(), Jan_01_2000)
```
dkhalanskyjb, line 186: Also worth testing that `last()` returns not the specified right boundary but the actual last emitted element, like `(1..10).step(5).last() == 6`.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,263 @@
...
+/**
+ * Returns a [LocalDateProgression] with the same start and end, but a changed step value.
+ *
+ * **Pitfall**: the value parameter represents the magnitude of the step, not the direction, and therefore must be positive.
+ * Its sign will be matched to the sign of the existing step, in order to maintain the directionality of the progression.
+ * If you wish to switch the direction of the progression, use [LocalDateProgression.reversed]
+ *
+ * @sample kotlinx.datetime.test.samples.LocalDateRangeSamples.progressionWithStep
+ */
+public fun LocalDateProgression.step(value: Long, unit: DateTimeUnit.DayBased) : LocalDateProgression = LocalDateProgression(longProgression.step(safeMultiply(value, unit.days.toLong())))
```
dkhalanskyjb, line 222: I don't think `step` should throw an exception if `value * unit.days` overflows a `Long`. We can just clamp the result to `Long.(MAX|MIN)_VALUE` depending on the signs of the inputs, and `step` will behave correctly.

Or even if we decide to throw a `DateTimeArithmeticException`, we must at least declare it.

In any case, it's also a thing worth an additional test.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,263 @@
...
+     */
+    override fun isEmpty(): Boolean = first > last
+
+    /**
+     * Returns a string representation of the range using the range operator notation.
+     */
+    override fun toString(): String = "$first..$last"
+
+    public companion object {
+        public val EMPTY: LocalDateRange = LocalDateRange(LocalDate(1970, 1, 2), LocalDate(1970, 1, 1))
```
dkhalanskyjb, line 142: The documentation is missing.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,263 @@
...
+     */
+    override val start: LocalDate get() = first
+
+    /**
+     * Returns the upper bound of the range, inclusive.
+     */
+    override val endInclusive: LocalDate get() = last
+
+    /**
+     * returns the upper bound of the range, exclusive.
```
dkhalanskyjb, line 111:
```suggestion
     * Returns the upper bound of the range, exclusive.
```

dkhalanskyjb (changes requested):
I noticed one important thing (`step()` throwing) and several small things. Once all of this is cleaned up, from my point of view, it's good to go! I added `@ilya-g`, as he may have useful insights, especially regarding the implementation of `Collection`.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,263 @@
...
+public fun LocalDateProgression.randomOrNull(random: Random = Random) : LocalDate? = longProgression.randomOrNull(random)
+    ?.let(LocalDate.Companion::fromEpochDays)
+
+internal fun LongProgression.random(random: Random = Random) : Long = random.nextLong(0L..(last - first) / step) * step + first
+
+internal fun LongProgression.randomOrNull(random: Random = Random) : Long? = if (isEmpty()) null else random(random)
+
+internal fun LongProgression.contains(element: Long) : Boolean = element in (if(step > 0) first..last else last..first) && (element - first) % step == 0L
+
+internal val LongProgression.size: Int
```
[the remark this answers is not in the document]
PeterAttardo: This looks to be an issue because the negative of `Long.MIN_VALUE` wraps back around to itself due to an overflow error. Do we want to solve this by implementing a `safeSubtract` method instead of using `safeAdd` on the negative of the subtrahend, or is it not worth the trouble, and we can get by with this explanatory comment?

core/common/src/LocalDateRange.kt
PeterAttardo: This feels like one of those cases where I have an intuitive feel of the nuances of English, but can't point to a hard and fast rule. A progression has directionality, as well as having _a_ direction. I think in this case either `directionality` or `direction` is valid, so I'll defer to whichever you prefer.

dkhalanskyjb: My (non-native) intuitive feel of English tells me that ranges as a concept have directionality, but a specific range has a direction. Also, if both words can be used here, I'd stick to the simpler one so as not to distract the reader. API references should strive to be boring to read, I think.

dkhalanskyjb (changes requested):
The compilation fails for me, and when I fix it, the `getSize` test fails.

core/common/test/LocalDateRangeTest.kt
```diff
@@ -0,0 +1,283 @@
...
+            (Jan_01_2000..Jan_05_2000).reversed()
+        )
+        assertContentEquals(
+            Jan_01_2000..Jan_05_2000,
+            (Jan_05_2000 downTo Jan_01_2000).reversed()
+        )
+        assertContentEquals(
+            Jan_01_2000..Jan_01_2000,
+            (Jan_01_2000..Jan_01_2000).reversed()
+        )
```
ilya-g, line 230: I believe it should check more stricter - for equality, not for content equality

core/common/test/LocalDateRangeTest.kt
```diff
@@ -0,0 +1,283 @@
...
+
+package kotlinx.datetime.test
+
+import kotlinx.datetime.*
+import kotlinx.datetime.internal.clampToInt
+import kotlin.random.Random
+import kotlin.random.nextLong
+import kotlin.test.*
+
+class LocalDateRangeTest {
```
ilya-g, line 14: Could you also add tests for equals/hashCode contract of ranges/progressions?
E.g. that a progression with the step of 1 day is equal to a range. Or that a progression with the step of 7 days is equal to a progression with the step of a week.

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,272 @@
...
+        if (last == LocalDate.MAX) error("Cannot return the exclusive upper bound of a range that includes LocalDate.MAX.")
+        return endInclusive.plus(1, DateTimeUnit.DAY)
+    }
+
+    /**
+     * Returns true iff [value] is contained within the range.
+     * i.e. [value] is between [start] and [endInclusive].
+     */
+    @Suppress("ConvertTwoComparisonsToRangeCheck")
+    override fun contains(value: LocalDate): Boolean = first <= value && value <= last
```
dkhalanskyjb, line 129: There's a compiler warning that's actually helpful: `The corresponding parameter in the supertype 'LocalDateProgression' is named 'element'. This may cause problems when calling this function with named arguments.` I agree, let's rename this to `contains(element)`.

dkhalanskyjb (approved):
Other than the small naming issue, I think it's good to go!

core/common/src/LocalDateRange.kt
PeterAttardo: I saw that earlier. Unfortunately `ClosedRange<T>` and `OpenEndRange<T>` both define the method `public operator fun contains(value: T): Boolean` while `Collection<E>` defines `public operator fun contains(element: E)`. `LocalDateRange` implements all three interfaces. They are all implemented with a single method, so whatever name one chooses for the parameter in that method is going to generate that warning for at least one of the parent interfaces.

core/common/test/LocalDateRangeTest.kt
```diff
@@ -0,0 +1,312 @@
...
+            Jan_01_2000,
+            (Jan_01_2000..Jan_01_2000).random()
+        )
+
+        assertEquals(
+            Jan_01_2000,
+            (Jan_01_2000 downTo Jan_01_2000).random()
+        )
+
+        assertFailsWith<IllegalArgumentException> {
```
ilya-g, line 129: Should be `NoSuchElementException`

core/common/src/LocalDateRange.kt
ilya-g: In the standard library, we chose to have the parameter named `value` both in range's and progression's `contains`

core/common/test/LocalDateRangeTest.kt
PeterAttardo: The underlying `Random.nextLong(range: LongRange)` call throws an `IllegalArgumentException` for an empty range. Do we want to wrap that call in a try-catch and rethrow as a `NoSuchElementException`?

core/common/src/LocalDateRange.kt
```diff
@@ -0,0 +1,272 @@
...
+     * Returns the number of dates in the progression.
+     * Returns [Int.MAX_VALUE] if the number of dates overflows [Int]
+     */
+    override val size: Int
+        get() = longProgression.size
+
+    /**
+     * Returns true iff every element in [elements] is a member of the progression.
+     */
+    override fun containsAll(elements: Collection<LocalDate>): Boolean = elements.all(::contains)
```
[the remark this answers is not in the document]
PeterAttardo: Is this a deeper issue than at first appears? The signature of the methods implies no such unsafety, however after digging into it I can see that the methods in `AbstractCollection` are annotated with `@UnsafeVariance`, while the methods in `UByteArray` contain explicit type checks. I suppose it's a concession to Java compatibility where the `contains` methods take instances of `Object` instead of a typed parameter?

In any event, I'll copy the type check logic from `UByteArray` and add the test.

core/common/test/LocalDateRangeTest.kt
ilya-g: Yes, that's what `random()` function throws on empty collections and lists in stdlib.

ilya-g: But it would be better to check for an empty range instead of wrapping a call to `nextLong()`

ilya-g (approved):

dkhalanskyjb:
Thank you a lot for the hard work! Merged manually in https://github.com/Kotlin/kotlinx-datetime/commit/d15ec21e1d53944168357bf70934fec46107730f to squash this into one commit.

lukellmann:
I noticed that `LocalDateProgression.random` [doesn't throw the documented exception](https://github.com/Kotlin/kotlinx-datetime/blob/06b389bb3a03ea97a41cf1662580bc2f00419c90/core/common/src/LocalDateRange.kt#L264-L269).

dkhalanskyjb:
@lukellmann, thanks, added the fix to the `year-month` branch (which, too, adds a `random` function): https://github.com/Kotlin/kotlinx-datetime/pull/457/commits/8c04d699e4fea5213b4fc4410f26aa8eb5c2c74a

PR #204, opened 2022-05-17
LocalTime#toSecondOfDay & LocalTime#ofSecondOfDay.

vanniktech:
Motivation: I'm storing my LocalTime using second of day to leverage sql sorting

### DIFF
--- a/core/common/src/LocalTime.kt
+++ b/core/common/src/LocalTime.kt
@@ -40,6 +40,20 @@ public expect class LocalTime : Comparable<LocalTime> {
          */
         public fun parse(isoString: String): LocalTime
 
+        /**
+         * Returns a LocalTime with the specified [secondOfDay]. The nanosecond field will be set to zero.
+         *
+         * @throws IllegalArgumentException if the boundaries of [secondOfDay] are exceeded.
+         */
+        public fun fromSecondOfDay(secondOfDay: Int): LocalTime
+
+        /**
+         * Returns a LocalTime with the specified [nanosecondOfDay].
+         *
+         * @throws IllegalArgumentException if the boundaries of [nanosecondOfDay] are exceeded.
+         */
+        public fun fromNanosecondOfDay(nanosecondOfDay: Long): LocalTime
+
         internal val MIN: LocalTime
         internal val MAX: LocalTime
     }
@@ -65,6 +79,12 @@ public expect class LocalTime : Comparable<LocalTime> {
     public val second: Int
     /** Returns the nanosecond-of-second time component of this time value. */
     public val nanosecond: Int
+
+    /** Returns the time as a second of a day, from 0 to 24 * 60 * 60 - 1. */
+    public val secondOfDay: Int
+
+    /** Returns the time as a nanosecond of a day, from 0 to 24 * 60 * 60 * 1000 - 1. */
+    public val nanosecondOfDay: Long
 
     /**
      * Compares `this` time value with the [other] time value.
--- a/core/common/src/LocalTime.kt
+++ b/core/common/src/LocalTime.kt
@@ -81,10 +81,10 @@ public expect class LocalTime : Comparable<LocalTime> {
     public val nanosecond: Int
 
     /** Returns the time as a second of a day, from 0 to 24 * 60 * 60 - 1. */
-    public val secondOfDay: Int
+    public fun toSecondOfDay(): Int
 
-    /** Returns the time as a nanosecond of a day, from 0 to 24 * 60 * 60 * 1000 - 1. */
-    public val nanosecondOfDay: Long
+    /** Returns the time as a nanosecond of a day, from 0 to 24 * 60 * 60 * 1_000_000_000 - 1. */
+    public fun toNanosecondOfDay(): Long
 
     /**
      * Compares `this` time value with the [other] time value.
--- a/core/common/test/LocalTimeTest.kt
+++ b/core/common/test/LocalTimeTest.kt
@@ -88,20 +88,20 @@ class LocalTimeTest {
         val data = mapOf(
             0L to LocalTime(0, 0),
             5000000001L to LocalTime(0, 0, 5, 1),
-            44105000000100L to LocalTime(12, 15, 5, 100),
-            86399000000999L to LocalTime(23, 59, 59, 999),
+            44105123456789L to LocalTime(12, 15, 5, 123456789),
+            86399999999999L to LocalTime(23, 59, 59, 999999999),
         )
 
         data.forEach { (nanosecondOfDay, localTime) ->
-            assertEquals(nanosecondOfDay, localTime.nanosecondOfDay)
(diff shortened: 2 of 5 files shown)


core/common/src/LocalTime.kt
```diff
@@ -40,6 +40,11 @@ public expect class LocalTime : Comparable<LocalTime> {
          */
         public fun parse(isoString: String): LocalTime
 
+        /**
+         * Returns a LocalTime with the specified [secondOfDay]. The nanosecond field will be set to zero.
+         */
+        public fun ofSecondOfDay(secondOfDay: Int): LocalTime
```
ilya-g, line 46: It should be documented when and what this function throws with the `@throws` documentation block.
I presume it should be an `IllegalArgumentException`.
There should be a test checking that such exception or its inheritor is indeed thrown in case of invalid input.

ilya-g (changes requested):
Indeed, it seems convenient to have such conversion functions.
Two points though:
- the underlying value is a number of nanoseconds since the beginning of day, so it would be consistent to be able to get both nanoseconds of day and seconds of day;
- naming should be discussed further, so it is consistent with other functions/properties. 
  - for example, additional factory methods in `Instant` are called `fromSomething(...)`. 
  - the opposite conversion methods can be either functions, like `toNanosecondsOfDay()`, or properties, like `nanosecondsOfDay`.
  - singular or plural form of seconds/nanoseconds

core/common/src/LocalTime.kt
```diff
@@ -66,6 +80,12 @@ public expect class LocalTime : Comparable<LocalTime> {
     /** Returns the nanosecond-of-second time component of this time value. */
     public val nanosecond: Int
 
+    /** Returns the time as a second of a day, from 0 to 24 * 60 * 60 - 1. */
+    public val secondOfDay: Int
+
+    /** Returns the time as a nanosecond of a day, from 0 to 24 * 60 * 60 * 1000 - 1. */
```
dkhalanskyjb, line 86:
```suggestion
    /** Returns the time as a nanosecond of a day, from 0 to 24 * 60 * 60 * 1_000_000_000 - 1. */
```

core/common/test/LocalTimeTest.kt
```diff
@@ -85,6 +83,58 @@ class LocalTimeTest {
         assertFailsWith<IllegalArgumentException> { LocalTime(0, 0, 0, 1_000_000_000) }
     }
 
+    @Test
+    fun fromNanosecondOfDay() {
+        val data = mapOf(
+            0L to LocalTime(0, 0),
+            5000000001L to LocalTime(0, 0, 5, 1),
+            44105000000100L to LocalTime(12, 15, 5, 100),
+            86399000000999L to LocalTime(23, 59, 59, 999),
```
dkhalanskyjb, line 92: All of these have the millisecond and microsecond triads zeroed. I'd add a test with `123456789` nanoseconds or something like that.

core/common/test/LocalTimeTest.kt
```diff
@@ -85,6 +83,58 @@ class LocalTimeTest {
...
+
+        data.forEach { (nanosecondOfDay, localTime) ->
+            assertEquals(nanosecondOfDay, localTime.nanosecondOfDay)
+            assertEquals(localTime, LocalTime.fromNanosecondOfDay(nanosecondOfDay))
+        }
+    }
+
+    @Test
+    fun fromNanosecondOfDayInvalid() {
+        LocalTime.fromNanosecondOfDay(0)
```
dkhalanskyjb, line 103: Checked by another test.

core/common/test/LocalTimeTest.kt
```diff
@@ -85,6 +83,58 @@ class LocalTimeTest {
...
+        assertFailsWith<IllegalArgumentException> { LocalTime.fromNanosecondOfDay(-1) }
+        assertFailsWith<IllegalArgumentException> { LocalTime.fromNanosecondOfDay(Long.MAX_VALUE) }
```
dkhalanskyjb, line 105: A good idea is to check the boundaries. For example, in the test above, in addition to (or instead of) testing `86399000000999L`, `86399999999999L`, the maximum admissible value, could be tested, and here we could test `86400000000000L`, the first non-admissible value.

core/common/test/LocalTimeTest.kt
```diff
@@ -85,6 +83,58 @@ class LocalTimeTest {
...
+            assertEquals(secondOfDay, localTime.secondOfDay)
+            assertEquals(localTime, LocalTime.fromSecondOfDay(secondOfDay))
+        }
+    }
+
+    @Test
+    fun fromSecondOfDayInvalid() {
+        LocalTime.fromSecondOfDay(0)
+        assertFailsWith<IllegalArgumentException> { LocalTime.fromSecondOfDay(-1) }
+        assertFailsWith<IllegalArgumentException> { LocalTime.fromSecondOfDay(Int.MAX_VALUE) }
```
dkhalanskyjb, line 127: Ditto: 86400 would be a good thing to check.

dkhalanskyjb (changes requested):
After an internal discussion, we decided to have `fromSecondOfDay`/`toSecondOfDay` and `fromNanosecondOfDay`/`toNanosecondOfDay` pairs (that is, the accessors shouldn't be properties).

We also decided that we want `[to/from]MillisecondOfDay` as well, but if you don't feel like adding it, we'll do it on our own. Also, there's native-only `ThreeTenBpLocalTimeTest` that has tests for the new functions, and some of them could be ported to common code, but this is also something that could be easier for us to do.

core/common/src/LocalTime.kt
```diff
@@ -40,6 +40,20 @@ public expect class LocalTime : Comparable<LocalTime> {
          */
         public fun parse(isoString: String): LocalTime
 
+        /**
+         * Returns a LocalTime with the specified [secondOfDay]. The nanosecond field will be set to zero.
+         *
+         * @throws IllegalArgumentException if the boundaries of [secondOfDay] are exceeded.
```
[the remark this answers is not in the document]
vanniktech: It's referencing the parameter. Also works for me.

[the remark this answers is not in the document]
dkhalanskyjb: Ah, ok, my bad.

core/js/src/LocalTime.kt
```diff
@@ -44,7 +46,21 @@ public actual class LocalTime internal constructor(internal val value: jtLocalTi
             throw e
         }
 
+        public actual fun fromSecondOfDay(secondOfDay: Int): LocalTime = try {
+            jtLocalTime.ofSecondOfDay(secondOfDay, 0).let(::LocalTime)
+        } catch (e: Throwable) {
+            if (e.isJodaDateTimeParseException()) throw DateTimeFormatException(e)
```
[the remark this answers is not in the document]
vanniktech: Removed it and also fixed the throwing part.

dkhalanskyjb (approved):
Good work, thanks!

PR #205, opened 2022-05-21
Add isToday() function to LocalDate

serbelga:


JakeWharton:
I am opposed to this function as written as it has a global system clock dependency which makes its usage impossible to test code that uses it.

serbelga:
I'm closing this pr according to what has been commented

PR #212, opened 2022-05-27
Remove the floorMod and floorDiv functions

dkhalanskyjb:
We have those in the standard library now.


ilya-g (approved):
`floorDiv` version here was slightly more optimized to the fact that the divisor is supposed to be non-negative, but I don't think the difference would be so critical to avoid using more universal stdlib version.

dkhalanskyjb:
Looking at the implementation of `floorDiv`, I think it's pretty much even. Here, on the happy path, we had one comparison and one division, and in the stdlib, we have one division, one comparison, and a `xor`. Also, the `mod` in the stdlib may be more performant than `floorMod` here.

Issue #210, opened 2022-05-25
Plurals vs singulars in ordinals


dkhalanskyjb:
This issue is for gathering inputs from people with an intuitive understanding of English to help us name things.

Preliminaries: we have `Instant.epochSeconds`, which returns *the number of seconds* that passed since the Unix epoch. On the epoch itself and after some number of nanoseconds that don't amount to a whole second, `Instant.epochSeconds` returns 0.

Other libraries, like `java.time`, provide the same functionality, but with a slightly different name: `Instant.epochSecond` (note it being singular). We assume that the idea behind this is "the nth second since the epoch".

However, colloquially, the nth second/etc. is the second/etc. that goes on before `n` whole seconds/etc. have passed. For example, January is the *first month of the year*, that is, the month that goes on before a whole month has passed.

Questions for the readers:
* If you see `Instant.epochSecond == 1`, what do you think it means—does it mean that it is *the first* second since the start of the epoch, or that 1 full second has passed since then?
* The same, but about `Instant.epochSeconds`. Does this naming seem natural, or slightly odd?
* The same about `LocalTime.secondOfDay` and `LocalTime.secondsOfDay`: do they mean the same thing to you, or do you feel the first one is an ordinal, but the second one is the count?

PeterAttardo:
I think the following generally holds:

* If it is a singular, there is an implied `nth {unit} of {time period}`. So `instant.epochSecond = 5` would naturally translate to `5th second of epoch` where `epoch` is the time period between the start of the Unix epoch and infinity.
* If it is a plural, there is an implied `n {units} since {point in time}` or `n {units} since start of {time period}`. So `instant.epochSeconds = 5` would naturally translate to `5 seconds since epoch` where `epoch` is the exact point in time that the Unix epoch started. 
 
(Interesting sidenote is that this illustrates `the Unix Epoch`'s ability to refer to either a period or a discrete moment in time, depending on context.)

For this reason `LocalTime.secondOfDay` works, because it is harmonious with these rules, but `LocalTime.secondsOfDay` feels off, because the implication of the plural clashes with the explicit use of `of` in the name.

[...]

PR #214, opened 2022-06-01
Implement LocalDate.fromEpochDays

dkhalanskyjb:
Naming is up to discussion. See https://github.com/Kotlin/kotlinx-datetime/issues/210. For now, I'm for the name `fromEpochDays` because days-of-anything are most commonly numbered from 1. `*EpochDays`, as opposed to `*EpochDay`, helps highlight us using the *number* of days and not their 1-based ordinal.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -68,7 +68,7 @@ kotlin {
         nodejs {
             testTask {
                 useMocha {
-                    timeout = "5s"
+                    timeout = "30s"
                 }
             }
         }
--- a/core/common/src/math.kt
+++ b/core/common/src/math.kt
@@ -201,3 +201,16 @@ internal fun multiplyAndAdd(d: Long, n: Long, r: Long): Long {
     }
     return safeAdd(safeMultiply(md, n), mr)
 }
+
+// org.threeten.bp.chrono.IsoChronology#isLeapYear
+internal fun isLeapYear(year: Int): Boolean {
+    val prolepticYear: Long = year.toLong()
+    return prolepticYear and 3 == 0L && (prolepticYear % 100 != 0L || prolepticYear % 400 == 0L)
+}
+
+internal fun Int.monthLength(isLeapYear: Boolean): Int =
+    when (this) {
+        2 -> if (isLeapYear) 29 else 28
+        4, 6, 9, 11 -> 30
+        else -> 31
+    }
--- a/core/common/src/DateTimePeriod.kt
+++ b/core/common/src/DateTimePeriod.kt
@@ -5,6 +5,7 @@
 
 package kotlinx.datetime
 
+import kotlinx.datetime.internal.*
 import kotlinx.datetime.serializers.DatePeriodIso8601Serializer
 import kotlinx.datetime.serializers.DateTimePeriodIso8601Serializer
 import kotlin.math.*
--- a/core/common/src/DateTimeUnit.kt
+++ b/core/common/src/DateTimeUnit.kt
@@ -5,6 +5,7 @@
 
 package kotlinx.datetime
 
+import kotlinx.datetime.internal.safeMultiply
 import kotlinx.datetime.serializers.*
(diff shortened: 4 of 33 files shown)


ilya-g:
That can happen on TC, because agents there are usually computationally weaker than dev laptops or workstations.
In this case I suggest to reduce the test complexity, since computing all these ~2mln datapoints, especially incurring allocations, can be indeed a tough call for JS.

dkhalanskyjb:
I think we do want to test this thoroughly on a large range of values, because of the complexity of the code. A lot of things can go wrong there. In particular, it uses the notion of 400-year cycles in its implementation, and march-based years. I wouldn't dare to change anything there without being sure that every math operation there was exercised via brute force.

dkhalanskyjb:
I fixed the test—despite the comment in its beginning, it was invoking `plus` instead of `next` and `previous`—and it seems for now that the problem is mitigated. Restarting the test a few times on the CI, I couldn't observe a failure. `next` and `previous` are a bit more efficient than `plus`, maybe that's the thing.

core/common/src/math.kt
```diff
@@ -200,4 +200,17 @@ internal fun multiplyAndAdd(d: Long, n: Long, r: Long): Long {
         mr -= n
     }
     return safeAdd(safeMultiply(md, n), mr)
-}
\ No newline at end of file
+}
+
+// org.threeten.bp.chrono.IsoChronology#isLeapYear
+internal fun isLeapYear(year: Int): Boolean {
```
ilya-g, line 206: This file contains common math functions, not specific to any date calculations. I suggest to find another file for these new ones, or introduce a new one, e.g. dateMath.kt/calendarMath.kt/IsoChronology.kt

dkhalanskyjb: It does define some constants like `internal const val HOURS_PER_DAY = 24` or `NANOS_PER_DAY` already, so it's not just pure math here. I don't think it's worth it yet to split the file, because
* it's relatively short,
* when writing code, I like to look over the available constants, and seeing all of them in one place helps, whereas we would need to move, say, `HOURS_PER_DAY` into the date-specific-math file, keeping `NANOS_PER_ONE` in the pure math one.

ilya-g: Now, when we've got two new functions, it's a good chance to group these constants together with them in a separate file.
NANOS_PER_ONE can also be seen as a time-related constant, not general math.

dkhalanskyjb: Peculiarly, trying to implement this refactoring slows down the code on JS considerably. The test for `fromEpochDays` now fails, and checking on my machine, before the refactoring, it takes 1.2 ± 0.2 seconds, while after the refactoring, it takes 1.6 ± 0.3 seconds. I propose that we revert this change.

ilya-g: Can't confirm such effect, on the opposite the test runs faster for me on the latest commit than on the previous.
I compared generated JS code and didn't find anything suspicious in the difference.

ilya-g: As the last resort, the timeout could be increased if it's essential to test all these values.

dkhalanskyjb: Yeah, let's do that.

core/build.gradle.kts
```diff
@@ -68,7 +68,7 @@ kotlin {
         nodejs {
             testTask {
                 useMocha {
-                    timeout = "5s"
+                    timeout = "30s"
```
ilya-g, line 71: The highest recorded test duration value was 8.6s, so probably 10-15 s should be fine

ilya-g (approved):

core/build.gradle.kts
dkhalanskyjb: Sure, I just don't see the point in this timeout at all. Don't understand what it brings to the table. So, I increased it to the point where it will most certainly not bother us. Do you have some ideas about why this limit could be useful? If so, maybe it makes sense to add it to the other targets as well?

PR #238, opened 2022-10-19
Update gradle wrapper and specify distribution sha256

ilya-g:
The result of executing `gradlew wrapper` command with the current Gradle version.
I've also checked that the wrapper.jar checksum matches the official one for 7.5.1: https://gradle.org/release-checksums/


dkhalanskyjb:
Did you manage to find out when the checksum could be useful for us? I only found this explanation: https://docs.gradle.org/current/userguide/gradle_wrapper.html#wrapper_checksum_verification If I understood correctly, we would need to enable a GitHub action in order for this to have at least some purpose.

ilya-g:
distributionSha256Sum in gradle-wrapper.properties is verified by the wrapper itself each time it downloads Gradle distribution, for example when building on a freshly started agent.
The integrity of the wrapper itself can be verified by setting up a github action.

dkhalanskyjb:
What's the point though? It's already transferred over https, so no one is going to be able to swap the distribution unless
* Gradle's servers are compromised, or
* Someone hacks the DNS record for the gradle server.

ilya-g:
Won't hurt anyway

dkhalanskyjb:
Sure, my point is, maybe we could also enable the GitHub Acton so that this line pulls its weight?

ilya-g:
Continued in #259

Issue #246, opened 2023-01-18
Considering the possibility to represent `0` of some `DateTimeUnit`?


kevincianfarini:
Right now all of the instances of `DateTimeUnit` guard against instantiating with zero of the underlying asset. For example: 

```kt
@Serializable(with = DayBasedDateTimeUnitSerializer::class)
public class DayBased(
    /**
     * The length of this unit in days.
     */
    public val days: Int
) : DateBased() {
    init {
        require(days > 0) { "Unit duration must be positive, but was $days days." }
    }
```

This feels really unintuitive. If someone is dynamically performing math operations of `Instant`, `LocalDate`, etc. it should be permissible to have an identity operation. Currently this requires null coalescing for identity-like operations. Eg: 

```kt
val dateTimeUnit: DateTimeUnit? = /* TODO */
val date = clock.now().toLocalDate(/* omitted */).let { date ->
  if (dateTimeUnit == null) date else date.minus(dateTimeUnit)
} 
```  

I'm hoping to understand the rationale for why zero of some `DateTimeUnit` is generally prohibited. Thank you!

ilya-g:
`DateTimeUnit` is not an alternative for `Duration` or `DateTimePeriod`. If you want to add zero of some `DateTimeUnit`, use the overload that takes an integer and that `DateTimeUnit`.
We're considering deprecating the overloads of `plus` and `minus` with a single `DateTimeUnit` parameter to avoid this confusion, see https://github.com/Kotlin/kotlinx-datetime/issues/95#issuecomment-767375456

kevincianfarini:
That makes more sense, thank you!

PR #247, opened 2023-01-19
Deprecate plus and minus operations taking DateTimeUnit without amount

### LINKED ISSUE #95: Make DateTimeUnit constructors non public
I just got hit by an exception when adding days to a `LocalDate`

I took a look at the function signature:

```kotlin
public actual fun LocalDate.plus(unit: DateTimeUnit.DateBased): LocalDate = plusNumber(1, unit)
```
So I was using `DateTimeUnit.DateBased` to dynamically add days:
```kotlin
loopDate.plus(DateTimeUnit.DateBased.DayBased(fastingDays.size))
```

This was throwing an exception:
```
java.lang.IllegalArgumentException: Unit duration must be positive, but was 0 days.
	at kotlinx.datetime.DateTimeUnit$DateBased$DayBased.<init>(DateTimeUnit.kt:68)
```
On the slack channel I discovered that I should not use the `DayBased` constructor at all and instead use a period:
```kotlin
.plus(DatePeriod(days = 0))
```

I would suggest to mention that in the documentation and make the constructors internal so such usage errors don't happen.

ilya-g:
fixes #95, #246

### DIFF
--- a/core/common/src/Instant.kt
+++ b/core/common/src/Instant.kt
@@ -351,7 +351,7 @@ public fun Instant.minus(other: Instant, timeZone: TimeZone): DateTimePeriod =
  *
  * @throws DateTimeArithmeticException if this value or the result is too large to fit in [LocalDateTime].
  */
-@Deprecated("Use the plus overload with explicit amount of units", ReplaceWith("this.plus(1, unit, timeZone)"))
+@Deprecated("Use the plus overload with an explicit number of units", ReplaceWith("this.plus(1, unit, timeZone)"))
 public expect fun Instant.plus(unit: DateTimeUnit, timeZone: TimeZone): Instant
 
 /**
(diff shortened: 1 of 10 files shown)


core/js/src/LocalDate.kt
```diff
@@ -60,6 +60,7 @@ public actual class LocalDate internal constructor(internal val value: jtLocalDa
     public actual fun toEpochDays(): Int = value.toEpochDay().toInt()
 }
 
+@Deprecated("Use the plus overload with explicit amount of units", ReplaceWith("this.plus(1, unit)"))
```
dkhalanskyjb, line 63:
```suggestion
@Deprecated("Use the plus overload with an explicit amount of units", ReplaceWith("this.plus(1, unit)"))
```
Ditto everywhere.

core/common/src/Instant.kt
```diff
@@ -381,6 +384,7 @@ public fun Instant.plus(unit: DateTimeUnit.TimeBased): Instant =
  *
  * The return value is clamped to the platform-specific boundaries for [Instant] if the result exceeds them.
  */
+@Deprecated("Use the minus overload with explicit amount of units", ReplaceWith("this.minus(1, unit)"))
 public fun Instant.minus(unit: DateTimeUnit.TimeBased): Instant =
     plus(-1L, unit)
```
dkhalanskyjb, line 389: I like the practice we have in coroutines: when deprecating something, we move it either to the bottom of the file or altogether to a different file designated for deprecated things. This way, the deprecated (and no longer all that relevant) things are out of sight. Maybe we could adopt it here as well?

dkhalanskyjb (commented):
Didn't find any problems in the proposed replacements, and the general idea is also good.

core/common/src/Instant.kt
ilya-g: We'll clean up these deprecations near the stabilization, so it's not relevant here. We can consider this practice after stabilization, however I'm not currently fond of it because it makes history digging with git blame harder.

PR #248, opened 2023-01-20
Improved LocalDateTime documentation

holgerbrandl:
Improved reasoning for missing date-time arithmetics and added an example

### DIFF
--- a/core/common/src/LocalDateTime.kt
+++ b/core/common/src/LocalDateTime.kt
@@ -20,7 +20,10 @@ import kotlinx.serialization.Serializable
  * The main purpose of this class is to provide human-readable representations of [Instant] values, or to transfer them
  * as data.
  *
- * The arithmetic on [LocalDateTime] values is not provided, since without accounting for the time zone transitions it may give misleading results.
+ * The arithmetic on [LocalDateTime] values is not provided, since without accounting for the time zone transitions it 
+ * may give misleading results. For example, assuming March 26, 2023, 22:00 as a value of `LocalDateTime`, adding 10 hours
+ * is ambigous because that is when daylight savings kick in in some timzones: For the US, if you just add 10 hours to 
+ * the clock at this particular day, it would be 8:00, but if you just waited 10 hours it would be 9:00.
  */
 @Serializable(with = LocalDateTimeIso8601Serializer::class)
 public expect class LocalDateTime : Comparable<LocalDateTime> {
--- a/core/common/src/LocalDateTime.kt
+++ b/core/common/src/LocalDateTime.kt
@@ -22,8 +22,8 @@ import kotlinx.serialization.Serializable
  *
  * The arithmetic on [LocalDateTime] values is not provided, since without accounting for the time zone transitions it 
  * may give misleading results. For example, assuming March 26, 2023, 22:00 as a value of `LocalDateTime`, adding 10 hours
- * is ambigous because that is when daylight savings kick in in some timzones: For the US, if you just add 10 hours to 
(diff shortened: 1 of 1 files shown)


core/common/src/LocalDateTime.kt
```diff
@@ -20,7 +20,10 @@ import kotlinx.serialization.Serializable
  * The main purpose of this class is to provide human-readable representations of [Instant] values, or to transfer them
  * as data.
  *
- * The arithmetic on [LocalDateTime] values is not provided, since without accounting for the time zone transitions it may give misleading results.
+ * The arithmetic on [LocalDateTime] values is not provided, since without accounting for the time zone transitions it 
+ * may give misleading results. For example, assuming March 26, 2023, 22:00 as a value of `LocalDateTime`, adding 10 hours
+ * is ambigous because that is when daylight savings kick in in some timzones: For the US, if you just add 10 hours to
```
ilya-g, line 25:
```suggestion
 * is ambiguous because that is when daylight savings kick in in some time zones: for the US, if you just add 10 hours to 
```

core/common/src/LocalDateTime.kt
```diff
@@ -20,7 +20,10 @@ import kotlinx.serialization.Serializable
  * The main purpose of this class is to provide human-readable representations of [Instant] values, or to transfer them
  * as data.
  *
- * The arithmetic on [LocalDateTime] values is not provided, since without accounting for the time zone transitions it may give misleading results.
+ * The arithmetic on [LocalDateTime] values is not provided, since without accounting for the time zone transitions it 
+ * may give misleading results. For example, assuming March 26, 2023, 22:00 as a value of `LocalDateTime`, adding 10 hours
+ * is ambigous because that is when daylight savings kick in in some timzones: For the US, if you just add 10 hours to 
+ * the clock at this particular day, it would be 8:00, but if you just waited 10 hours it would be 9:00.
```
ilya-g, line 26:
```suggestion
 * the clock at this particular day, it would be 8:00, but if you just waited 10 hours, it would be 9:00.
```

holgerbrandl:
I figured later that this is greatly documented in https://github.com/Kotlin/kotlinx-datetime#date--time-arithmetic but maybe extending the javadoc makes sense as well. 

What I also tend to do in my projects is to bluntly cross reference the main documentation via hyperlinks from the java-doc to keep docs centralized while avoiding redundancy. For example https://github.com/holgerbrandl/kalasim/blob/master/src/main/kotlin/org/kalasim/Component.kt#L131. I still wonder about the best practice here.

dkhalanskyjb:
What do you think about writing something like this instead?
```kotlin
* See the documentation [LocalDateTime.toInstant] for an explanation of problems with raw [LocalDateTime] values.
```
It's nice to have information discoverable, but I also think that it's better to write one good explanation and link to it from ten places instead of writing ten unrelated explanations.

Or maybe it would be better to write one good explanation in some place other than `LocalDateTime.toInstant`. Or maybe the explanation in `LocalDateTime.toInstant` is also not good enough. WDYT?

holgerbrandl:
Currently the actual explanation sits at https://github.com/Kotlin/kotlinx-datetime#date--time-arithmetic, but definitely, keeping it at once place with proper references is better.

dkhalanskyjb:
I know, I'm the one who wrote the explanation https://github.com/Kotlin/kotlinx-datetime/commit/ff43dd650d28c347e74a7200611e3417b1c6b545. I agree with the spirit of this PR though, it would be nice for such things to be discoverable through KDoc only. The only question is, where to put the explanation in a systematic manner.

dkhalanskyjb:
Superseded by #367.

PR #249, opened 2023-02-07
Add `java.time.Clock` converter

GrahamBorland:
Adds a function to convert a `java.time.Clock` to a `kotlinx.datetime.Clock`.

### DIFF
--- a/core/jvm/src/Converters.kt
+++ b/core/jvm/src/Converters.kt
@@ -93,3 +93,11 @@ public fun UtcOffset.toJavaZoneOffset(): java.time.ZoneOffset = this.zoneOffset
  */
 public fun java.time.ZoneOffset.toKotlinUtcOffset(): UtcOffset = UtcOffset(this)
 
+/**
+ * Converts this [java.time.Clock][java.time.Clock] to a [kotlinx.datetime.Clock][Clock].
+ */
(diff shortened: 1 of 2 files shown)


core/jvm/src/Converters.kt
```diff
@@ -93,3 +93,11 @@ public fun UtcOffset.toJavaZoneOffset(): java.time.ZoneOffset = this.zoneOffset
  */
 public fun java.time.ZoneOffset.toKotlinUtcOffset(): UtcOffset = UtcOffset(this)
 
+/**
+ * Converts this [java.time.Clock][java.time.Clock] to a [kotlinx.datetime.Clock][Clock].
+ */
+public fun java.time.Clock.toKotlinClock(): Clock = this.let {
```
hfhbd, line 99: Why do you use let? Just remove it and return the object directly.

ilya-g:
Could you describe cases when such conversion could be useful?

GrahamBorland:
When you have a Java clock (e.g. injected from DI on Android) and you want to pass it to a KMP library using kotlinx-datetime.

dkhalanskyjb:
`kotlinx-datetime` removes `Clock` in favor of `kotlin.time.Clock`, so this is no longer applicable. Filed https://youtrack.jetbrains.com/issue/KT-85461/Add-kotlin.time.TestClock for the stdlib.

Issue #236, opened 2022-10-15
Instant.EPOCH


Nek-12:
This is simple: could we add an equivalent of java.time.Instant.EPOCH? 
This would be a static equivalent of  

```
val Instant.Companion.EPOCH: Instant
    get() = fromEpochMilliseconds(0)
```

elizarov:
What kind of usage do you have in mind for it?

Nek-12:
I'm sometimes modeling last updated property of an object as EPOCH by default because I don't want to make the field nullable. EPOCH would represent a value that would make the model "old" immediately. Since databases or other libraries might not support the exact level of precision that Instant.MIN provides, it's safer to go with EPOCH.

dkhalanskyjb:
See https://github.com/Kotlin/kotlinx-datetime/issues/227. The same discussion applies to your use case: `Instant.EPOCH` is not, logically, a minimum value. Using `Instant.EPOCH` for this would obscure the meaning of the code, leading to doubt about its correctness. "This value was last changed in 1970, that is, never" sounds absurd if you're out of context, even if the reader knows that `EPOCH` is the Unix epoch and not, say, the Apple reference date, which is in 2001, or something else entirely.

Why don't you want the field to be nullable? If there is actually no last-changed value present, then doing anything with it other than updating is likely a mistake. Making it nullable forces you to think about what would happen if you access it when there's nothing there. This would prevent embarrassing bugs like showing `1970-01-01` to your users.

In any case, what you want to write can be written in a more semantically clear fashion, for example, like this:
```kotlin
/**
 * We believe that no changes to models happen earlier than this instant.
 * So, this value is used as a placeholder for "there have never been any changes" in the model-was-last-changed fields.
 */
internal val Instant.Companion.BEFORE_ANY_CHANGES: Instant
    get() = fromEpochSeconds(0)
```

Nek-12:
Okay, it wasn't pleasant to admit, but after some research I have come to the conclusion that I should indeed represent "unchanged yet" values as NULLs, so for my particular use-case, usage of EPOCH does not apply anymore. 

Other things to consider here is parity with the java.time library (which has an EPOCH constant) and other possible use-cases that weren't reported yet.

If there are no plans on implementing this, the issue may be closed

PR #259, opened 2023-03-14
Upgrade Kotlin to 1.8.10 and Gradle to 8.0.2

dkhalanskyjb:
This is done simultaneously because Gradle 8.0.2 wants Kotlin 1.8 metainformation format, and Kotlin 1.8.10 uses the new Gradle API.


ilya-g:
What metainformation format do you mean and where Gradle wants it?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.0.2-bin.zip
 zipStoreBase=GRADLE_USER_HOME
```
ilya-g, line 4: Please also add .gitattributes from #236 (this is to avoid line ending problems in .bat files and fix line endings in all other files to LF) and specify `distributionSha256Sum` property here.

dkhalanskyjb: > this is to avoid line ending problems in .bat files and fix line endings in all other files to LF

Why are they problems? Does this cause any issues? I see value in sticking to the defaults unless this can be helped.

> and specify `distributionSha256Sum` property here.

I'd rather not. Without it, updating the Gradle version is as simple as `./gradlew wrapper --gradle-version 8.0.2`, but if we stick to adding `distributionSha256Sum`, this introduces a bit of friction that may be tiny but is still ultimately useless.

dkhalanskyjb:
The specific error message when upgrading Kotlin but not Gradle:
> Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.8.0, expected version is 1.6.0.

gradle/wrapper/gradle-wrapper.properties
ilya-g: > this introduces a bit of friction 

Still a good practice

> Why are they problems? Does this cause any issues?

I remember there were line ending differences in .bat files last time we updated the wrapper and this could make .bat non-functional on Windows, .gitattributes are intended to unify line ending handling on different platforms and not depend on contributors' autocrlf setting.

ilya-g:
> The specific error message when upgrading Kotlin but not Gradle:

Hmm, haven't noticed that locally when I updated only kotlin version. In general Kotlin 1.8 should be compatible with Gradle 7.2+, so probably errors like that should be investigated. I don't mind upgrading Gradle, if it's needed for some other reasons.

gradle/wrapper/gradle-wrapper.jar
ilya-g, line 1: Please check if the wrapper task doesn't rewrite with different content when running again. It seems that this wrapper jar doesn't correspond to the Gradle 8.0.2 release. Minor, but since we're changing it anyway, makes sense to use the corresponding version.

gradle/wrapper/gradle-wrapper.properties
dkhalanskyjb: > [w.r.t. .gitattributes]

Ok, thank you, that makes sense. Added.

> Still a good practice

I tried to check what the hash does manually, and I couldn't find any special behavior around it other than updating the wrapper itself. I intentionally wrote a wrong hash there, stopped all the Gradle daemons, and removed the `.gradle/` and `~/.gradle` directories, but after that, the build still proceeded as usual. So, what is this good practice good for?

dkhalanskyjb:
Regarding the error message: I managed to get rid of it by doing https://stackoverflow.com/a/75262359/20886821, so it is possible to upgrade Kotlin without upgrading Gradle. However, the issue remains if I specify plugins as dependencies in `buildSrc/build.gradle.kts` (like is done in https://github.com/Kotlin/kotlinx-datetime/pull/257) and not in the `plugins` block in `settings.gradle.kts`.

gradle/wrapper/gradle-wrapper.properties
ilya-g: It checks the integrity of the Gradle dist when downloading it first time, if the check is failed, an error like this is prinited:
```
Downloading https://services.gradle.org/distributions/gradle-8.0.2-bin.zip
...........10%............20%............30%............40%............50%............60%...........70%............80%............90%............100%
Exception in thread "main" java.lang.RuntimeException: Verification of Gradle distribution failed!

Your Gradle distribution may have been tampered with.
Confirm that the 'distributionSha256Sum' property in your gradle-wrapper.properties file is correct and you are downloading the wrapper from a trusted source.

 Distribution Url: https://services.gradle.org/distributions/gradle-8.0.2-bin.zip
Download Location: C:\Users\...\.gradle\wrapper\dists\gradle-8.0.2-bin\25jlreiuz6u3xu2phlpa2vv4m\gradle-8.0.2-bin.zip
Expected checksum: 'f6b8596b10cce501591e92f229816aa4046424f3b24d771751b06779d58c8ec4'
  Actual checksum: 'ff7bf6a86f09b9b2c40bb8f48b25fc19cf2b2664fd1d220cd7ab833ec758d0d7'
```

dkhalanskyjb: After thinking about it, I could imagine a scenario in which this SHA-256 could be helpful: given the recent wave of Github implementing automated checks across repositories (notifications about vulnerabilities, leaked keys, etc.), maybe Github will also notify about a hash mismatch.

ilya-g:
It looks like this frequent failure on Windows agents is somehow correlated with updating to Gradle 8.

ubuntudroid:
Any plans to release Kotlin 1.8 support this for everyone in the meantime? The releases are all marked as pre-release anyway, so IMHO having an issue with Windows builds only shouldn't block a pre-release for everyone.

dkhalanskyjb:
@ubuntudroid does something prevent you from using Kotlin 1.8 with the latest `kotlinx-datetime` release?

ubuntudroid:
@dkhalanskyjb no, I was stumbling on an error message when building with Kotlin 1.8, and it seemed like it pointed at kotlinx-datetime being the culprit at the time of when I wrote that comment. In the end, the issue turned out to be somewhere else. I should have pointed out the exact error I was getting in my earlier comment, and I should have investigated more instead of rashly posting that comment. Sorry! 🙏 

Please take your time with this, and thanks for maintaining this awesome library in your free time! ❤️

ilya-g (approved):

PR #261, opened 2023-03-15
Add all the supported Kotlin/Native targets that we can

### LINKED ISSUE #165: Add linuxArm64 target
This should be a one-line change in `core/build.gradle.kts` and `serialization/build.gradle.kts` considering kotlinx.serialization supports it. Happy to send a PR, if it'd be accepted.

dkhalanskyjb:
Only the Android NDK and the non-ILP32 WatchOS Arm64 targets are not supported, as we depend on the serialization library, which also doesn't support them.

Fixes #165

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -37,24 +37,41 @@ kotlin {
     explicitApi()
 
     infra {
+        // Tiers are in accordance with <https://kotlinlang.org/docs/native-target-support.html>
+        // Tier 1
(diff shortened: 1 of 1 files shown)


dkhalanskyjb:
Looks like this only adds the `linuxArm64` target, the rest is blocked by https://github.com/Kotlin/kotlinx.serialization/pull/2216

ilya-g (approved):

PR #264, opened 2023-03-28
Fix adding small `Duration` to large `Instant` on JS and Native

### LINKED ISSUE #263: Bug in `Instant.plus(Duration)` on JS for large `Instant`s and small `Duration`s
Running this code on JS produces an unexpected result
```kt
val instant = Instant.fromEpochMilliseconds(Long.MAX_VALUE) + 999_999.nanoseconds
println(instant) // expected some positive (clamped) value, got -1000000-01-01T00:00:00Z
```
The reason for this seems to be [this line of code](https://github.com/Kotlin/kotlinx-datetime/blob/cd6dcd1abb547bc90455471e644e1e8fe2a49e5c/core/js/src/Instant.kt#L38).
1) `Instant.fromEpochMilliseconds(Long.MAX_VALUE)` gives the max instant on JS (`+1000000-12-31T23:59:59.999999999Z`)
2) [`plusFix`](https://github.com/Kotlin/kotlinx-datetime/blob/cd6dcd1abb547bc90455471e644e1e8fe2a49e5c/core/js/src/Instant.kt#LL42C18-L42C25) overflows and throws a `DateTimeException`
3) [here](https://github.com/Kotlin/kotlinx-datetime/blob/cd6dcd1abb547bc90455471e644e1e8fe2a49e5c/core/js/src/Instant.kt#L38) is the actual bug: because the added amount is smaller than one second, `seconds` is `0`, resulting in `Instant.MIN` being returned

The line should probably be something like this instead:
```kt
if (duration.isPositive()) MAX else MIN
```

Lukellmann:
When adding a positive `Duration` < 1 second to an `Instant` far in the future for which the addition implementation overflows and throws an exception, the `catch` blocks would only check if the `seconds` component of the `Duration` is `> 0`. However since the `Duration` is < 1 second, the `seconds` component is not `> 0` and `Instant.MIN` is returned instead of `Instant.MAX`.

This was fixed by replacing the check against the seconds component of the `Duration` with a check against the whole `Duration` like it was already done in the JVM implementation.

fixes #263

### DIFF
--- a/core/common/test/InstantTest.kt
+++ b/core/common/test/InstantTest.kt
@@ -609,4 +609,13 @@ class InstantRangeTest {
         (maxValidInstant + 1.nanoseconds).until(maxValidInstant, DateTimeUnit.NANOSECOND)
         maxValidInstant.until(maxValidInstant + 1.nanoseconds, DateTimeUnit.NANOSECOND)
     }
+
+    // https://github.com/Kotlin/kotlinx-datetime/issues/263
+    @Test
+    fun addSmallDurationsToLargeInstants() {
+        for (smallDuration in listOf(1.nanoseconds, 999_999.nanoseconds, 1.seconds - 1.nanoseconds)) {
+            assertEquals(expected = Instant.MAX, actual = Instant.MAX + smallDuration)
+            assertEquals(expected = Instant.MIN, actual = Instant.MIN - smallDuration)
+        }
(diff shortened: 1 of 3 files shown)


ilya-g (approved):

ilya-g:
Thanks for the fix!

PR #278, opened 2023-05-30
Enabled target watchosDeviceArm64

### LINKED ISSUE #277: Enable target watchosDeviceArm64
**kotlinx.serialization** currently [has support](https://github.com/Kotlin/kotlinx.serialization/issues/2210) for watchosDeviceArm64. According to the comments on the code I believe we didn't have it on **kotlinx-datetime** because of it.

It would be great if you could launch a new version of **kotlinx-datetime** with this support 🙂

cmota:
Updated Kotlin version to 1.8.21
Updated kotlinx.serialization to 1.5.1
Enabled watchosDeviceArm64 target (which is currently supported by kotlinx.serialization)

It should fix #277 🙂


severn-everett:
You should update the image badge in the README file so that it indicates the updated Kotlin version.

hfhbd:
The android targets are also added with serialization 1.5.1.

cmota:
I believe you're referring to:
- androidNativeArm32
- androidNativeArm64
- androidNativeX86
- androidNativeX64

right @hfhbd ?

Should I also add them to the PR @severn-everett ? 🙂

hfhbd:
Yes, I mean these targets but maybe adding these targets isn't enough and you might need to implement some native actual code.

ilya-g (approved):

ilya-g:
Thanks, merged

PR #279, opened 2023-06-06
Adding Clock lambda helper

### LINKED ISSUE #273: Clock interface should be a "fun" interface
Since it is a SAM should be declared as a fun interface to allow using a lambda to provide an implementation.

severn-everett:
Added helper function to create a Clock instance by passing in a lambda to be invoked for Clock.now(); Updated Kotlin shield in README to reflect usage of version 1.8.10

Fixes #273


dalewking:
I prefer the idea of just making Clock be a "fun" interface which accomplishes the same thing.

dkhalanskyjb:
`kotlinx-datetime` removes `Clock` in favor of `kotlin.time.Clock`, so this is no longer applicable. In any case, in my opinion, the presented use case is better covered by https://youtrack.jetbrains.com/issue/KT-85461/Add-kotlin.time.TestClock

PR #285, opened 2023-06-23
Make helper_macros.hpp self-contained.

mpetrov:
Previously this header relied on include order to work correctly, as it didn't include all required headers.


dkhalanskyjb:
No need to fix this. The whole file is going away soon.

PR #286, opened 2023-06-26
Implement the timezone database for Linux and Windows in Kotlin

### LINKED ISSUE #262: What's about Windows 7?
We have got the report in our project: https://github.com/akuleshov7/ktoml/issues/193 that kotlinx-datetime native has some problems on windows 7. Any solution for that?

dkhalanskyjb:
Fixes #262


core/native/src/internal/TimeZoneRules.kt
```diff
@@ -0,0 +1,388 @@
...
+ * Copyright 2019-2023 JetBrains s.r.o. and contributors.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.datetime.internal
+
+import kotlinx.datetime.*
+import kotlin.math.*
+
+internal sealed interface OffsetInfo
```
ilya-g, line 11: This file has too many classes. Could you extract hierarchies like `OffsetInto` and `DateOfYear` to separate files?
Also consider using nested classes, because otherwise it is unclear, for example, where `kotlinx.datetime.internal.Gap` belongs to.

core/nix/test/Util.kt
```diff
@@ -0,0 +1,158 @@
...
+ */
+
+package kotlinx.datetime.test
+
+// od --format=x1 --output-duplicates --address-radix=n --width=16 /usr/share/zoneinfo/Europe/Berlin |
+// sed -e 's/\b\(\w\)/0x\1/g' -e 's/\(\w\)\b/\1,/g'
+// Do not remove the type annotation, otherwise the compiler slows down to a crawl for this file even more.
+// This constant is in a separate file to avoid recompiling it on every change to the test file, which is slow to the
+// point of freezing the IDE.
+internal val EuropeBerlinTzFile = listOf<Int>(
```
ilya-g, line 13: Makes sense to specify the version in the name.

core/nix/src/internal/Tzfile.kt
```diff
@@ -0,0 +1,353 @@
...
+private fun BinaryDataReader.readPosixTzString(): PosixTzString? {
+    var c = readAsciiChar()
+    fun readName(): String? {
+        if (c == '\n') return null
+        var name = ""
+        if (c == '<') {
+            c = readAsciiChar()
+            while (c != '>') {
+                check(c.isLetterOrDigit() || c == '-' || c == '+') { "Invalid char '$c' in the std name in POSIX TZ string" }
+                name += c
```
ilya-g, line 211: How long is usually a name constructed this way?

dkhalanskyjb: This is for abbreviations, so three or four ASCII chars.

ilya-g: I'd suggest to use a StringBuilder then.

core/build.gradle.kts
```diff
@@ -318,11 +298,13 @@ val downloadWindowsZonesMapping by tasks.registering {
         xmlDoc.documentElement.normalize()
         val mapZones = xmlDoc.getElementsByTagName("mapZone")
         val mapping = linkedMapOf<String, String>()
+        mapping["UTC"] = "UTC"
         for (i in 0 until mapZones.length) {
             val mapZone = mapZones.item(i)
             val windowsName = mapZone.attributes.getNamedItem("other").nodeValue
             val usualNames = mapZone.attributes.getNamedItem("type").nodeValue
```
ilya-g, line 305: I've noticed that each `mapZone` also contains `territory` attribute. Can it potentially be used to make Windows->IANA mapping less ambiguous?

dkhalanskyjb: I don't see how. AFAIK, the only pieces of information about the current time zone are those in this picture: https://static1.makeuseofimages.com/wordpress/wp-content/uploads/2022/11/Change-Time-Zone-Using-Registry-Editor.jpg?q=50&fit=crop&w=1500&dpr=1.5 I don't see any territorial markings here.

core/nix/src/internal/Tzfile.kt
dkhalanskyjb: Ok. Is this a matter of style? Because I didn't see any difference in performance.

core/build.gradle.kts
```diff
@@ -118,6 +118,9 @@ kotlin {
         kotlin.srcDir("$targetName/${suffix ?: "src"}")
         resources.srcDir("$targetName/${suffix?.let { it + "Resources" } ?: "resources"}")
         languageSettings {
+            compilerOptions{
```
ilya-g, line 121: This DSL block doesn't belong to this scope. I fixed it in a different way in #319, however I can try to use global compilerOptions for that.

core/common/src/DayOfWeek.kt
```diff
@@ -25,8 +23,7 @@ public expect enum class DayOfWeek {
  */
 public val DayOfWeek.isoDayNumber: Int get() = ordinal + 1
 
-@SharedImmutable
-private val allDaysOfWeek = DayOfWeek.values().asList()
+private val allDaysOfWeek = DayOfWeek.entries
```
ilya-g, line 26: Please revert enum entires from this PR, it is fixed more thoroughly in #319

ilya-g (changes requested):

core/common/src/DayOfWeek.kt
dkhalanskyjb: Sure, I'll have to do that while rebasing on top of #319 anyway.

core/build.gradle.kts
```diff
@@ -110,11 +113,14 @@ kotlin {
     sourceSets.all {
         val suffixIndex = name.indexOfLast { it.isUpperCase() }
         val targetName = name.substring(0, suffixIndex)
-        val suffix = name.substring(suffixIndex).lowercase().takeIf { it != "main" }
+        val suffix = name.substring(suffixIndex).toLowerCase().takeIf { it != "main" }
```
ilya-g, line 116: Better not to use `toLowerCase` when `lowercase` is available

ilya-g (approved):
Ok, I think it's good to merge. Please squash some intermediate changes.

core/build.gradle.kts
```diff
@@ -328,8 +309,8 @@ tasks {
 
 val downloadWindowsZonesMapping by tasks.registering {
     description = "Updates the mapping between Windows-specific and usual names for timezones"
-    val output = "$projectDir/native/cinterop/public/windows_zones.hpp"
-    val initialFileContents = File(output).readBytes()
+    val output = "$projectDir/windows/src/WindowsZoneNames.kt"
+    val initialFileContents = try { File(output).readBytes() } catch(e: Throwable) { ByteArray(0) }
```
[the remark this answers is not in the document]
dkhalanskyjb: You're right, this does sometimes lead to the file being read when it doesn't need to be, e.g. when invoking `gradle tasks`. Good job noticing this!

PR #288, opened 2023-06-27
Set timezone on NSDateFormatter explicitly

MarkCMann:
Fix test case in US timezones by explicitly formatting date with the UTC timezone


dkhalanskyjb:
Thanks!

PR #308, opened 2023-10-03
Reduce the number of suppressions for internal access 

qwwdfsad:
While backporting K2 commits, I've spotted one related to `internal` declarations.

Instead of backporting it, I decided to tide it up in the face of https://youtrack.jetbrains.com/issue/KT-60866/Phase-out-usages-of-SuppressINVISIBLEREFERENCE-INVISIBLEMEMBER-in-libraries

### DIFF
--- a/core/common/src/serializers/DateTimeUnitSerializers.kt
+++ b/core/common/src/serializers/DateTimeUnitSerializers.kt
@@ -23,7 +23,7 @@ public object TimeBasedDateTimeUnitSerializer: KSerializer<DateTimeUnit.TimeBase
 
     // https://youtrack.jetbrains.com/issue/KT-63939
     override val descriptor: SerialDescriptor by lazy(LazyThreadSafetyMode.PUBLICATION) {
-        buildClassSerialDescriptor("TimeBased") {
+        buildClassSerialDescriptor("kotlinx.datetime.TimeBased") {
             element<Long>("nanoseconds")
         }
     }
@@ -69,7 +69,7 @@ public object DayBasedDateTimeUnitSerializer: KSerializer<DateTimeUnit.DayBased>
 
     // https://youtrack.jetbrains.com/issue/KT-63939
     override val descriptor: SerialDescriptor by lazy(LazyThreadSafetyMode.PUBLICATION) {
-        buildClassSerialDescriptor("DayBased") {
+        buildClassSerialDescriptor("kotlinx.datetime.DayBased") {
             element<Int>("days")
         }
     }
@@ -115,7 +115,7 @@ public object MonthBasedDateTimeUnitSerializer: KSerializer<DateTimeUnit.MonthBa
 
     // https://youtrack.jetbrains.com/issue/KT-63939
     override val descriptor: SerialDescriptor by lazy(LazyThreadSafetyMode.PUBLICATION) {
-        buildClassSerialDescriptor("MonthBased") {
+        buildClassSerialDescriptor("kotlinx.datetime.MonthBased") {
             element<Int>("months")
         }
     }
--- a/core/common/src/serializers/DateTimeUnitSerializers.kt
+++ b/core/common/src/serializers/DateTimeUnitSerializers.kt
@@ -101,7 +101,7 @@ public object DayBasedDateTimeUnitSerializer: KSerializer<DateTimeUnit.DayBased>
                 }
             }
         }
-        if (!seen) throw MissingFieldException(missingField = "days", serialName = TimeBasedDateTimeUnitSerializer.descriptor.serialName)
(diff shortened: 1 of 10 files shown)


core/common/src/serializers/DateTimeUnitSerializers.kt
```diff
@@ -48,12 +47,12 @@ public object TimeBasedDateTimeUnitSerializer: KSerializer<DateTimeUnit.TimeBase
...
                         }
                         CompositeDecoder.DECODE_DONE -> break@loop // https://youtrack.jetbrains.com/issue/KT-42262
-                        else -> throw UnknownFieldException(elementIndex)
+                        else -> throwUnknownIndexException(elementIndex)
                     }
                 }
             }
         }
-        if (!seen) throw MissingFieldException("nanoseconds")
+        if (!seen) throw MissingFieldException(missingField = "nanoseconds", serialName = descriptor.serialName)
```
qwwdfsad, line 55: Explicit parameter names so one can be sure `missingField` and `serialName` (both are strings) are in the right order

core/common/src/serializers/DayOfWeekSerializers.kt
```diff
@@ -16,14 +16,7 @@ import kotlinx.serialization.internal.*
...
-
-    override val descriptor: SerialDescriptor
-        get() = impl.descriptor
-
-    override fun deserialize(decoder: Decoder): DayOfWeek = impl.deserialize(decoder)
-
-    override fun serialize(encoder: Encoder, value: DayOfWeek): Unit = impl.serialize(encoder, value)
-}
+public object DayOfWeekSerializer : KSerializer<DayOfWeek> by createEnumSerializer<DayOfWeek>(
+    "kotlinx.datetime.serializers.DayOfWeek",
```
qwwdfsad, line 20: Note that it is technically a breaking change: it provide FQ name as serial name (as we do and recommend doing in the serialization) and also replaces `Month` (which is a copy-paste error I believe)

ilya-g:
Please exclude InlineOnly annotation removal, there's no problems with it in K2, as I was told in KT-60866

qwwdfsad:
Done

core/common/src/serializers/DateTimeUnitSerializers.kt
```diff
@@ -98,12 +96,12 @@ public object DayBasedDateTimeUnitSerializer: KSerializer<DateTimeUnit.DayBased>
...
                         }
                         CompositeDecoder.DECODE_DONE -> break@loop // https://youtrack.jetbrains.com/issue/KT-42262
-                        else -> throw UnknownFieldException(elementIndex)
+                        else -> throwUnknownIndexException(elementIndex)
                     }
                 }
             }
         }
-        if (!seen) throw MissingFieldException("days")
+        if (!seen) throw MissingFieldException(missingField = "days", serialName = TimeBasedDateTimeUnitSerializer.descriptor.serialName)
```
ilya-g, line 104: This looks suspicious. Is it intended to reference the descriptor of another type?

core/common/src/serializers/DayOfWeekSerializers.kt
ilya-g: By what principle this FQ name is built? Is it the name of the serialized type?

ilya-g (changes requested):

core/common/src/serializers/DateTimeUnitSerializers.kt
qwwdfsad: No, it's not. Thanks!

core/common/src/serializers/DayOfWeekSerializers.kt
qwwdfsad: According to the documentation:

>For generated serializers, serial name is equal to the corresponding class's fully-qualified name or, if overridden, SerialName. Custom serializers should provide a unique serial name that identify both the serializable class and the serializer itself, ignoring type arguments, if they are present.

We consider this one as "the default generated serializer" in fact, thus FQN of the serialized type.
I also opened https://github.com/Kotlin/kotlinx.serialization/pull/2581 to make it clear

qwwdfsad:
@ilya-g sorry for the delay, it got lost from my radar. PTAL

core/common/src/serializers/DayOfWeekSerializers.kt
ilya-g: "FQN of the serialized type" - ok, then it should be changed as there is no type with this FQN :) 
Perhaps to "kotlinx.datetime.DayOfWeek"

qwwdfsad: Changed it consistently in every serializer, PTAL

ilya-g (approved):

PR #309, opened 2023-10-03
Update Kotlin to 1.9.10

qwwdfsad:


core/build.gradle.kts
```diff
@@ -88,7 +88,7 @@ kotlin {
 
     }
 
-    js {
+    js(IR) {
```
lppedd, line 91: @FenstonSingel can I ask why you specified IR explicitly? Isn't the default IR now?
Just curious as in my libraries I'm now omitting it.

ilya-g: Yes, it shouldn't be needed. Instead, the gradle property `kotlin.js.compiler=both` should be removed.

qwwdfsad:
@ilya-g PTAL

kotlin-js-store/yarn.lock
ilya-g, line 1: Should it be regenerated after removal?

ilya-g (changes requested):

ilya-g:
Obsoleted by #318

PR #310, opened 2023-10-03
Update kotlinx.team.infra version

ilya-g:
- Update TC settings scripts version
- Cleanup TC settings
- Cleanup deprecations in gradle build scripts


dkhalanskyjb (approved):
Sure, makes sense.

PR #315, opened 2023-11-15
Wasm

igoriakovlev:
This is `WasmJs` implementation target.

We update kotlin version to 1.9.20 and implement `WasmJs` target similar to `JS` target implementation.

The `experimental` repository dependency will be removed before merging and replaced with maven central as soon as `kotlinx.serialization` for `WasmJs` target become published.


core/build.gradle.kts
```diff
@@ -196,20 +200,41 @@ kotlin {
             }
         }
 
-        val jsMain by getting {
+        val jsAndWasmSharedMain by creating {
```
ilya-g, line 203: I'd use a different name for this source set, e.g. `jsWasmJs` similar to how in stdlib we use `nativeWasm`

core/build.gradle.kts
```diff
@@ -386,3 +411,14 @@ tasks.withType<AbstractDokkaLeafTask>().configureEach {
         }
     }
 }
+
+tasks.whenTaskAdded {
```
ilya-g, line 415: Why it's required to use a callback?

core/jsAndWasmShared/src/Date.kt
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright 2014-2023 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.datetime
+
+/**
+ * Exposes the [Date API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) to Kotlin.
+ */
+public external class Date() {
```
ilya-g, line 10: We definitely do not want to have `Date` class in `kotlinx.datetime` package with this API.

gradle.properties
```diff
@@ -5,16 +5,15 @@ group=org.jetbrains.kotlinx
 version=0.4.1
 versionSuffix=SNAPSHOT
 
-defaultKotlinVersion=1.8.21
+defaultKotlinVersion=1.9.20
 dokkaVersion=1.8.20
-serializationVersion=1.5.1
+serializationVersion=1.6.1-wasm1
```
ilya-g, line 10: I don't think it's ok to expose this version transitively

core/build.gradle.kts
igoriakovlev: This name was suggested by reviewers in `kotlinx.coroutines` in sake of readability between shared code between `js` and `wasm` sourceset and `wasmJs` target sourceset. But I can rename it to `jsWasmJs ` if you see this is still better.

core/jsAndWasmShared/src/Date.kt
igoriakovlev: Do you suggest to remove `Date` dependent API from Wasm target?

gradle.properties
igoriakovlev: Please comment what do you mean by expose transitively

core/build.gradle.kts
igoriakovlev: This callback allows to disable task when it added. Such tasks added lazily by kotlin.gradle plugin so I think it is safer to make it via callback.

ilya-g: I believe usual task configuration would be ok.
And why do you need to disable it in the first place?

igoriakovlev: The metadata compiler task cannot compile intermediate `js-wasm` shared sourceset (no `js-wasm stdlib`).

I tried to replace it by `getByName` but task is not exists in this script execution point so it failed.

ilya-g: > The metadata compiler task cannot compile intermediate

What is the problem? Some missing dependencies?

ilya-g: > I tried to replace it by getByName but task is not exists in this script execution point so it failed.

Can you get a definitive answer for that from KMP team? I was under impression that all compilation tasks are created as soon as targets and source sets are configured.

core/jsAndWasmShared/src/Date.kt
ilya-g: I'm not sure about its use cases for WasmJs target, but for JS this change would be breaking and also contradicting the interop scenarios it was developed for.

igoriakovlev: I pushed the variant how to escape this issue. Please take a look.

core/build.gradle.kts
igoriakovlev: Asking `KMP team` that does not exists anymore sounds more difficult than replace `whenTaskAdded` with `configureEach` that works as well. Is it ok for you?

core/jsAndWasmShared/src/kotlinx.datetime.internal.JSJoda.module_@js-joda_core.kt
```diff
@@ -68,7 +68,6 @@ external open class ValueRange : InteropInterface {
 external open class TemporalAmount : InteropInterface {
     open fun <T : Temporal> addTo(temporal: T): T
     open fun get(unit: TemporalUnit): Int
-//    open fun units(): JsArray<TemporalUnit>
```
igoriakovlev: I removed some methods from datetime js public API. This API is kotlin layer over jsJoda. Is this API needed to preserve?

core/build.gradle.kts
```diff
@@ -51,8 +50,6 @@ kotlin {
         target("androidNativeX86")
         target("androidNativeX64")
          */
-        // Tier 4 (deprecated, but still in demand)
-        target("linuxArm32Hfp")
```
ilya-g, line 55: Needs to be restored.

core/jsAndWasmShared/src/PlatformSpecifics.kt
```diff
@@ -0,0 +1,13 @@
...
+ */
+package kotlinx.datetime
+
+internal expect fun getAvailableZoneIdsSet(): Set<String>
+
+public expect interface InteropInterface
+
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.FILE)
+public expect annotation class JsNonModule()
```
ilya-g, line 13: Why it is required to typealias JsNonModule? It should be available in common stdlib together with JsModule.

core/wasmJs/src/PlatformSpecifics.kt
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2019-2023 JetBrains s.r.o. and contributors.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+package kotlinx.datetime
+
+import kotlinx.datetime.internal.JSJoda.ZoneId
+
+internal actual fun getAvailableZoneIdsSet(): Set<String> {
+    val result = mutableSetOf<String>()
```
ilya-g, line 10: better to use `buildSet`

ilya-g (changes requested):

core/jsAndWasmShared/src/LocalDate.kt
```diff
@@ -68,23 +68,27 @@ public actual fun LocalDate.plus(value: Long, unit: DateTimeUnit.DateBased): Loc
 
 private fun LocalDate.plusNumber(value: Number, unit: DateTimeUnit.DateBased): LocalDate =
         try {
-            when (unit) {
-                is DateTimeUnit.DayBased -> this.value.plusDays(value.toDouble() * unit.days)
-                is DateTimeUnit.MonthBased -> this.value.plusMonths(value.toDouble() * unit.months)
-            }.let(::LocalDate)
+            jsTry {
+                when (unit) {
+                    is DateTimeUnit.DayBased -> this.value.plusDays(value.toDouble() * unit.days)
```
dkhalanskyjb, line 73: The `jsTry` change is extremely intrusive and pollutes all JS code. If I understand correctly what's going on, this is a measure to support the exceptions thrown from pure JS (not Kotlin/JS) code, which seems incompatible with how Kotlin/WASM treats exceptions. Is this correct?

If so, we'd need to remember to surround all the JS code we call with such checks. It is easy to miss, and I don't even know what will happen if we do miss this. How badly will something break? This question needs to be answered in form of a comment somewhere in the code, no matter what solution we end up with.

Here's an idea of how this could be done in a more robust manner:

```kotlin
// jsAndWasmShared
internal expect class JodaTimeLocalDate {
  fun plusDays(days: Number): Number
  // ...
}

class LocalDateTime(value: JodaTimeLocalDate) // instead of `value: jtLocalDate`

// js
internal actual typealias JodaTimeLocalDate = jtLocalDate

// wasm
internal actual value class JodaTimeLocalDate(val value: jtLocalDate) {
  fun plusDays(days: Number): Number = JodaTimeLocalDate(jsTry { value.plusDays(days) })
  // ...
}
```

This way, the shared source set will stay almost the same, and all WASM-specific logic will be in one place where it's easy to review and spot if something's wrong.

What do you think? Is this approach viable?

core/jsAndWasmShared/src/TimeZone.kt
```diff
@@ -19,9 +19,9 @@ public actual open class TimeZone internal constructor(internal val zoneId: Zone
     public actual fun LocalDateTime.toInstant(): Instant = toInstant(this@TimeZone)
 
     override fun equals(other: Any?): Boolean =
-            (this === other) || (other is TimeZone && this.zoneId == other.zoneId)
+            (this === other) || (other is TimeZone && (this.zoneId === other.zoneId || this.zoneId.equals(other.zoneId)))
```
dkhalanskyjb, line 22: With the approach I described w.r.t. `jsTry`, this could also be hidden in the implementation of `equals` for the WASM-specific wrapper classes without polluting the shared code.

core/jsAndWasmShared/src/LocalDateTime.kt
```diff
@@ -25,34 +27,36 @@ public actual class LocalDateTime internal constructor(internal val value: jtLoc
     public actual constructor(date: LocalDate, time: LocalTime) :
             this(jtLocalDateTime.of(date.value, time.value))
 
-    public actual val year: Int get() = value.year().toInt()
-    public actual val monthNumber: Int get() = value.monthValue().toInt()
+    public actual val year: Int get() = value.year()
+    public actual val monthNumber: Int get() = value.monthValue()
```
dkhalanskyjb, line 31: Why are some casts added and some removed?

dkhalanskyjb (changes requested):

core/build.gradle.kts
dkhalanskyjb: I think `jsAndWasmSharedMain` is more readable.

core/jsAndWasmShared/src/PlatformSpecifics.kt
igoriakovlev: `JsNonModule` is JS only.

core/jsAndWasmShared/src/LocalDateTime.kt
igoriakovlev: The js-joda external declarations were updated - now there is precise parameter and return types were specified  (instead of `Number`).  It is done because `Wasm` do not have interop for `Number`. This change turned out that some of casts are become redundant and some casts became necessary.

core/jsAndWasmShared/src/LocalDate.kt
igoriakovlev: Yeah, when you commonize a code between two platforms with different interop it always looks tricky. This is a price for commonization. 

Yes. In kotlin.wasm platform we have no ability (so far) to catch exceptions thowed from javascript code. This is why I put special wrapper to catch these exceptions inside js world and rethrow it as wasm exceptions in wasm world. For js platform its just a nop.
When I was writing this I had a simple mental rule - every time you want to try-catch exception from js library you should wrap `jsTry` this block.

Your suggest is indeed more readable from a perspective of `jsTry` but one more indirection level makes a mess and boilerplate.

There is several cons for this:
1) Now we have two `Instant` types - `DateTime.Instant` and `jsJoda.Instant` external. With this approach we will have three similar types - `DateTime.Instant`, `jsJoda.Instant` and `WasmWrapper.Instant`.
2) Everywhere we use joda externals we need now make 2 inbox operations, like `jtDuration.between(other.value.jodaInstant, this.value.jodaInstant)`.
3) Everywhere we pass `jsJoda` declarations to `DateTime` declarations we need to pack twice.
4) You need to overlap almost all Joda external hierarchy.
5) It does not solves major problem, because you still need to know which methods should be wrapped with jsTry and which is not. 

Now you can formulate a common rule - any place I put `try-catch` for js exception whould be wrapped with `jsTry` also. Not sure that wrappers can do better.

But:
If you still think that it can do better, I can try to do that but for now it does't seem any better for me.

core/jsAndWasmShared/src/PlatformSpecifics.kt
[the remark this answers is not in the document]
igoriakovlev: Fixed

core/build.gradle.kts
igoriakovlev: Fixed

core/wasmJs/src/PlatformSpecifics.kt
igoriakovlev: Fixed

build.gradle.kts
```diff
@@ -22,9 +22,16 @@ allprojects {
...
     tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
         // outputs the compiler version to logs so we can check whether the train configuration applied
         kotlinOptions.freeCompilerArgs += "-version"
+        kotlinOptions.freeCompilerArgs += "-Xexpect-actual-classes"
     }
 }
+
+
+tasks.withType<org.jetbrains.kotlin.gradle.targets.js.npm.tasks.KotlinNpmInstallTask>().configureEach {
+    args.add("--ignore-engines")
```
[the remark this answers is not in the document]
igoriakovlev: Fixed

core/build.gradle.kts
```diff
@@ -391,3 +413,15 @@ tasks.withType<AbstractDokkaLeafTask>().configureEach {
...
+tasks.configureEach {
+    if (name == "compileJsAndWasmSharedMainKotlinMetadata") {
+        enabled = false
+    }
+}
+
+with(org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin.apply(rootProject)) {
+    nodeVersion = "21.0.0-v8-canary202309167e82ab1fa2"
+    nodeDownloadBaseUrl = "https://nodejs.org/download/v8-canary"
+}
```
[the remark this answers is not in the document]
igoriakovlev: Fixed

core/common/src/serializers/DateTimeUnitSerializers.kt
```diff
@@ -21,8 +21,10 @@ import kotlin.reflect.KClass
  */
 public object TimeBasedDateTimeUnitSerializer: KSerializer<DateTimeUnit.TimeBased> {
 
-    override val descriptor: SerialDescriptor = buildClassSerialDescriptor("TimeBased") {
-        element<Long>("nanoseconds")
+    override val descriptor: SerialDescriptor by lazy {
```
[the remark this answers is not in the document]
igoriakovlev: This bug observable only for wasm target (which seems is more strict for nullability in this case). Could you please to suggest how to file such bug and where.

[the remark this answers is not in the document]
dkhalanskyjb: I don't know how nullability comes into play here: these are all not null and don't even depend on one another.

I'd file this under https://kotl.in/issue, specifying that the behavior is different on wasm and crashes the code that works on all other platforms.

core/build.gradle.kts
ilya-g: I have a better idea - `commonJs`. It's a code shared by Js flavors, be it classic Js or Wasm-Js.

core/jsAndWasmShared/src/PlatformSpecifics.kt
ilya-g: Ok, probably the same thing is with `JsModule` then - it's also Js only and its usage is red in the IDE in commonJs source set.

core/jsAndWasmShared/src/LocalDate.kt
ilya-g: I'm ok with the current jsTry approach, we could refactor it later if we find it disadvantageous.

core/build.gradle.kts
ilya-g: I still see it missing here. I believe we can only restore it after kotlinx-serialization with this target is released. I left this comment not to forget about it.

core/jsAndWasmShared/src/LocalDate.kt
igoriakovlev: Plus it is hard to actualize class with value class. Moreover actual class cannot be open so you cannot build hierarchy with value classes. I believe that code become more and more unreadable if we wish to escape using jsTry with suggested approach. As far I do this refactoring as more hard it to read and support.

core/js/test/JsConverterTest.kt
```diff
@@ -22,7 +22,7 @@ class JsConverterTest {
     @Test
     fun toInstantTest() {
         val kotlinReleaseEpochMilliseconds = 1455494400000
-        val releaseDate = Date(milliseconds = kotlinReleaseEpochMilliseconds)
+        val releaseDate = Date(milliseconds = kotlinReleaseEpochMilliseconds.toDouble())
```
ilya-g, line 25: This is likely no longer needed, since we reverted the approach with expect Date

core/common/src/serializers/DateTimeUnitSerializers.kt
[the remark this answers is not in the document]
igoriakovlev: The initialisation sequence reads `null` value from `not-null` uninitialised property. Here, I put the code into lazy block to prevent reading uninitialised property. The bug exists at all platforms but seems it is not observable.

core/jsAndWasmShared/src/LocalDateTime.kt
igoriakovlev: Here I made half of work to separate Joda API from external declarations. This code will be doubled to provide support for the Wasm.
https://github.com/Kotlin/kotlinx-datetime/commit/e78597df98eb0a0aade70f43e52c40299ce5ddd8

core/jsAndWasmShared/src/LocalDate.kt
igoriakovlev: Here I made half of work to separate Joda API from external declarations. This code will be doubled to provide support for the Wasm.
https://github.com/Kotlin/kotlinx-datetime/commit/e78597df98eb0a0aade70f43e52c40299ce5ddd8

core/js/test/JsConverterTest.kt
igoriakovlev: Fixed

core/jsAndWasmShared/src/PlatformSpecifics.kt
igoriakovlev: Fixed

core/jsAndWasmShared/src/LocalDate.kt
dkhalanskyjb: > You cant make typealias actual in js for these external declarations.

If this is true, I agree that this is a deal-breaker. The ridiculous boilerplate would be a greater evil than `jsTry`.

But I don't understand why this is true. I just tried it in your `wasm_add_layer` branch, and it seems to work:

```kotlin
// core/jsAndWasmShared/src/JsJodaPlatform.kt
public expect class JodaTimeZoneId {
    public fun id(): String
    public fun normalized(): JodaTimeZoneId

    public companion object {
        public fun systemDefault(): JodaTimeZoneId
        public fun of(zoneId: String): JodaTimeZoneId
    }
}

// core/js/src/JsJodaPlatform.kt
actual typealias JodaTimeZoneId = kotlinx.datetime.internal.JSJoda.ZoneId
```
At least in the IDE, the only red place in this code is `Expected class 'JodaTimeZoneId' has no actual declaration in module Kotlin-DateTime-library.kotlinx-datetime.wasmJsMain for Wasm`. Are there any fundamental reasons why this approach wouldn't work everywhere? Note that we don't need Joda Time's class hierarchies, only the final classes.

core/common/src/serializers/DateTimeUnitSerializers.kt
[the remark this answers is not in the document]
dkhalanskyjb: I feel we're speaking past each other.

> The bug exists at all platforms but seems it is observable only on wasm.

The way I'm reading this, if this bug *never* surfaces on other platforms, then we can't call it a bug. Kotlin/Wasm behaves differently, making code that always works correctly on other platforms buggy. This is a problem with Kotlin/Wasm and needs to be reported.

Or did you mean that you *can* cause this code to misbehave on other platforms somehow, we just never encountered it, and there actually is a bug in the library code that's fixed by `lazy`? In what circumstances can it happen?

Either way, could you elaborate on what the bug is? At which point is a `null` value read from a non-`null` property?

core/jsAndWasmShared/src/LocalDate.kt
igoriakovlev: The only way I found to do what you suggest is here:
https://github.com/Kotlin/kotlinx-datetime/blob/b3dc7699fa116292890be6971e7904cc50fbdc63/core/wasmJs/src/JsJodaPlatform.kt#L15

I cannot use `value` classes because you cannot actualize value class with non-value classes and vice versa.

core/common/src/serializers/DateTimeUnitSerializers.kt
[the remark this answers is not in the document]
igoriakovlev: I am not really sure that it is worth to discuss this. 
Here: https://github.com/Kotlin/kotlinx-datetime/blob/master/core/common/src/serializers/DateTimeUnitSerializers.kt#L192
I see implementation of internal class `AbstractPolymorphicSerializer` which is not supposed to implement by user (with overriding internal members with a suppression and using internal API. So, the hacky way of implementing serialisation already here and I do not see how I should file a bug for a such code.
At least I do not have expertise for serialisation guts.
Anyway, if we look into `AbstractPolymorphicSerializer` implementations inside `serialisation` library, there is always usages of `lazy` for this class implementors. I believe it worth to ask about it for somebody who works with serialisation implementation. As for me, I would remove these `lazy` from `KSerializer` implementations and put it into `AbstractPolymorphicSerializer` implementations to stick with `serialisation` library internals.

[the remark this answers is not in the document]
dkhalanskyjb: > I do not see how I should file a bug for a such code

Something like this: "There's a bug in the Kotlin compiler for Wasm. Here's some code that throws a `NullPointerException` on Wasm but not on any other platform." If you can think of a minimal example (that ideally fits in a single file) where this happens without any library dependencies, it would also help the people writing the compiler. If not, it's usually enough to send them a link to any project where this reproduces, like "https://github.com/Kotlin/kotlinx-datetime/blob/e41ad76db5e52d2d2e8997f82a5704b21bdba9b7/core/common/src/serializers/DateTimeUnitSerializers.kt here, removing `lazy` on line 162 causes crashes"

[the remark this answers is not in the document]
igoriakovlev: I do not think this is Kotlin/Wasm compiler bug.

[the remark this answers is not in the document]
dkhalanskyjb: Ok, where is this difference coming from then? Is the serialization library using `expect`/`actual` in this part to have different code on different platforms and uses a different implementation from everything else on Kotlin/Wasm?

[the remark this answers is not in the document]
igoriakovlev: The fact that at all backends except wasm the behaviour is different does not mean that this is a kotlin/wasm bug, especially for the code that uses frontend suppressors.

core/jsAndWasmShared/src/LocalDate.kt
dkhalanskyjb: I double-checked, and it looks like it's true, you can't do `actual value class` for an `expect class`. Interesting.

Yes, I think it's the better separation of responsibilities, thank you for researching this! @ilya-g, do you have anything against this approach?

igoriakovlev: If @ilya-g agree that it is better then I will complete this refactoring next week. Thanks!

ilya-g: This refactoring would complicate code and prolong the review. I would rather go with the current `jsTry` in shared code approach and postpone this refactoring.

ilya-g:
master is updated to Kotlin 1.9.21, please rebase on it.

core/build.gradle.kts
```diff
@@ -386,3 +413,18 @@ tasks.withType<AbstractDokkaLeafTask>().configureEach {
         }
     }
 }
+
+// Disable intermediate sourceSet compilation because we do not need js-wasmJs artifact
+tasks.configureEach {
+    if (name == "compileJsAndWasmSharedMainKotlinMetadata") {
```
ilya-g, line 419: While it disables the compilation, the information about the corresponding shared source set is still present in kotlin-project-structure-metadata.json embedded into common klib jar.
So it needs to be checked what effects it would have in IDE in projects having similar shared source sets.

core/jsAndWasmShared/src/LocalDate.kt
dkhalanskyjb: > would complicate code and prolong the review

Funnily enough, I disagree with both points. Still, we have "one subjective opinion + effort to refactor" on one side and "one subjective opinion" on the other, so since we are split on this, not doing the refactoring wins in any case. Let's drop this line of inquiry.

core/wasmJs/src/JSJodaExceptions.kt
```diff
@@ -0,0 +1,43 @@
...
+private fun withCaughtJsException(body: () -> Unit): JsAny? = js("""{
+    try {
+        body();
+        return null;
+    } catch(e) {
+        return e;        
+    }
+}""")
+
+internal class JsException(val jsException: JsAny): Throwable()
```
dkhalanskyjb, line 30: This exception needs more care.
* It will be the `cause` of most of the exceptions we throw.
* If JS throws anything unusual, something we don't process, the user won't have any idea at all what happened.

For example, when I attempt to construct `LocalDateTime(2023, 2, 30)`, I get `IllegalArgumentException: JsException` on wasm.

At the very least, the exception message should be preserved. If it's easy to also preserve the stacktrace, it would be nice, but it's not critical and I expect it not to be simple.

dkhalanskyjb (changes requested):

build.gradle.kts
```diff
@@ -22,9 +22,17 @@ allprojects {
     repositories {
         addTrainRepositories(project)
         mavenCentral()
+        maven(url = "https://maven.pkg.jetbrains.space/kotlin/p/wasm/experimental")
     }
     tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
         // outputs the compiler version to logs so we can check whether the train configuration applied
         kotlinOptions.freeCompilerArgs += "-version"
+        kotlinOptions.freeCompilerArgs += "-Xexpect-actual-classes"
```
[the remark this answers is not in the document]
ilya-g: Let's wait for that opt-in merged to master and rebase on that.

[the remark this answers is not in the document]
ilya-g: > This line doesn't do anything, there are still warnings all over the place like

I suppose because it's applied to JVM compile tasks only.

core/wasmJs/src/JSJodaExceptions.kt
igoriakovlev: Now, in WasmJs, we got `JsException` which is caused from joda and not processed by DateTime. The message now is preserved from joda exception. The callstack a bit mess and there is no good way for now to put original call stack from joda side into `JsException` (we can make an extra method for `JsException` that receive the original stack trace from `jsException` field)

core/jsAndWasmShared/src/LocalDateTime.kt
```diff
@@ -25,34 +27,36 @@ public actual class LocalDateTime internal constructor(internal val value: jtLoc
...
+    public actual val second: Int get() = value.second()
     public actual val nanosecond: Int get() = value.nano().toInt()
 
     public actual val date: LocalDate get() = LocalDate(value.toLocalDate()) // cache?
 
     public actual val time: LocalTime get() = LocalTime(value.toLocalTime())
 
     override fun equals(other: Any?): Boolean =
-            (this === other) || (other is LocalDateTime && this.value == other.value)
+        (this === other) || (other is LocalDateTime && (this.value === other.value || this.value.equals(other.value)))
```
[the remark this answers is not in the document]
igoriakovlev: There was some interoperability problem for equality of js objects in wasm. I put all special this code to the wrapper.

core/jsAndWasmShared/src/JSJodaExceptions.kt
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright 2019-2023 JetBrains s.r.o. and contributors.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.datetime
+
+internal expect fun Throwable.isJodaArithmeticException(): Boolean
+internal expect fun Throwable.isJodaDateTimeException(): Boolean
+internal expect fun Throwable.isJodaDateTimeParseException(): Boolean
```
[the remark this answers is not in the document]
igoriakovlev: I put these changes in.

build.gradle.kts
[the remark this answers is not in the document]
igoriakovlev: I removed this flag.

gradle.properties
igoriakovlev: I replace this dependency with 1.6.1 (no wasm) but there need to be new version for kotlin 1.9.1 as far as I understand.

core/jsAndWasmShared/src/LocalDate.kt
dkhalanskyjb: Some miscommunication probably happened: I agreed not to have this refactoring. Sorry if I worded this incorrectly!

igoriakovlev: Ohh, I missed this post. :(
So, need I revert this?

ilya-g: Yes, please do, if you want this review to be completed sooner than later

dkhalanskyjb: @ilya-g, why do you believe that this approach would take longer to review? With this separation, we barely have to look at the shared code; the changes there are trivial. We only have to look through a small list of definitions and check that they are wrapped.

What *does* make this PR difficult to review now is the amount of code that's touched but not meaningfully changed. This needs to be fixed. After that, the changes would be self-contained.

igoriakovlev: I reduced unnecessary touched code. If you still want me to remove wrappers then I will.

gradle.properties
ilya-g: @sandwwraith Could you please release kotlinx-serialization with wasm-js and 1.9.21 to maven central?

core/build.gradle.kts
```diff
@@ -193,20 +202,43 @@ kotlin {
...
             dependencies {
                 implementation(npm("@js-joda/timezone", "2.3.0"))
             }
         }
 
+        val jsMain by getting {
+            dependsOn(commonJsMain)
+            dependencies {
+                api("org.jetbrains.kotlin:kotlin-stdlib-js")
+            }
```
ilya-g, line 224: This dependency should not be needed, after rebase to master when #319 is merged

core/build.gradle.kts
```diff
@@ -193,20 +202,43 @@ kotlin {
...
+
+        val jsTest by getting {
+            dependsOn(commonJsTest)
+        }
+
+        val wasmJsMain by getting {
+            dependsOn(commonJsMain)
+            dependencies {
+                api("org.jetbrains.kotlin:kotlin-stdlib-wasm-js")
+            }
```
ilya-g, line 235: Same here: not needed

gradle.properties
sandwwraith: Do you need 1.9.21 specifically? 1.6.1 already has wasm-js and is based on 1.9.20

igoriakovlev: If 1.6.1 has linuxArm32Hfp and have not to be updated in datetime then I am perfectly fine with 1.6.1. @ilya-g

core/common/src/serializers/DateTimeUnitSerializers.kt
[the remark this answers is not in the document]
igoriakovlev: I found the difference of implementation for associated objects in the `wasm` that triggered the problem.
It was a difference of associated objects initialization that was not specified and tested.
In a matter of same behavior for different targets I have filed a [bug](//https://youtrack.jetbrains.com/issue/KT-63939) for kotlin/wasm stdlib.
As far, to workaround this difference, I put `lazy` for all serialiser constructor initialisers in `DateTime` with a comment.

core/build.gradle.kts
igoriakovlev: I used a such approach for all libraries and did't find any problems with it so far. The only problem that I found so far on the K1 IDE now reports error that it is no actual for JS and Wasm targets (in a same time it found these actuals in the shared sourceSet).

core/common/src/serializers/DateTimeUnitSerializers.kt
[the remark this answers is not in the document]
dkhalanskyjb: Good job figuring this one out!

core/build.gradle.kts
igoriakovlev: Removed

core/build.gradle.kts
igoriakovlev: Removed

core/wasmJs/src/JSJodaExceptions.kt
```diff
@@ -0,0 +1,43 @@
...
+    val exception = withCaughtJsException {
+        result = body()
+    }
+
+    if (exception != null) {
+        throw JsException(exception)
+    } else {
+        return result as T
+    }
+}
```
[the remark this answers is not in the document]
igoriakovlev: Then need to move `withCaughtJsException` or make it internal. For me looks better to have all exception related code in the one place.

core/jsAndWasmShared/src/LocalDateTime.kt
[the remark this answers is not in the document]
dkhalanskyjb: The new version doesn't actually check `===` of JS objects, only of the wrappers, doesn't it? I don't notice any issues, though. Maybe it's alright not to add `===`? What kind of interoperability problem was that?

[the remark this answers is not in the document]
igoriakovlev: > maybe it's alright not to add ===
Now it is `this === other || (other is XXX && value.equals(other.value)` in wrappers (almost every wrapper which has a equals in external declarations). You suggest to change it?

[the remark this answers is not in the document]
dkhalanskyjb: I see that the original code is untouched; I just don't understand what purpose `this === other` has. If it's for JS interoperability, shouldn't it be `this.value === other.value` instead, since `value` is a JS object but `this` and `other` are not?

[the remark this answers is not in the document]
igoriakovlev: But `this === other` is also has good decision because you dont need then casting for the `other`. We can do both.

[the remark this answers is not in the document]
dkhalanskyjb: But all of this is purely as performance optimizations, right? Would simply doing `value.equals(other.value)` still work reliably?

dkhalanskyjb (approved):

gradle.properties
igoriakovlev: Super! Thanks!

ilya-g (approved):

PR #316, opened 2023-11-16
Enable binary compatibility validation

FenstonSingel:
The change is related to building datetime as a Kotlin user project.
See [KT-61621](https://youtrack.jetbrains.com/issue/KT-61621) for more details.


ilya-g:
I doubt this is necessary now.

This library is not at that stage when it provides binary compatibility guarantees. Enabling BCV check now will introduce burden on developers.

We'd rather enable it after major API changes are completed.

FenstonSingel:
Fair point. We'll keep BCV checks in the K2 user project branch at the moment then.

PR #319, opened 2023-11-28
Fix build warnings

ilya-g:


core/jvm/test/ConvertersTest.kt
```diff
@@ -5,6 +5,7 @@
 package kotlinx.datetime.test
 
 import kotlinx.datetime.*
+import java.time.Month
```
ilya-g, line 8: accidental import

dkhalanskyjb (commented):
Good idea to concentrate all the cleanup effort, thanks!

Two extra suggestions:
* Let's add a project-wide opt-in `@kotlinx.cinterop.ExperimentalForeignApi`. Right now, I have to add it to all files where any interaction with the platform happens, and I don't understand why it's worth doing explicitly.
* WDYT about enabling "warnings as errors?" This way, we'll have to do such comprehensive cleanups along the Kotlin version update PR, without the code rot biting us during normal development.

ilya-g:
> Let's add a project-wide opt-in @kotlinx.cinterop.ExperimentalForeignApi

This has to be done surgically accurate only in native source sets otherwise we'll get a warning about unresolved opt-in annotation in non-native ones. Previously this was easy when languageSettings were inherited, but now it has to be done in each source set. I've opened [KT-63930](https://youtrack.jetbrains.com/issue/KT-63930/Convenient-way-to-apply-opt-ins-and-specify-compiler-arguments-in-source-set-hierarchies) with a feature request to make it easier again.

Probably it would be easier at the moment to use file-level opt-ins.

> warnings as errors
 
 There are still a couple of warnings in Parser.kt, I didn't touch them because I suppose this code will go soon.

PR #321, opened 2023-11-30
Fix an uncaught exception on JS and Wasm

dkhalanskyjb:


dkhalanskyjb:
@igoriakovlev, well done putting `jsTry` in all the right places! After a thorough review, I found only a single place where we didn't properly place a `try`, so as far as I'm concerned, after adding this fix and dealing with the ARM Linux situation, your changes can be merged.

PR #327, opened 2023-12-01
Implementation of timezone databases for Darwin in Kotlin

dkhalanskyjb:
This **almost** works. The only issue is that, in the iOS Simulator only, not on actual devices, `NSTimeZone` and `kotlinx.datetime.TimeZone` use different timezone databases with this implementation, so there are some differences between them. During my testing, there was one extra timezone that our library recognized but `NSTimeZone` didn't, breaking `TimeZone.toNSTimeZone()` occasionally.

https://github.com/HowardHinnant/date/pull/577 suggests a solution. I tried to implement it, and I think I'm accessing the correct paths, but for some reason, this still doesn't work.

Maybe I'll have to dig into what timezone database iOS Simulator actually uses, or maybe I made a mistake somewhere and will just need to fix our code.

EDIT: turns out, the problem goes beyond different timezone databases, so we'll have to live with it.


dkhalanskyjb:
@ilya-g, I've rewritten the history to separate the concerns better.

PR #330, opened 2023-12-13
Ensure better ISO 8601 compliance

dkhalanskyjb:
ISO 8601 defines the period of one month as a fixed number of calendar days, specified by the calendar.

Before this change, we treated a month as a separate time scale unit during `DatePeriod` arithmetics, namely, we resolved the `LocalDateTime` on which we did arithmetics internally after adding the months but before adding the days.

This change makes it so that adding months and days is done as a single change, only consulting the calendar, and the time zone is only used once all calendar-based operations are performed.

The code that calculates the `DateTimePeriod` between two `Instant` values is changed symmetrically.

### DIFF
--- a/core/common/src/Instant.kt
+++ b/core/common/src/Instant.kt
@@ -8,6 +8,7 @@ package kotlinx.datetime
 import kotlinx.datetime.internal.*
 import kotlinx.datetime.serializers.InstantIso8601Serializer
 import kotlinx.serialization.Serializable
+import kotlin.math.*
 import kotlin.time.*
 
 /**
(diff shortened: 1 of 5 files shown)


core/common/src/Instant.kt
```diff
@@ -255,7 +266,23 @@ public fun Instant.minus(period: DateTimePeriod, timeZone: TimeZone): Instant =
...
+    var thisDate = thisLdt.date
+    val otherDate = otherLdt.date
+    val months = thisDate.until(otherDate, DateTimeUnit.MONTH) // `until` on dates never fails
+    thisDate = thisDate.plus(months, DateTimeUnit.MONTH) // won't throw: thisLdt + months <= otherLdt, which is known to be valid
+    val days = thisDate.until(otherDate, DateTimeUnit.DAY) // `until` on dates never fails
+    if (days.absoluteValue > 31) {
+        throw DateTimeArithmeticException("The number of months between $this and $other does not fit in an Int")
+    }
+    thisDate = thisDate.plus(days, DateTimeUnit.DAY) // won't throw: thisLdt + days <= otherLdt
+    val thisInstant = LocalDateTime(thisDate, thisLdt.time).toInstant(timeZone)
```
ilya-g, line 281: Does this `toInstant` use the same resolve mode as `ZDT.plus(days)` that was used before?

core/jvm/src/Instant.kt
```diff
@@ -11,6 +11,7 @@ import kotlinx.datetime.internal.*
 import kotlinx.datetime.serializers.InstantIso8601Serializer
 import kotlinx.serialization.Serializable
 import java.time.DateTimeException
+import java.time.ZonedDateTime
```
ilya-g, line 14: Accidental import?

core/common/src/Instant.kt
dkhalanskyjb: Not exactly.

Imagine a situation:
* UTC offset before is `+1:00`;
* An overlap happens;
* UTC offset after is `+0:00`.

If we subtract, say, a day and land into this overlap, then, as is currently implemented, the `+0:00` offset will be used. In this PR, the UTC offset `+1:00` will be used instead.

ilya-g: I believe it would be better to preserve resolve behavior in these operations, so that
 `Instant + DatePeriod(days = 1) == Instant.plus(1, DAY)`

PR #331, opened 2023-12-15
Remove `ZonedDateTime` from the Native implementation

dkhalanskyjb:
Also, fix #330 in the Native code while preserving the specific behavior of `Instant.minus` when landing in a time overlap.


dkhalanskyjb:
Closed in favor of #489.

PR #335, opened 2024-01-18
Tweaks and fixes for the locale-invariant datetime formatting

dkhalanskyjb:


dkhalanskyjb:
So, we're blocked on https://youtrack.jetbrains.com/issue/KT-65484/No-X-found-for-symbol-Y-during-compilation. No idea how to fix or work around this. Maybe we'll just have to postpone the merging of `parse` overloads into a single function.

core/jvm/test/InstantParsing.kt
```diff
@@ -0,0 +1,37 @@
...
+        fun formatTwoDigits(i: Int) = if (i < 10) "0$i" else "$i"
+        for (hour in 23..25) {
+            for (minuteRange in listOf(0..5, 58..62)) {
+                for (minute in minuteRange) {
+                    for (secondRange in listOf(0..5, 58..62)) {
+                        for (second in secondRange) {
+                            val input = "2020-03-16T${hour}:${formatTwoDigits(minute)}:${formatTwoDigits(second)}Z"
+                            assertEquals(
+                                runCatching { java.time.Instant.parse(input) }.getOrNull(),
+                                kotlin.runCatching { parseInstantLikeJavaDoes(input).toJavaInstant() }.getOrNull()
```
ilya-g, line 28: Strange that here we need to use runCatching qualified.

core/jvm/test/InstantParsing.kt
```diff
@@ -0,0 +1,37 @@
...
+                        setDate(toLocalDate().plus(1, DateTimeUnit.DAY))
+                        hour = 0
+                    }
+                    hour == 23 && minute == 59 && second == 60 -> second = 59
+                }
+            }.toInstantUsingOffset()
+        fun formatTwoDigits(i: Int) = if (i < 10) "0$i" else "$i"
+        for (hour in 23..25) {
+            for (minuteRange in listOf(0..5, 58..62)) {
+                for (minute in minuteRange) {
```
ilya-g, line 22: Minor: `for (minute in listOf(0..5, 58..62).flatten())`

core/common/test/UtcOffsetTest.kt
```diff
@@ -26,6 +26,11 @@ class UtcOffsetTest {
             "-01_00", "-01;00", "-01@00", "-01:AA",
             "@01:00")
 
+        val invalidStandaloneOffsets = listOf(
```
ilya-g, line 29: What does "standalone" means here?

ilya-g (commented):
Could you squash the commits related to overloads of parse?

core/common/test/UtcOffsetTest.kt
dkhalanskyjb: Offsets that are invalid on their own but can be used in timezone identifiers.

core/jvm/test/InstantParsing.kt
dkhalanskyjb: We don't, IDEA did this for some reason during a refactoring.

core/common/test/InstantTest.kt
```diff
@@ -128,7 +128,16 @@ class InstantTest {
...
+        val offsetFormat = UtcOffset.Format {
+            optional("Z") {
+                offsetHours()
+                optional {
+                    char(':'); offsetMinutesOfHour()
+                    optional { char(':'); offsetSecondsOfMinute() }
+                }
+            }
+        }
+        val offsets = offsetStrings.map { UtcOffset.parse(it, offsetFormat) }
```
ilya-g, line 140: Kinda strange that `DateTimeComponents.Formats.ISO_DATE_TIME_OFFSET` cannot be represented as a combination of `ISO_DATE_TIME` and `ISO_OFFSET`

core/common/test/UtcOffsetTest.kt
ilya-g: Ok, not quite clear from the name, I'd rather merge with `fixedOffsetTimeZoneIds`

ilya-g (approved):

PR #336, opened 2024-01-19
Add a Proguard rule for the missing serialization classes

### LINKED ISSUE #297: Missing classes detected while running R8 / Proguard
Attempting to obfuscate and shrink any android app that uses `Instant` results in the following errors

```
Missing classes detected while running R8. 
Please add the missing classes or apply additional keep rules that are generated in
.../app/build/outputs/mapping/release/missing_rules.txt.
```

The file mentioned contains:
```
-dontwarn kotlinx.serialization.KSerializer
-dontwarn kotlinx.serialization.Serializable
```

**To Reproduce**

1. Create a new Android app.
2. Add the following dependency: implementation("org.jetbrains.kotlinx:kotlinx-datetime:0.4.0").
3. Create a new class which includes a field that is of type "Instant" and use it somewhere.
4. Attempt to obfuscate and shrink the app in release.

**Minimal sample**
See this github project: https://github.com/MarcelReiter/dateTimeTest
You NEED to build it in release mode.
Also i made nice small commits so it's understandable what i'm doing

**Environment**
- Kotlin version: 1.8.10
- Library version: 0.4.0
- Kotlin platforms: Android JVM

A nearly identical issue has been created by someone else in the `kotlinx.serialization` github:
https://github.com/Kotlin/kotlinx.serialization/issues/2392 - but it seems like it is `kotlinx.datetime` related

dkhalanskyjb:
This permits using kotlinx-datetime without kotlinx-serialization on Android without complaints from the build tooling.

Fixes #297

### DIFF
--- a/core/jvm/resources/META-INF/proguard/datetime.pro
+++ b/core/jvm/resources/META-INF/proguard/datetime.pro
@@ -1,0 +1,11 @@
+# We depend on kotlinx-serialization as compileOnly.
+# If the serializers don't get used, the library works properly even without the
+# kotlinx-serialization dependency, but Proguard emits warnings about datetime
+# classes mentioning some serialization-related entities.
+# These rules should not cause problems: if a project actually relies on
+# serialization, then much more than just these two classes will be required,
+# so telling Proguard not to worry if these two are missing will not prevent it
+# from emitting errors for code that does use serialization but somehow forgot
+# to depend on it.
+-dontwarn kotlinx.serialization.KSerializer
+-dontwarn kotlinx.serialization.Serializable


dkhalanskyjb:
The reproducer for this issue no longer fails with an error if I publish the library locally with these changes and add a dependency on it.

ilya-g (approved):

LouisCAD:
Hello, thank you for the fix!

Can we expect a release with these changes soon?

dkhalanskyjb:
@LouisCAD, yes, a couple more things, and we'll be able to publish a release.

PR #343, opened 2024-02-20
Locale-invariant parsing and formatting

dkhalanskyjb:
The same thing as #251 (except for small tidying up here and there), with the commit history rewritten to be more manageable. @ilya-g, does this commit history look ok?


core/common/src/format/DateTimeFormat.kt
```diff
@@ -0,0 +1,144 @@
...
+     */
+    ZERO,
+
+    /**
+     * Pad with spaces.
+     */
+    SPACE
+}
+
+internal fun Padding.toKotlinCode(): String = when (this) {
```
ilya-g, line 74: May be worth to move to `toKotlinCode.kt`

ilya-g (approved):
Ok, all is clear

core/common/src/format/DateTimeFormat.kt
dkhalanskyjb: I tried to stick to the principle that the code expected to change should be kept together. For example, `DayOfWeekNames.toKotlinCode` is directly below `DayOfWeekNames`, the field specifications are right next to the directives involving the field, etc. WDYT?

ilya-g: ok

Amejonah1200:
G.O.A.T. 🐐

PR #344, opened 2024-02-26
Implement kotlinx-datetime for the Android native targets

### LINKED ISSUE #311: Support Native Tier 3 targets
As said in the official recommendations, library authors are encouraged to provide builds for tier 3 targets.
Currently, they're commented out:
```
// the following targets are not supported by kotlinx.serialization:
        /*
        target("androidNativeArm32")
        target("androidNativeArm64")
        target("androidNativeX86")
        target("androidNativeX64")
         */
```

But kotlinx.serialization actually supports them starting from 1.5.1

dkhalanskyjb:
To test, run an Android emulator, and then, in the command line, ./gradlew androidNativeArm64TestBinaries && adb push core/build/bin/androidNativeArm64/debugTest/test.kexe /data/local/tmp/ && adb shell /data/local/tmp/test.kexe

Change `Arm64` to another platfom (`X86`, `X64`, or `Arm32`) as needed.

Fixes #311


core/build.gradle.kts
```diff
@@ -47,13 +47,12 @@ kotlin {
...
                 target("linuxArm32Hfp")
             }
-            // the following targets are not supported, as we don't have timezone database implementations for them:
-            /*
-            target("androidNativeArm32")
-            target("androidNativeArm64")
-            target("androidNativeX86")
-            target("androidNativeX64")
-             */
+            common("android") {
```
ilya-g, line 50: Better "androidNative" in order not to confuse with JVM Android.

ilya-g (approved):
I wonder would it be possible to restructure native sources later, so that the source tree resembled source set tree more?
E.g.
native 
--  tzfile
---- tzdbOnFilesystem
------ darwin
------ linux
---- androidNative
-- windows

dkhalanskyjb:
So, `native/` would be
```
src/
test/
tzfile/
windows/
```
? Looks a bit strange to me. But I agree that the directory structure is too cumbersome now.

core/androidNative/src/internal/TzdbBionic.kt
```diff
@@ -0,0 +1,72 @@
...
+        fun readRules(): TimeZoneRules = readTzFile(file.copyOfRange(offset, offset + length)).toTimeZoneRules()
+    }
+}
+
+// see https://android.googlesource.com/platform/bionic/+/master/libc/tzcode/bionic.cpp for the format
+internal fun TzdbBionic(): TimeZoneDatabase = TzdbBionic(buildMap<String, TzdbBionic.Entry> {
+    for (path in listOf(
+        Path.fromString("/system/usr/share/zoneinfo/tzdata"), // immutable fallback tzdb
+        Path.fromString("/apex/com.android.tzdata/etc/tz/tzdata"), // an up-to-date tzdb, may not exist
+    )) {
```
natario1, line 28: Shouldn't the immutable fallback be last in list? @dkhalanskyjb

natario1: I see, sorry for bothering then! It just felt wrong, I'm used to see 'fallbacks' at the end of lists. Thanks for working on this 🙏

hfhbd:
Unrelated to the actual changes but there is also a new API to run android emulator tests by using Gradle: https://developer.android.com/studio/test/gradle-managed-devices
See https://github.com/hfhbd/kotlinx-uuid/blob/main/kotlinx-uuid-core/build.gradle.kts#L63 for a sample. Do you want to migrate to this API? If so, I can provide a PR.

dkhalanskyjb:
@hfhbd, it would be very nice to have tests for the Android native targets, but aren't these Android tests for the JVM version?

hfhbd:
@dkhalanskyjb Honestly, I don't know! They do work for the JVM version (of course) and you can bundle your c/native code in a `.aar` file to call it from Java too.
But the docs (and many information on the internet too) about android native are very thin compared to the Java ones, so asking someone from Google/create an issue would be helpful.

dkhalanskyjb:
@hfhbd, if you want to dive deep into this and discover a way to test native Android code on the emulator, we'll gladly accept your PR. I'd advise not spending time on this, though. If my intuitive understanding is correct, this would be a lot of work. The way Kotlin/Native builds tests is by compiling them into a separate executable, and you'd need to write some JVM wrapper to call that executable and check the test results.

natario1:
In case it helps, I built https://github.com/deepmedia/multiplatform-testing a few years ago, when trying to push for androidNative support in kotlinx repos (lack of testing was said to be one of the blockers).

It can download Android tools, create and spin up matching emulators, and run Android Native tests as a linux executable. At the time, I found a couple of bugs in atomicfu and k/n runtime thanks to it.

Issue #359, opened 2024-03-01
Failed requirement on strange time zone


kunyavskiy:
Attempt of parsing with following (minimized) format crashes with exception saying "Failed requirement"

```
    val dateTimeFormatter = DateTimeComponents.Format {
        optional {
            offset(UtcOffset.Format {
               offsetHours(Padding.NONE)
               optional { offsetMinutesOfHour(Padding.NONE) }
            })
        }
    }
 ```
 
 It's not very helpful to understand what is wrong.

dkhalanskyjb:
Yes, this exception needs to be improved. A quick answer as to why the exception is thrown: the format is ambiguous. For example, `+111` could be parsed as either `+01:11` or `+11:01`. Our system recognizes this and fails instead of trying to salvage the situation.

PR #360, opened 2024-03-04
Improve error descriptiveness in the parsing API

dkhalanskyjb:
Fixes https://github.com/Kotlin/kotlinx-datetime/issues/359
Fixes https://github.com/Kotlin/kotlinx-datetime/issues/361

### DIFF
--- a/core/common/src/internal/format/parser/ParserOperation.kt
+++ b/core/common/src/internal/format/parser/ParserOperation.kt
@@ -50,7 +50,12 @@ internal class NumberSpanParserOperation<Output>(
 
     init {
         require(consumers.all { (it.length ?: Int.MAX_VALUE) > 0 })
-        require(consumers.count { it.length == null } <= 1)
+        require(consumers.count { it.length == null } <= 1) {
+            val fieldNames = consumers.filter { it.length == null }.map { it.whatThisExpects }
+            "At most one variable-length numeric field in a row is allowed, but got several: $fieldNames. " +
+                "Parsing is undefined: for example, with variable-length month number " +
+                "and variable-length day of month, '111' can be parsed as Jan 11th or Nov 1st."
+        }
     }
 
     private val whatThisExpects: String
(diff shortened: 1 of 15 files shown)


core/common/src/internal/format/parser/ParserOperation.kt
```diff
@@ -50,7 +50,12 @@ internal class NumberSpanParserOperation<Output>(
 
     init {
         require(consumers.all { (it.length ?: Int.MAX_VALUE) > 0 })
```
qwwdfsad, line 52: This guard cannot be trigged by user-input, right?

core/common/src/internal/format/parser/ParserOperation.kt
```diff
@@ -50,7 +50,12 @@ internal class NumberSpanParserOperation<Output>(
...
-        require(consumers.count { it.length == null } <= 1)
+        require(consumers.count { it.length == null } <= 1) {
+            val fieldNames = consumers.filter { it.length == null }.map { it.whatThisExpects }
+            "At most one variable-length numeric field in a row is allowed, but got several: $fieldNames. " +
```
qwwdfsad, line 55: WDYT about adding a recommendation how to avoid the error here?
E.g. `Consider adding a `char(' ') separator between ...`

qwwdfsad (approved):

core/common/src/internal/format/parser/ParserOperation.kt
dkhalanskyjb: Yes, users can't form their own number consumers by hand at the time.

core/common/src/internal/format/parser/ParserOperation.kt
dkhalanskyjb: If you're an absolute novice and are just trying this functionality out, I think you're extremely unlikely to encounter an exception like this. You need to either explicitly unset the padding or to write a format that contains both the year and the fractional part of the second without any separators.

ilya-g (approved):

PR #362, opened 2024-03-05
Fix using `optional` between numbers

dkhalanskyjb:
See the commit message for details of the bug.

### DIFF
--- a/core/common/src/format/DateTimeFormatBuilder.kt
+++ b/core/common/src/format/DateTimeFormatBuilder.kt
@@ -129,6 +129,8 @@ public sealed interface DateTimeFormatBuilder {
          *
          * [am] is used for the AM marker (0-11 hours), [pm] is used for the PM marker (12-23 hours).
          *
+         * Empty strings can not be used as markers.
+         *
          * @see [amPmHour]
          */
         public fun amPmMarker(am: String, pm: String)
--- a/core/common/src/format/DateTimeFormatBuilder.kt
+++ b/core/common/src/format/DateTimeFormatBuilder.kt
@@ -130,6 +130,7 @@ public sealed interface DateTimeFormatBuilder {
          * [am] is used for the AM marker (0-11 hours), [pm] is used for the PM marker (12-23 hours).
          *
          * Empty strings can not be used as markers.
+         * [IllegalArgumentException] is thrown if either [am] or [pm] is empty or if they are equal.
          *
          * @see [amPmHour]
          */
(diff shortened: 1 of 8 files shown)


dkhalanskyjb:
There is still a way to make parsers fail when their specification is unambiguous, for example, with something like
```
dayOfMonth(Padding.NONE)
monthName(MonthNames("01", "02", "03", "04", "05", "06", "07", "08", "09", "10", "11", "12"))
```
as the trie doesn't get merged with the neighboring number consumers. Fixing it is completely doable, which is why we don't restrict this usage, but it would severely overcomplicate the code and would probably not help anyone at all.

core/common/src/format/DateTimeFormatBuilder.kt
```diff
@@ -129,6 +129,8 @@ public sealed interface DateTimeFormatBuilder {
          *
          * [am] is used for the AM marker (0-11 hours), [pm] is used for the PM marker (12-23 hours).
          *
+         * Empty strings can not be used as markers.
```
ilya-g, line 132: Do we check this requirement in the `amPmMarker` call, or it fails the operation (the parsing I suppose) later?
Do we need the requirement to have distinct strings here?

ilya-g (approved):

core/common/src/format/DateTimeFormatBuilder.kt
dkhalanskyjb: The `amPmMarker` call checks this requirement. Documented this and added the tests.
Also, added the requirement for distinct strings for month days, days of the week, and AM/PM.

core/common/src/format/LocalDateFormat.kt
```diff
@@ -21,6 +23,14 @@ public class MonthNames(
 ) {
     init {
         require(names.size == 12) { "Month names must contain exactly 12 elements" }
+        names.indices.forEach { ix ->
```
ilya-g, line 26: On the second thought about duplicates, this way users would not be able to use single-letter month names for formatting.

ilya-g (approved):

core/common/src/format/LocalDateFormat.kt
dkhalanskyjb: After an internal discussion, we decided that it's not an issue with the typical use cases for single-letter month names: they are mostly used independently from the other formatting, so you probably won't see things like `2024-J-06` in the wild. Instead, you can find a row of `J`, `F`, `M`, `A`, etc., as labels on a chart or buttons in a calendar app, but then, there is no reason to build a custom format, you can just do something like `date.month.name[0]`.

PR #363, opened 2024-03-05
Update Dokka to 1.9.20

whyoleg:
No issues comparing to Dokka 1.9.10

**Additional note:**
During testing I've found that some declarations are not rendered on kotlinlang.org, specifically `NSDate` - https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/to-n-s-date.html
rendered as `<Error class: unknown class>`

Actual:
image

Expected (built locally on macOS):
image

So most likely teamcity setup should be adapted to build dokka on macOS - not sure who should do it

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,7 +6,7 @@ version=0.6.0-RC.2
 versionSuffix=SNAPSHOT
 
 defaultKotlinVersion=1.9.21
-dokkaVersion=1.9.10
+dokkaVersion=1.9.20
 serializationVersion=1.6.2
 benchmarksVersion=0.7.2
 


ilya-g (approved):

ilya-g:
> So most likely teamcity setup should be adapted to build dokka on macOS - not sure who should do it

Thanks for noticing. We should open an issue in KTL for that.

dkhalanskyjb:
@whyoleg, can this be merged?

whyoleg:
Yes, sure, somehow missed the approve!

Issue #17, opened 2020-07-11
TimeZone.SYSTEM encourages static dependency


JakeWharton:
Following on from the `Instant.now()` static dependency on the system clock, `TimeZone.SYSTEM` has a similar static dependency on the system.

Would it make sense to put this as an instance member on `Clock` as well? That's how (one of) our clock abstractions exposes it.

Without this you'll have the ability to control time in places like tests. Anyone calling into `TimeZone.SYSTEM` and combining it with an instant from a fake clock will produce a time in the hosts zone rather than a stable zone likely producing flaky tests.

kevincianfarini:
Is there any interest in accepting this proposal? I'm finding people on my team using the static dependency quite a lot.

dalewking:
I am very much against the idea of adding TimeZone into the Clock interface. One of the best parts of kotlinx-datetime is the way that there is a stark separation between the world of instants and the world of local representations of those instants. Coming from Klock i can tell you that the blurring of the line leads to lots of confusion. When you start having to do things like x.local.local something is wrong.

I also understand that there is a great need for a single interface for dealing with timezones along with instants for having a way to inject the sense of current time to code to control tests. But that should not be done by polluting the Clock interface, but instead should be done by extending the Clock interface.

That is what I did in our project where I created a `DateTimeSource` interface like this:
```
interface DateTimeSource : Clock, TimeSource {
    val timeZone: TimeZone
    val localDate: LocalDate
    val localTime: LocalTime
    val localDateTime: LocalDateTime

    fun toLocalDateTime(instant: Instant): LocalDateTime
    fun fromLocalDateTime(dateTime: LocalDateTime): Instant
}
```

Any part of the code that has to deal with knowing what the current time is or converting between utc and local has an instance of this interface injected into it. I have a Standard implementation that is injected for production:
```
open class StandardDateTimeSource : DateTimeSource, Clock by Clock.System {
    override val localDateTime get() = now().toLocalDateTime(timeZone)
    override val localDate get() = todayIn(timeZone)
    override val localTime get() = localDateTime.time

    override val timeZone: TimeZone get() = TimeZone.currentSystemDefault()

    override fun markNow() = asTimeSource().markNow()

    override fun toLocalDateTime(instant: Instant) = instant.toLocalDateTime(timeZone)

    override fun fromLocalDateTime(dateTime: LocalDateTime) = dateTime.toInstant(timeZone)
}
```

[...]

PR #366, opened 2024-03-11
Wasm/WASI target implementation

igoriakovlev:
This is Wasm/WASI target implementation. It based on native version with hardcoded tzdb.

1. Extracted native implementation into intermediate sourceset.
2. Added WASI implementation
3. Added timezones-full database project

Before building timezones project you should run generateWasmWasiZoneInfo task to download and convert timezones db (nodejs and npm should be installed in the system). We decided to put the result of this task into the repository but for review it is not generated yet.


timezones/full/build.gradle.kts
```diff
@@ -0,0 +1,76 @@
...
+    @OptIn(ExperimentalWasmDsl::class)
+    wasmWasi {
+        nodejs()
+        NpmResolverPlugin.apply(project) //Workaround KT-66373
+    }
+
+    sourceSets {
+        commonMain {
+            dependencies {
+                compileOnly(project(":kotlinx-datetime"))
```
igoriakovlev, line 34: Applying timezones without datetime lead to broken compilation (or exception on usage times zone when PL is enabled). This helps us to avoid version collisions on gradle side and circullar dependency while testing.

core/wasmWasi/src/internal/TimeZonesInitializer.kt
```diff
@@ -0,0 +1,45 @@
...
+    check(timeZonesProvider != provider) { "TimeZone database redeclaration" }
+    timeZonesProvider = provider
+}
+
+@InternalDateTimeApi
+private var timeZonesProvider: TimeZonesProvider? = null
+
+@OptIn(InternalDateTimeApi::class)
+private fun zoneDataByName(name: String): ByteArray =
+    timeZonesProvider?.zoneDataByName(name) ?: error("TimeZones are not supported")
```
igoriakovlev, line 34: We need to decide how to lead user to use timezones dependency in this case

igoriakovlev:
The build fails because of none of TZ binaries in the PR.

core/wasmWasi/src/internal/TimeZonesInitializer.kt
```diff
@@ -0,0 +1,45 @@
...
+package kotlinx.datetime.internal
+
+@RequiresOptIn
+internal annotation class InternalDateTimeApi
+
+/*
+This is internal API which is not intended to use on user-side.
+ */
+@InternalDateTimeApi
+public interface TimeZonesProvider {
```
dkhalanskyjb, line 15: If you mark this as `internal` and `@Suppress` the warning, will anything break? It would be nicer not to expose unnecessary symbols to autocompletion.

core/wasmWasi/src/internal/TimeZonesInitializer.kt
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package kotlinx.datetime.internal
+
+@RequiresOptIn
+internal annotation class InternalDateTimeApi
+
+/*
+This is internal API which is not intended to use on user-side.
+ */
```
dkhalanskyjb, line 13: I think Dokka will generate the HTML for this API entry if we don't `/** @suppress */` it, no?

core/wasmWasi/src/internal/TimeZonesInitializer.kt
```diff
@@ -0,0 +1,45 @@
...
+@InternalDateTimeApi
+private var timeZonesProvider: TimeZonesProvider? = null
+
+@OptIn(InternalDateTimeApi::class)
+private fun zoneDataByName(name: String): ByteArray =
+    timeZonesProvider?.zoneDataByName(name) ?: error("TimeZones are not supported")
+
+@OptIn(InternalDateTimeApi::class)
+private fun getTimeZones(): Set<String> =
+    timeZonesProvider?.getTimeZones() ?: emptySet()
```
dkhalanskyjb, line 38: This should contain at least `UTC`.

core/wasmWasi/src/internal/TimeZonesInitializer.kt
```diff
@@ -0,0 +1,45 @@
...
+
+internal class TzdbOnData: TimeZoneDatabase {
+    override fun rulesForId(id: String): TimeZoneRules =
+        readTzFile(zoneDataByName(id)).toTimeZoneRules()
+
+    override fun availableTimeZoneIds(): Set<String> = getTimeZones()
+}
```
dkhalanskyjb, line 45: (Minor) Is there a reason to have the private functions above instead of putting the implementation here directly?

timezones/full/build.gradle.kts
```diff
@@ -0,0 +1,76 @@
...
+val tzdbDirectory = File(project.projectDir, "tzdb")
+
+kotlin {
+    @OptIn(ExperimentalWasmDsl::class)
+    wasmWasi {
+        nodejs()
+        NpmResolverPlugin.apply(project) //Workaround KT-66373
+    }
+
+    sourceSets {
```
dkhalanskyjb, line 31: Please also use the `sourceSets.all` block from the other build scripts in the library to set the directory structure like `common/test`.

timezones/full/build.gradle.kts
```diff
@@ -0,0 +1,76 @@
...
+
+val tzdbDownloadAndCompile = task<TzdbDownloadAndCompile>("tzdbDownloadAndCompile") {
+    dependsOn(timeTzdbInstall)
+    timeTzdbDirectory.set(timeTzdbInstall.outputDirectory)
+    outputDirectory.set(tzdbDirectory)
+    ianaVersion.set(tzdbVersion)
+    convertType.set(ConvertType.LARGE)
+}
+
+val generateWasmWasiZoneInfo = task("generateWasmWasiZoneInfo") {
```
dkhalanskyjb, line 62:
```suggestion
val generateWasmWasiZoneInfo by tasks.registering {
```
?

timezones/full/build.gradle.kts
```diff
@@ -0,0 +1,76 @@
...
+                runtimeOnly(project(":kotlinx-datetime"))
+                implementation(kotlin("test"))
+            }
+        }
+    }
+}
+
+val timeTzdbInstall = task<InstallTimeTzdb>("timeTzdbInstall") { }
+
+val tzdbDownloadAndCompile = task<TzdbDownloadAndCompile>("tzdbDownloadAndCompile") {
```
dkhalanskyjb, line 54:
```suggestion
val tzdbDownloadAndCompile by tasks.creating(TzdbDownloadAndCompile::class) {
```

timezones/full/build.gradle.kts
```diff
@@ -0,0 +1,76 @@
...
+        val wasmWasiTest by getting {
+            dependencies {
+                runtimeOnly(project(":kotlinx-datetime"))
+                implementation(kotlin("test"))
+            }
+        }
+    }
+}
+
+val timeTzdbInstall = task<InstallTimeTzdb>("timeTzdbInstall") { }
```
dkhalanskyjb, line 52:
```suggestion
val timeTzdbInstall by tasks.creating(InstallTimeTzdb::class) { }
```

gradle.properties
```diff
@@ -1,10 +1,13 @@
-org.gradle.jvmargs=-Xmx1G -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx3G -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
 org.gradle.java.installations.fromEnv=JDK_8
 
 group=org.jetbrains.kotlinx
 version=0.6.0-RC.2
 versionSuffix=SNAPSHOT
 
+timezonesMajorVersion=1.0
+tzdbVersion=2024a
```
dkhalanskyjb, line 9: Note (we'll have to discuss this internally): SemVer doesn't allow letters in the initial components, so `1.0.2024a` is an invalid version number. We could stick to this anyway, or we could use the less natural `1.0.2024-a` (note the dash).

timezones/full/build.gradle.kts
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+import TzdbTasks.ConvertType
+import TzdbTasks.InstallTimeTzdb
+import TzdbTasks.TzdbDownloadAndCompile
+import org.jetbrains.kotlin.gradle.targets.js.dsl.ExperimentalWasmDsl
+import org.jetbrains.kotlin.gradle.targets.js.npm.NpmResolverPlugin
+
+plugins {
+    kotlin("multiplatform")
```
dkhalanskyjb, line 13: It's a good call to have `timezonesMajorVersion` so we could describe the compatibility between `kotlinx-datetime` and the timezones package. It's easy to imagine people sticking to stable versions of the library but still upgrading the time zones.

But the Kotlin version is also part of the compatibility: there are no compatibility guarantees when someone uses the library compiled with the compiler version newer than the one used by the project itself, so to upgrade the timezone database *correctly*, you will need to upgrade Kotlin, which I think is strange.

I propose this scheme: `kotlinx-datetime-timezones-full` is compiled with the oldest compiler that works, and when some incompatibility arises between the compiler used for `kotlinx-datetime` and the one for `kotlinx-datetime-timezones-full`, only then to upgrade the compiler here, while simultaneously upgrading the `timezonesMajorVersion`.

Thoughts?

dkhalanskyjb (commented):
Also, not to forget: when we decide on the versions, artifact names, etc., we'll need to update the README.

gradle.properties
dkhalanskyjb, line 9: This may be out of this PR's scope, but it's possible to automatically detect that a new version of tzdb was published: https://github.com/ThreeTen/threetenbp/blob/main/.github/workflows/tzdbupdate.yml If you feel like implementing this, it would be nice; if not, we'll just do it later ourselves.

core/wasmWasi/src/internal/TimeZonesInitializer.kt
```diff
@@ -0,0 +1,45 @@
...
+    public fun zoneDataByName(name: String): ByteArray
+    public fun getTimeZones(): Set<String>
+}
+
+/*
+This is internal API which is not intended to use on user-side.
+ */
+@InternalDateTimeApi
+public fun initializeTimeZonesProvider(provider: TimeZonesProvider) {
+    check(timeZonesProvider != provider) { "TimeZone database redeclaration" }
```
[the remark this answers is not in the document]
igoriakovlev: I have no idea how to to resolve such conflicts for different flavors of timezones libraries without a special gradle plugin. We, with gradle plugin team, do not see the way for now, how to get one dependency with gradle resolution, the only way I see now is to allow redeclaration in runtime with a special "priority" level. But how to manage this level's needs a consideration.

core/wasmWasi/src/internal/Platform.kt
```diff
@@ -0,0 +1,44 @@
...
+        error("WASI call failed with $ret")
+    }
+}
+
+internal actual fun currentTime(): Instant = clockTimeGet().let { time ->
+    Instant(time / NANOS_PER_ONE, (time % NANOS_PER_ONE).toInt())
+}
+
+internal actual fun currentSystemDefaultZone(): Pair<String, TimeZoneRules?> =
+    throw UnsupportedOperationException("WASI platform does not support system timezone obtaining")
```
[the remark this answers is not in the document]
igoriakovlev: You would like to add it in common `TimeZone` KDoc or in make a separate KDoc for platform-specific  `currentSystemDefaultZone`?

core/wasmWasi/src/internal/TimeZonesInitializer.kt
igoriakovlev: Well, I would prefer not suppressing anything in libraries. Only with special request from libraries team.

core/wasmWasi/src/internal/TimeZonesInitializer.kt
igoriakovlev: Yes I think so. Do you know the way how to escape it?

core/wasmWasi/src/internal/TimeZonesInitializer.kt
igoriakovlev: Nope, I will inline it

core/wasmWasi/src/internal/TimeZonesInitializer.kt
igoriakovlev: It will not, because all declarations matching `.*\.internal(\..*)?` regex are suppressed.

igoriakovlev: But now dokka NOT generates `wasm` targets (both) and generates intermediate targets `commonJs` and `pureKotlin`

core/wasmWasi/src/internal/TimeZonesInitializer.kt
igoriakovlev: You would like to add it in common `TimeZone.of` KDoc or in make a separate KDoc for platform-specific `TimeZone.of`?

core/wasmWasi/src/internal/TimeZonesInitializer.kt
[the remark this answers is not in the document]
dkhalanskyjb: Here's a scheme:
```
kotlinx-datetime-timezones-full -> kotlinx-datetime-timezones-basic -> kotlinx-datetime
```
`A -> B` means "A depends on B". `kotlinx-datetime-timezones-full` only provides new data, the bulk of the data is in `kotlinx-datetime-timezones-basic`. Then, `kotlinx-datetime-timezones-basic` could have another thing like `initializeFullTimeZonesProvider`, which `kotlinx-datetime-timezones-full` would call. Do you see any problems with this approach?

core/wasmWasi/src/internal/Platform.kt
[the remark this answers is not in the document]
dkhalanskyjb: The common one, because when writing `TimeZone.currentSystemDefault()` in common code, you should still be aware that this may fail.

core/wasmWasi/src/internal/TimeZonesInitializer.kt
dkhalanskyjb: We prioritize user experience over how clean our internal implementation is, so yes, if it's possible to avoid extra public entry points in a robust manner, we try to do so.

core/wasmWasi/src/internal/TimeZonesInitializer.kt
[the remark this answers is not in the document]
igoriakovlev: Yes, for now we are not decided yet what could `-basic` be but a first thoughts was about to remove some information from timezones (like make only `future` implementation or `+-5 years`). But not amount timezones. So the information in `timezones` then will be not additive.

core/wasmWasi/src/internal/TimeZonesInitializer.kt
dkhalanskyjb: Yeah, you are right. It's strange that it generates something for `pureKotlin` and `commonJs`, but this PR has nothing to do with it. Thanks!

core/wasmWasi/src/internal/TimeZonesInitializer.kt
```diff
@@ -0,0 +1,62 @@
...
+    84,   90,  105,  102,   50,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
+    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
+    0,    0,    0,    0,    0,    0,    0,    1,    0,    0,    0,    1,    0,    0,    0,    0,
+    0,    0,    0,   84,   90,  105,  102,   50,    0,    0,    0,    0,    0,    0,    0,    0,
+    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
+    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    1,    0,    0,    0,    4,    0,
+    0,    0,    0,    0,    0,   85,   84,   67,    0,   10,   85,   84,   67,   48,   10,
+)
+
+private val utcNames get() = setOf("Universal", "Etc/Universal", "Etc/Zulu", "Etc/UCT", "Etc/UTC", "Zulu", "UCT", "UTC")
```
dkhalanskyjb, line 44: No-no, you don't need to do any of this: we already have special handling for `UTC` in the `TimeZone` initialization, it just needs to be present in the `availableTimeZones` set. Sorry for not being clear.

Though now that I look into it, you can even ignore that requirement, since on the JS, this rule also isn't enforced. We'll need to solve this in a centralized way.

igoriakovlev: Ahh, so I roll it back

timezones/full/build.gradle.kts
igoriakovlev: The only thing here is that we have no possibility (in current approach) to stick a range of major versions of `timezones` to specific range of `datetime` versions in gradle. So for now it could be only documented as compatibility table.

gradle.properties
```diff
@@ -1,10 +1,13 @@
-org.gradle.jvmargs=-Xmx1G -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx3G -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
dkhalanskyjb, line 1: The build passes fine on my machine even without this change when I do `./gradlew tzdbDownloadAndCompile clean check`. Why is this needed?

buildSrc/src/main/kotlin/zoneInfosResourcesGenerator.kt
```diff
@@ -0,0 +1,102 @@
...
+            zoneDataByNameBody.appendLine("    \"$name\" -> $tzDataName")
+            getTimeZonesBody.appendLine("    \"$name\",")
+        }
+    }
+
+    val content = buildString {
+        append(header)
+        appendLine()
+        appendLine()
+        appendLine("internal fun zoneDataByName(name: String): ByteArray = when(name) {")
```
dkhalanskyjb, line 86: Seems extremely suboptimal to compare 600 strings just to get to `CET`. Luckily, we have the trie implementation that was recently introduced to kotlinx-datetime. I've implemented some code to do this, here:
```kotlin
private class TrieNode<T>(
    val children: MutableList<Pair<String, TrieNode<T>>> = mutableListOf(),
    var terminalValue: T? = null
)

internal fun String.toKotlinCode(): String = buildString {
    append('"')
    for (c in this@toKotlinCode) {
        when (c) {
            '"' -> append("\\\"")
            '\\' -> append("\\\\")
            '\b' -> append("\\b")
            '\t' -> append("\\t")
            '\n' -> append("\\n")
            '\r' -> append("\\r")
            else -> append(c)
        }
    }
    append('"')
}

internal fun Char.toKotlinCode(): String = when (this) {
    '\'' -> "'\\''"
    else -> "'$this'"
}

fun<V> Map<String, V>.toTrieAsKotlinCode(valueToKotlinCode: (V) -> String): String {
    val trie = TrieNode<V>()
    for ((string, v) in this) {
        var node = trie
        for (char in string) {
            val searchResult = node.children.binarySearchBy(char.toString()) { it.first }
            node = if (searchResult < 0) {
                TrieNode<V>().also { node.children.add(-searchResult - 1, char.toString() to it) }
            } else {
                node.children[searchResult].second
            }
        }
        check(node.terminalValue == null)
        node.terminalValue = v
    }
    fun reduceTrie(trie: TrieNode<V>) {
        for ((_, child) in trie.children) {
            reduceTrie(child)
        }
        val newChildren = mutableListOf<Pair<String, TrieNode<V>>>()
        for ((key, child) in trie.children) {
            if (child.terminalValue == null && child.children.size == 1) {
                val (grandChildKey, grandChild) = child.children.single()
                newChildren.add(key + grandChildKey to grandChild)
            } else {
                newChildren.add(key to child)
            }
        }
        trie.children.clear()
        trie.children.addAll(newChildren.sortedBy { it.first })
    }
    reduceTrie(trie)
    val indent = "    "
    fun StringBuilder.printTrie(trie: TrieNode<V>, inputVar: String, index: Int = 0, currentIndent: String = "    ") {
        val terminalValue = trie.terminalValue
        if (terminalValue != null) {
            append("${currentIndent}if ($inputVar.length == $index) ${valueToKotlinCode(terminalValue)} else ")
        } else {
            append(currentIndent)
        }
        if (trie.children.isEmpty()) {
            appendLine("null")
        } else {
            appendLine("when {")
            for ((key, child) in trie.children) {
                val check = if (key.length == 1) {
                    "$inputVar[$index] == ${key[0].toKotlinCode()}"
                } else {
                    "$inputVar.startsWith(${key.toKotlinCode()}, startIndex = $index)"
                }
                appendLine("${currentIndent + indent}$check -> {")
                printTrie(child, inputVar, index + key.length, currentIndent + indent + indent)
                appendLine("${currentIndent + indent}}")
            }
            appendLine("${currentIndent + indent}else -> null")
            appendLine("${currentIndent}}")
        }
    }
    return buildString { printTrie(trie, "input") }
}

// test
fun main() {
    val strings = ZoneId.getAvailableZoneIds()
    strings.associateWith { it }.toTrieAsKotlinCode { "ZoneId.of(${it.toKotlinCode()})" }.also { println(it) }
}
```

Here's a sample of what this outputs:
```kotlin
    when {
        input[0] == 'A' -> {
            when {
                input.startsWith("frica/", startIndex = 1) -> {
                    when {
                        input[7] == 'A' -> {
                            when {
                                input.startsWith("bidjan", startIndex = 8) -> {
                                    if (input.length == 14) ZoneId.of("Africa/Abidjan") else null
                                }
                                input.startsWith("ccra", startIndex = 8) -> {
                                    if (input.length == 12) ZoneId.of("Africa/Accra") else null
                                }

```

dkhalanskyjb: Probably makes more sense to output the code immediately, though, and not store it as a string like I did.

buildSrc/src/main/kotlin/zoneInfosResourcesGenerator.kt
```diff
@@ -0,0 +1,102 @@
...
+        append(zoneDataByNameBody)
+        appendLine()
+        append("    else -> throw kotlinx.datetime.IllegalTimeZoneException(\"Invalid timezone name\")")
+        appendLine()
+        append("}")
+        appendLine()
+        appendLine()
+        append("internal fun getTimeZones(): Set<String> = setOf(")
```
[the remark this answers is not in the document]
igoriakovlev: Well, to set memory free. I do not think this method called a lot of times and that it expected to be fast, but with `lazy` ones it called all strings never be freed.

buildSrc/src/main/kotlin/zoneInfosResourcesGenerator.kt
dkhalanskyjb: Oh, is that so? Cool, then there isn't much point.

buildSrc/src/main/kotlin/zoneInfosResourcesGenerator.kt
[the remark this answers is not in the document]
igoriakovlev: Well, I checked it once more and see that in Wasm all loaded strings are cached in string pool. This means that once they are loaded - they are persist in memory. So yes, `lazy` is much better here.

core/wasmWasi/src/internal/TimeZonesInitializer.kt
ilya-g: > @Suppress the warning

Note that accessing internals outside of the module is an error, not a warning.

core/wasmWasi/src/internal/TimeZonesInitializer.kt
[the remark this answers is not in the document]
dkhalanskyjb: In any case, we'll be able to work around this somehow. This is a technical thing invisible to the users, so we can safely do anything we want here.

core/wasmWasi/src/internal/Platform.kt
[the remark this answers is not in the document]
dkhalanskyjb: After an internal discussion, we decided that it makes more sense to return the UTC time zone by default. The reason is, because `TimeZone.currentSystemDefault()` is not easily mockable (see https://github.com/Kotlin/kotlinx-datetime/issues/17), sometimes, there are calls to `currentSystemDefault()` deep inside the existing code, and to have them throw exceptions now would be problematic. If/when we replace `currentSystemDefault()` with a version that supports dependency injection better, that version will throw on WASI.

This should still be documented, and it should be in the common code.

gradle.properties
dkhalanskyjb: In the end, this isn't a SemVer anyway, so dashes aren't needed, `2024a` is ok.

core/wasmWasi/src/internal/TimeZonesInitializer.kt
dkhalanskyjb: Ok, in that case, why not `@Suppress` the error?

settings.gradle.kts
```diff
@@ -16,6 +16,8 @@ rootProject.name = "Kotlin-DateTime-library"
 
 include(":core")
 project(":core").name = "kotlinx-datetime"
+include(":timezones/full")
+project(":timezones/full").name = "kotlinx-datetime-timezones-full"
```
dkhalanskyjb, line 20: The artifact name we decided on is `kotlinx-datetime-zoneinfo`. At least for now, without `full`.

igoriakovlev: And if we decide to make an intermediate timezones db?

dkhalanskyjb: When we decide to do it, we'll know from the reported use cases what exactly people want, and then we'll choose the names appropriately.

core/wasmWasi/src/internal/Platform.kt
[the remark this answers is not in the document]
dkhalanskyjb: Yes.

settings.gradle.kts
dkhalanskyjb: `-zoneinfo`, not `-timezones`.

igoriakovlev:
I have rebased onto the master and merged all the AMEND commits by @dkhalanskyjb request.

dkhalanskyjb (approved):

core/wasmWasi/src/internal/TimeZonesInitializer.kt
ilya-g: This is not advisable by the compiler team and can stop working in K2

ilya-g (changes requested):

timezones/full/build.gradle.kts
```diff
@@ -0,0 +1,84 @@
...
+val tzdbDownloadAndCompile by tasks.creating(TzdbDownloadAndCompile::class) {
+    dependsOn(timeTzdbInstall)
+    timeTzdbDirectory.set(timeTzdbInstall.outputDirectory)
+    outputDirectory.set(tzdbDirectory)
+    ianaVersion.set(tzdbVersion)
+    convertType.set(ConvertType.LARGE)
+}
+
+val generateWasmWasiZoneInfo by tasks.registering {
+    inputs.dir(tzdbDirectory)
```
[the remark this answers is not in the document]
igoriakovlev: As we agreed with @dkhalanskyjb it is going to be committed.

buildSrc/src/main/kotlin/TzdbTasks/TzdbTasks.kt
```diff
@@ -0,0 +1,83 @@
...
+
+        val outputDir = project.layout.buildDirectory.zip(typePostfix) { build, type -> build.dir("tzdbCompiled$type") }
+
+        outputDirectory.convention(outputDir)
+        convertType.convention(ConvertType.LARGE)
+        executable = ""
+    }
+
+    override fun exec() {
+        executable = File(timeTzdbDirectory.get().asFile, "node_modules/.bin/tzc").path
```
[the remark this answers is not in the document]
igoriakovlev: You sure we need to support this? The result of this is going to be committed so this called only when someone wants to update tzdb.

buildSrc/src/main/kotlin/TzdbTasks/TzdbTasks.kt
```diff
@@ -0,0 +1,83 @@
...
+    @get:Optional
+    abstract val outputDirectory: DirectoryProperty
+
+    @get:Input
+    @get:Optional
+    abstract val version: Property<String>
+
+    init {
+        outputDirectory.convention(project.layout.buildDirectory.dir("time-tzdb"))
+        executable = "npm"
```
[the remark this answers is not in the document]
igoriakovlev: > Would it be possible to use npm downloaded by gradle? 
As far as I know there is no good way to download npm package with gradle.

> Or at least allow specifying a full path using a gradle property?
full path to what? to the `npm`

core/build.gradle.kts
```diff
@@ -221,14 +225,34 @@ kotlin {
             dependsOn(commonJsTest)
         }
 
-        val nativeMain by getting {
+        val pureKotlinMain by creating {
```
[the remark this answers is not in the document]
igoriakovlev: done

timezones/full/build.gradle.kts
```diff
@@ -0,0 +1,84 @@
...
+    outputs.dir(convertedKtFilesDir)
+    doLast {
+        generateZoneInfosResources(tzdbDirectory, convertedKtFilesDir, tzdbVersion)
+    }
+}
+
+tasks.getByName("compileKotlinWasmWasi") {
+    dependsOn(generateWasmWasiZoneInfo)
```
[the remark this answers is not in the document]
igoriakovlev: done

buildSrc/src/main/kotlin/TzdbTasks/TzdbTasks.kt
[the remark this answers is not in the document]
ilya-g: > As far as I know there is no good way to download npm package with gradle.

Isn't it what `node` gradle plugin does? https://github.com/node-gradle/gradle-node-plugin

buildSrc/src/main/kotlin/TzdbTasks/TzdbTasks.kt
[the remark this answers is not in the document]
igoriakovlev: Do we really need support this?

buildSrc/src/main/kotlin/TzdbTasks/TzdbTasks.kt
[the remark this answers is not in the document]
igoriakovlev: Well, with this plugin we could make something like this:
https://github.com/Kotlin/kotlinx-datetime/pull/366/commits/6832cde6562d6204225a7b9933a8caefe1ae465e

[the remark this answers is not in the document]
ilya-g: Yes, that's better. Perhaps we can replace the second task with NpxTask, though I'm not sure.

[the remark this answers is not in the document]
igoriakovlev: I tried it [here](https://github.com/Kotlin/kotlinx-datetime/commit/8c2e9b1e71eed7abe8b7d1aeabcfe71589b2cf89)
I could not change a directory for npm project (`node.nodeProjectDir` did't work) so now `node_modules` located in the project dir

timezones/full/build.gradle.kts
```diff
@@ -0,0 +1,84 @@
...
+
+val timeTzdbInstall by tasks.creating(NpmTask::class) {
+    args.addAll(
+        "install",
+        "@tubular/time-tzdb",
+    )
+}
+
+val tzdbDownloadAndCompile by tasks.creating(NpxTask::class) {
+    tzdbDirectory.mkdirs()
```
ilya-g, line 32: Please move it to the execution phase, e.g. by wrapping in `doFirst { }`

ilya-g (changes requested):

timezones/full/build.gradle.kts
```diff
@@ -0,0 +1,84 @@
...
+
+        val commonTest by getting {
+            dependencies {
+                runtimeOnly(project(":kotlinx-datetime"))
+                implementation(kotlin("test"))
+            }
+        }
+
+        val wasmWasiMain by getting {
+            languageSettings.optIn("kotlinx.datetime.internal.InternalDateTimeApi")
```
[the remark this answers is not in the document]
igoriakovlev: No, it does not work if you apply it on common code, only when you compile target module.

timezones/full/build.gradle.kts
igoriakovlev: Done

.gitignore
```diff
@@ -6,3 +6,7 @@ target
 build
 /local.properties
 benchmarks.jar
+node_modules
+package.json
+timezones.json
+package-lock.json
```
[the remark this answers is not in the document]
igoriakovlev: Done

timezones/full/build.gradle.kts
```diff
@@ -0,0 +1,84 @@
...
+    dependsOn(timeTzdbInstall)
+    command.set("@tubular/time-tzdb")
+    args.addAll("-b", "-o", "--large")
+    if (tzdbVersion.isNotEmpty()) {
+        args.addAll("-u", tzdbVersion)
+    }
+    args.add(tzdbDirectory.toString())
+}
+
+val generateWasmWasiZoneInfo by tasks.registering {
```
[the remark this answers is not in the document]
igoriakovlev: Done

buildSrc/src/main/kotlin/TzdbTasks/TzdbTasks.kt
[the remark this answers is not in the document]
igoriakovlev: Seems works indeed, done

timezones/full/build.gradle.kts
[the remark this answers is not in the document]
ilya-g: Strange, but ok. We'll move it to sourceSets.all { } if we'll be commonizing this artifact further.

ilya-g (approved):
We'll wait for 0.6.1 to be released from master before merging this

igoriakovlev:
I sqashed commits so PR is ready to merge. All tests passed. Thank you a lot for the review!

igoriakovlev:
> If this is good to go, could you rebase it on top of `master` so that we can merge it?

It is done.

dkhalanskyjb:
Great job, thanks!

PR #370, opened 2024-03-18
Fix `Instant.parse` succeeding even when seconds are omitted on the JVM and JS

### LINKED ISSUE #369: Instant.parse(..) inconsistent between jvm and ios targets
From [ISO_DATE_TIME_OFFSET format](https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime.format/-date-time-components/-formats/-i-s-o_-d-a-t-e_-t-i-m-e_-o-f-f-s-e-t.html):

> .., except that during parsing, specifying the minutes is optional.
>
> This format differs from [LocalTime.Formats.ISO](https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/-local-time/-formats/-i-s-o.html) in its time part in that specifying the seconds is not optional.

JVM allows to ommit seconds (i.e. parse "2024-03-19T17:50Z") while builds for ios do not (fails with `kotlinx.datetime.DateTimeFormatException: Parse error at char 16: expected char ':', got 'Z`)

Omitting minutes, while allowed as per documentation, always fails for both targets.

dkhalanskyjb:
Fixes #369

### DIFF
--- a/core/common/src/format/DateTimeComponents.kt
+++ b/core/common/src/format/DateTimeComponents.kt
@@ -67,7 +67,7 @@ public class DateTimeComponents internal constructor(internal val contents: Date
          * ISO 8601 extended format for dates and times with UTC offset.
          *
          * For specifying the time zone offset, the format uses the [UtcOffset.Formats.ISO] format, except that during
-         * parsing, specifying the minutes of the offset is optional.
+         * parsing, specifying the minutes of the offset is optional (so offsets like `+03` are also allowed).
          *
          * This format differs from [LocalTime.Formats.ISO] in its time part in that
          * specifying the seconds is *not* optional.
--- a/core/common/test/InstantTest.kt
+++ b/core/common/test/InstantTest.kt
@@ -86,6 +86,7 @@ class InstantTest {
 
         assertInvalidFormat { Instant.parse("1970-01-01T23:59:60Z")}
         assertInvalidFormat { Instant.parse("1970-01-01T24:00:00Z")}
+        assertInvalidFormat { Instant.parse("1970-01-01T23:59Z")}
         assertInvalidFormat { Instant.parse("x") }
         assertInvalidFormat { Instant.parse("12020-12-31T23:59:59.000000000Z") }
         // this string represents an Instant that is currently larger than Instant.MAX any of the implementations:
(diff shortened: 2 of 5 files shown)


ilya-g:
Why should we fix inconsistency between platforms in this direction and not the other way?
Could it be that users would expect string representations produced by Java Time entities be parseable with Instant.parse?

dkhalanskyjb:
After an internal discussion, we decided to keep this implementation similar to Java, which intentionally rejects `Instant` values without the seconds specified.

hrach:
@dkhalanskyjb I'd love to see the reasoning behind this :) Could you share please?

dkhalanskyjb:
Sure, here are several arguments for always requiring seconds:

* Not breaking consistency with Java without a good reason is already an argument in itself, and we didn't find a good reason.
* It should now be straightforward to create the format that allows omitting seconds, so we are not blocking any use cases.
* Symmetry with `Instant.toString()`, which always emits seconds.
* The conceptual model of `Instant` is that of a *timezone-independent* moment of time. When you're in a given time zone, even an implicit one, it makes sense to interpret `13:30` as `13:30:00.000000000`, using common conventions. But an `Instant` can be interpreted in any time zone, including those with non-whole-minute offsets from UTC (like `+01:23:45`). If you receive a `2024-01-02T03:47+01:00`, parse it as `2024-01-02T02:47:00Z`, and then interpret it in the `+01:23:45` offset, does it make sense that you have rounded the omitted seconds to zero, but now received a non-rounded wall-clock result back?

Counterarguments:

* `Instant` would be able to parse the output of `OffsetDateTime.toString()`. This should be easy to implement even if the default `parse` doesn't support it: https://github.com/Kotlin/kotlinx-datetime/issues/375#issuecomment-2031422298
* `Instant` would be able to parse the output of `LocalDateTime.toString()` if an offset is appended to it. But then, why not parse a `LocalDateTime` directly and then convert it to an `Instant`?

dkhalanskyjb:
We decided to implement `Instant.parse` in a way that a sufficiently smart DCO would be able to strip away the parsing and formatting machinery if only default `parse` and `toString` were used thoughout. @ilya-g, what do you think about me doing this in a separate PR, so as not to block the release? AFAIK, this DCO doesn't even exist yet, so no one should be negatively affected by this PR.

ilya-g:
> what do you think about me doing this in a separate PR, so as not to block the release?

ok, let's leave a TODO

ilya-g (approved):

PR #371, opened 2024-03-19
Add context to the DayOfWeek constructor precondition

qwwdfsad:


core/common/src/DayOfWeek.kt
```diff
@@ -27,6 +27,6 @@ public val DayOfWeek.isoDayNumber: Int get() = ordinal + 1
  * Returns the [DayOfWeek] instance for the given ISO-8601 week day number. Monday is 1, Sunday is 7.
  */
 public fun DayOfWeek(isoDayNumber: Int): DayOfWeek {
-    require(isoDayNumber in 1..7)
+    require(isoDayNumber in 1..7) { "Expected day number in 1..7, got $isoDayNumber" }
```
dkhalanskyjb, line 30:
```suggestion
    require(isoDayNumber in 1..7) { "Expected ISO day-of-week number in 1..7, got $isoDayNumber" }
```
Given that the argument could be used unnamed, I think it's worth emphasizing.

PR #373, opened 2024-03-22
Implement java.io.Serializable for some of the classes

### LINKED ISSUE #143: Implement java.io.Serializable on the JVM
Hello,
Please make the Android `actual` types implement also [`java.io.Serializable`](https://developer.android.com/reference/java/io/Serializable) (not Kotlin `@Serializable` as described in #37).
This will allow us to make the complex types, which depends on `kotlinx.datetime.*` types either `Serializable` or `Parcelable` in Android app.

For instance, [`java.time.Instant`](https://developer.android.com/reference/java/time/Instant) IS `Serializable`, however the  `actual` Android class [`kotlinx.datetime.Instant`](https://github.com/Kotlin/kotlinx-datetime/blob/44b8ee3caad84919f539fd1a44e124965e975877/core/jvm/src/Instant.kt#L21) is NOT.

dkhalanskyjb:
Implement java.io.Serializable for
* Instant
* LocalDate
* LocalTime
* LocalDateTime
* UtcOffset

TimeZone is not `Serializable` because its behavior is system-dependent. We can make it `java.io.Serializable` later if there is demand.

We are using string representations instead of relying on Java's entities being `java.io.Serializable` so that we have more freedom to change our implementation later.

Fixes #143


Moozart:
Can you please merge this commit if possible? @ilya-g

ilya-g (changes requested):
For better control over serialized representation I would suggest using writeReplace/readResolve with an Externalizable container.

akhbulatov:
A very necessary implementation on Android to be able to pass `LocalDate`/`LocalDateTime` objects between screens

dkhalanskyjb:
@akhbulatov, it's not "necessary", just convenient: see https://github.com/Kotlin/kotlinx-datetime/issues/143#issuecomment-975380754

wkornewald:
Do you have an estimate when this will get merged? It's a pretty huge inconvenience for us to wrap Instant because we want to export an SDK and the wrapping shouldn't be exposed to our users. It seems to be impossible without writing huge amounts of boilerplate.

core/jvm/src/Instant.kt
```diff
@@ -97,6 +99,24 @@ public actual class Instant internal constructor(internal val value: jtInstant)
 
         internal actual val MIN: Instant = Instant(jtInstant.MIN)
         internal actual val MAX: Instant = Instant(jtInstant.MAX)
+
+        private const val serialVersionUID: Long = 1L
+    }
+
+    private fun writeObject(oStream: java.io.ObjectOutputStream) {
+        oStream.defaultWriteObject()
```
ilya-g, line 107: I believe this would write all object fields first.

ilya-g (changes requested):
Take a look at how the list and set builders are serialized in stdlib. I believe we'd better have something similar here.
If you want, I can take over this branch and implement it accordingly.

dkhalanskyjb:
Do you mean https://github.com/JetBrains/kotlin/blob/3be0aa15f714726e4872c09b35f789652c46876d/libraries/stdlib/jvm/src/kotlin/collections/builders/ListBuilder.kt#L674? Ok, I think I got the idea, thanks. This does look better than reflection.

It's not important to me who does this.

core/jvm/src/internal/SerializedValue.kt
```diff
@@ -0,0 +1,74 @@
...
+ * Copyright 2019-2024 JetBrains s.r.o. and contributors.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.datetime.internal
+
+import kotlinx.datetime.*
+import java.io.*
+
+internal class SerializedValue(var typeTag: Int, var value: Any?) : Externalizable {
```
ilya-g, line 11: Alternatively, we can place it in `kotlinx.datetime` package and name it shortly, e.g. `Ser`, to shave off some bytes.

dkhalanskyjb, line 11: Can these values be `private`?

dkhalanskyjb: Do you mean, shave off some bytes from the serialized representation? I'm okay with that, as long as it still stays in the `internal/` directory.

Does it make sense to mark this with `@PublishedApi` to signal that it's an incompatible change to move/rename this class?

ilya-g: They can, but does it makes a difference for an internal class? Or you mean in order not to generate getters/setters?

ilya-g: > Does it make sense to mark this with @PublishedApi to signal that it's an incompatible change to move/rename this class?

Yes, I can do that

dkhalanskyjb:
I think it's good to go, but can't approve a pull request that I myself opened.

core/jvm/src/internal/SerializedValue.kt
dkhalanskyjb: Yep, for the latter.

dkhalanskyjb:
The reason I'm not clicking the "merge" button is that we are going to move `kotlinx.datetime.Instant` to `kotlin.time.Instant` (https://github.com/Kotlin/kotlinx-datetime/issues/382), so I don't think it makes sense to make it `java.io.Serializable`. @ilya-g, WDYT about removing `java.io.Serializable` for `Instant` from this PR and merging the implementation for the rest of the classes?

wkornewald:
Does this mean the stdlib will get a nice solution for adding Serializable support from within KMP code? I've contributed something to Ktor which could possibly serve as a starting point for a simple helper API in the stdlib: https://github.com/ktorio/ktor/pull/4421

dkhalanskyjb:
@ilya-g, FYI: only the last commit is new, the rest are only updated because I've had to rebase this PR.

dkhalanskyjb:
Now there are two new commits.

ilya-g (approved):

dkhalanskyjb:
Merged manually in 98c3e53 to preserve authorship.

core/jvm/src/LocalDate.kt
```diff
@@ -100,6 +102,8 @@ public actual class LocalDate internal constructor(internal val value: jtLocalDa
     @PublishedApi
     @JvmName("toEpochDays")
     internal fun toEpochDaysJvm(): Int = value.toEpochDay().clampToInt()
+
+    private fun writeReplace(): Any = Ser(Ser.DATE_TAG, this)
```
lukellmann, line 106: The `java.time` classes also have a [`readObject` method](https://github.com/openjdk/jdk/blob/5dd0acb3cddb96845062c0b7cee1e384e69f43cb/src/java.base/share/classes/java/time/LocalDate.java#L2176-L2185) that always throws an exception with a comment that this is about defending against malicious streams. I'm not very familiar with Java's serialization, so I don't know if this is needed.

lukellmann: I found [this post](https://nipafx.dev/java-serialization-proxy-pattern/) about serialization via `writeReplace`/`readResolve`. [This section](https://nipafx.dev/java-serialization-proxy-pattern/#artificial-byte-stream) is about the always throwing `readObject` method. Without it, a hand crafted byte stream could cause deserialization of e.g. `LocalDate` without going through `Ser.readResolve`. This could violate invariants of the classes. I think it could for example sneak in `null` to the non-nullable property `LocalDate.value`.

dkhalanskyjb: Thanks! Yes, `deserialize("aced00057372001a6b6f746c696e782e6461746574696d652e4c6f63616c44617465618443f17dae33e70200014c00097472756556616c75657400154c6a6176612f74696d652f4c6f63616c446174653b787070"
                .hexToByteArray(HexFormat { bytes.byteSeparator = "" }))` does indeed create a `LocalDate` whose `value` is `null`.

dkhalanskyjb: @lukellmann, would you like to make a PR that fixes (and tests!) this problem?

lukellmann: sure, but it will probably take a few days

lukellmann: I'm on it but it will take a bit (I'm going through the [spec](https://docs.oracle.com/en/java/javase/24/docs/specs/serialization/index.html)).

lukellmann: I've started implementing it, still refining the tests.

lukellmann: I've started implementing it, still refining the tests.

lukellmann: #522

Sorry that it took so long, I was initially trying to do more than what's in that PR (testing that classes stick to a certain pattern when implementing `java.io.Serializable`, see the other branches in [my fork](https://github.com/lukellmann/kotlinx-datetime)). But I've now shrunk it down to the most important fixes and might open additional PRs later.

PR #377, opened 2024-04-02
Upgraded Joda Core to 5.6.2; Upgraded Joda Timezone to 2.18.3

severn-everett:


dkhalanskyjb:
A user of our library may have a valid reason to stay on an old version of our dependency, and we need to have a good reason to force them to upgrade. If our users do want a fresher version of js-joda-core, they can specify it, but if they need an older version than what we require, they're in trouble.

Do you know any specific reasons to upgrade js-joda-core? Their changelog at https://github.com/js-joda/js-joda/blob/main/CHANGELOG.md and https://github.com/js-joda/js-joda/blob/%40js-joda/core%405.2.0/packages/core/CHANGELOG.md only refers to code `kotlinx-datetime` isn't using anyway, unless I've missed something.

PR #378, opened 2024-04-03
No longer make Month and DayOfWeek type aliases on the JVM

### LINKED ISSUE #96: Don't typealias Month (and others)

dkhalanskyjb:
Fixes #96

### DIFF
--- a/core/jvm/test/ConvertersTest.kt
+++ b/core/jvm/test/ConvertersTest.kt
@@ -5,7 +5,6 @@
 package kotlinx.datetime.test
 
 import kotlinx.datetime.*
-import org.junit.experimental.theories.suppliers.TestedOn
 import kotlin.random.Random
 import kotlin.test.*
 import java.time.Instant as JTInstant
@@ -38,7 +37,6 @@ class ConvertersTest {
         }
     }
 
-    @OptIn(ExperimentalStdlibApi::class)
     private fun randomDate(): LocalDate {
         val year = Random.nextInt(-20000, 20000)
         val month = Month.entries.random()
@@ -197,6 +195,7 @@ class ConvertersTest {
         fun test(month: Month) {
             val jtMonth = month.toJavaMonth()
             assertEquals(month, jtMonth.toKotlinMonth())
+            assertEquals(month.name, jtMonth.name)
         }
         Month.entries.forEach(::test)
         assertEquals(Month.JANUARY, java.time.Month.JANUARY.toKotlinMonth())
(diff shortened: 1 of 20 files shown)


core/jvm/test/ConvertersTest.kt
```diff
@@ -5,6 +5,7 @@
 package kotlinx.datetime.test
 
 import kotlinx.datetime.*
+import org.junit.experimental.theories.suppliers.TestedOn
```
ilya-g, line 8: Is this import required?

core/jvm/test/ConvertersTest.kt
```diff
@@ -187,4 +191,24 @@ class ConvertersTest {
         test("+08")
         test("-103030")
     }
+
+    @Test
+    fun month() {
+        fun test(month: Month) {
+            val jtMonth = month.toJavaMonth()
+            assertEquals(month, jtMonth.toKotlinMonth())
```
ilya-g, line 199: We can also test that the names of `month` and `jtMonth` are equal.

ilya-g (approved):
I remember we discussed this change, but cannot find the document.

dkhalanskyjb:
I actually don't remember us discussing this change. We mentioned this issue recently while discussing UUIDs and whether to make them type aliases on the JVM.

If you have doubts about this change, we can hold a meeting. What led me to go and implement this is that `expect enum`s in common code are still non-exhaustive, which I think looks extremely strange.

ilya-g:
Yes, there are some arguments for this change, like exhaustiveness of matching over expected enum entries and being able to provide companion for static extensions (though these shortcomings will be eventually fixed in Kotlin). On the other hand, this is a binary breaking change.

dkhalanskyjb:
@ilya-g, PTAL at the last commit.

ilya-g (approved):

PR #380, opened 2024-04-10
Added coverage measurement using Kover

shanshin:
When running the `check` task, the validation rule will be checked automatically: minimum coverage 85% of lines.

If on-the-fly instrumentation affects the execution of the tests, then the Kover plugin must be reconfigured.

_**JVM-only targets are supported now**_


dkhalanskyjb (approved):
Thanks!

PR #384, opened 2024-04-17
Make ISO serializers for LocalTime/LocalDateTime more predictable

### LINKED ISSUE #351: Fix the ISO serializers omitting seconds when zero and emitting fractional parts in groups of three signs
* `23:59:00.000000000` is serialized as `23:59`;
* `23:59:01.000000000` is serialized as `23:59:01`;
* `23:59:01.100000000` is serialized as `23:59:01.100`.

It looks like people value consistency over prettiness when it comes to values produced by serializers, and it's more straightforward to parse a value when all parts of the format are always there. See https://github.com/FasterXML/jackson-modules-java8/issues/76.

Other links where people are confused/irritated even by the behavior of `LocalTime.toString` in Java:
* https://stackoverflow.com/questions/49967139/java-datetimeformatter-fail-to-parse-time-string-when-seconds-and-milliseconds
* https://bugs.openjdk.org/browse/JDK-8193307
* https://stackoverflow.com/questions/62699157/zero-seconds-are-not-printed-in-spite-of-the-formatter-pattern-java-time
* https://stackoverflow.com/questions/50786482/java-8-localdatetime-dropping-00-seconds-value-when-parsing-date-string-value-wi

https://github.com/Kotlin/kotlinx-datetime/issues/333 is a similar issue.

dkhalanskyjb:
Now, they will always include seconds and will not add trailing zeros for prettiness to the fractional part.

Fixes #351

### DIFF
--- a/core/common/src/serializers/LocalDateTimeSerializers.kt
+++ b/core/common/src/serializers/LocalDateTimeSerializers.kt
@@ -27,7 +27,7 @@ public object LocalDateTimeIso8601Serializer: KSerializer<LocalDateTime> {
         LocalDateTime.parse(decoder.decodeString())
 
     override fun serialize(encoder: Encoder, value: LocalDateTime) {
-        encoder.encodeString(value.toString())
+        encoder.encodeString(LocalDateTime.Formats.ISO.format(value))
     }
 
 }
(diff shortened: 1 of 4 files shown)


ilya-g:
IMO, this does not solve consistency problem, so we should first discuss it.

dkhalanskyjb:
Thanks, this was a good call! After an internal discussion, we determined that in general, this wouldn't help *a lot* to interface with our serializers, given that fractional parts are still optional. For example, in Python, this would lead to code like this anyway: https://stackoverflow.com/questions/30584364/python-strptime-format-with-optional-bits A better way to address this concern seems to be to provide custom serializers and remove the ISO-8601 serializers altogether. We could still provide `toString()/parse()`-powered ones by some other name.

PR #387, opened 2024-04-19
Replace `dayOfMonth` with `day` and `monthNumber` with `month`

### LINKED ISSUE #84: Simplify LocalDate(Time) constructor parameter names
The parameter names are very weird for day-to-day use:

```kt
LocalDateTime(year = 2021, monthNumber = 1, dayOfMonth = 1, hour = 12)
```

I don't care that `month` is a number. That's what we have a type system for.
`OfMonth` for `day` is also redundant as that's obvious from the context.

https://github.com/Kotlin/kotlinx-datetime/blob/3af71d2e874592fc70282de441a09d024dcefb54/core/common/src/LocalDateTime.kt#L50

dkhalanskyjb:
`monthNumber` is still called that in the formatting facilities to distinguish `monthNumber` from `monthName`, and also in `DateTimeComponents` because `monthNumber` can contain out-of-bounds data and is useful even aside from being a view of `Month`.

Fixes #84


dkhalanskyjb:
With IntelliJ IDEA 2024.2.4, typing `date.dayOfMonth` and accepting the replacement suggestion from the IDE will erase `dayOfMonth` (see https://youtrack.jetbrains.com/issue/KTIJ-29647), but I didn't manage to break any realistic code with it. Whenever `dayOfMonth` is actually assigned to a variable, passed to a function, or has a method call performed on it, `dayOfMonth` will be correctly replaced with `day`. I think we can simply proceed with this the usual way, despite the IDE behavior.

core/common/src/LocalDate.kt
```diff
@@ -8,6 +8,7 @@ package kotlinx.datetime
 import kotlinx.datetime.format.*
 import kotlinx.datetime.serializers.*
 import kotlinx.serialization.Serializable
+import kotlin.internal.*
```
ilya-g, line 11: Minor: why this new import is needed (in several files)?

ilya-g (approved):

core/common/src/LocalDate.kt
dkhalanskyjb: It's for `LowPriorityInOverloadResolution`. The import can be removed by writing the FQN in the annotation.

ilya-g: Ah, ok, confused it with kotlinx.datetime.internal.

PR #390, opened 2024-04-29
Rework the Windows timezone implementation

dkhalanskyjb:
* Improved memory usage during timezone database initialization: now, buffer allocations are performed outside of loops.
* Removed the last production-facing piece of cinterop. cinterop is still used in tests.
* Added a thorough test to check that every timezone transition that we discovered is also a valid transition from the point of view of Windows.
* Used the test to determine the logic that Windows uses for time zone handling, fixing all inconsistencies but a couple of small ones:
  - In some cases, Windows claims that the transition happens on `23:59:59.999`, whereas we can express `24:00:00.000`, the value that was actually intended.
  - For a couple of time zones, for years 2030+, Windows seems to use some custom logic that doesn't seem correct anyway, so we don't reimplement it.


core/windows/src/internal/WindowsZoneNames.kt
```diff
@@ -152,33 +152,28 @@ internal val standardToWindows: Map<String, String> = mutableMapOf(
   "America/Moncton" to "Atlantic Standard Time",
   "America/Monterrey" to "Central Standard Time (Mexico)",
   "America/Montevideo" to "Montevideo Standard Time",
-  "America/Montreal" to "Eastern Standard Time",
```
ilya-g, line 155: What caused these removals? Is it no longer a valid time zone?

core/native/src/internal/TimeZoneRules.kt
```diff
@@ -128,12 +128,12 @@ internal class RecurringZoneRules(
...
         val offsetAfter: UtcOffset,
     ) {
-        override fun toString(): String = "transitioning to $offsetAfter on $transitionDateTime"
+        override fun toString(): String = "transitioning from $offsetBefore to $offsetAfter on $transitionDateTime"
     }
 
     // see `tzparse` in https://data.iana.org/time-zones/tzdb/localtime.c: looks like there's no guarantees about
     // a way to pre-sort the transitions, so we have to do it for each query separately.
-    private fun rulesForYear(year: Int): List<Rule<Instant>> {
+    fun rulesForYear(year: Int): List<Rule<Instant>> {
```
ilya-g, line 136: Is it used outside this class?

dkhalanskyjb: It's used for tests in `core/windows/test/TimeZoneRulesCompleteTest.kt`.

core/windows/src/internal/WindowsZoneNames.kt
dkhalanskyjb: `zone.tab` doesn't include `America/Montreal` since 2013: https://mm.icann.org/pipermail/tz-announce/2013-September/000013.html

dkhalanskyjb: List of zones on the MacOS filesystem but not in the Windows zone name mapping:

<details>
  <summary>Put it under a spoiler, as there's many of them</summary>

```
Africa/Asmara
Africa/Timbuktu
America/Argentina/Buenos_Aires
America/Argentina/Catamarca
America/Argentina/ComodRivadavia
America/Argentina/Cordoba
America/Argentina/Jujuy
America/Argentina/Mendoza
America/Atikokan
America/Atka
America/Ensenada
America/Fort_Wayne
America/Indiana/Indianapolis
America/Kentucky/Louisville
America/Knox_IN
America/Montreal
America/Nipigon
America/Nuuk
America/Pangnirtung
America/Porto_Acre
America/Rainy_River
America/Rosario
America/Santa_Isabel
America/Shiprock
America/Thunder_Bay
America/Virgin
America/Yellowknife
Antarctica/South_Pole
Antarctica/Troll
Asia/Ashkhabad
Asia/Chongqing
Asia/Chungking
Asia/Dacca
Asia/Harbin
Asia/Ho_Chi_Minh
Asia/Istanbul
Asia/Kashgar
Asia/Kathmandu
Asia/Kolkata
Asia/Macao
Asia/Tel_Aviv
Asia/Thimbu
Asia/Ujung_Pandang
Asia/Ulan_Bator
Asia/Yangon
Atlantic/Faroe
Atlantic/Jan_Mayen
Australia/ACT
Australia/Canberra
Australia/Currie
Australia/LHI
Australia/NSW
Australia/North
Australia/Queensland
Australia/South
Australia/Tasmania
Australia/Victoria
Australia/West
Australia/Yancowinna
Brazil/Acre
Brazil/DeNoronha
Brazil/East
Brazil/West
CET
Canada/Atlantic
Canada/Central
Canada/Eastern
Canada/Mountain
Canada/Newfoundland
Canada/Pacific
Canada/Saskatchewan
Canada/Yukon
Chile
Chile/Continental
Chile/EasterIsland
EET
Eire
Etc/GMT+0
Etc/GMT-0
Etc/GMT0
Etc/Greenwich
Etc/UCT
Etc/Universal
Etc/Zulu
Europe/Belfast
Europe/Kyiv
Europe/Nicosia
Europe/Tiraspol
Europe/Uzhgorod
Europe/Zaporozhye
GB
GB-Eire
GMT+0
GMT-0
GMT0
HST
Hongkong
Iceland
Japan
MET
Mexico/BajaNorte
Mexico/BajaSur
Mexico/General
NZ
NZ-CHAT
Navajo
PRC
Pacific/Chuuk
Pacific/Johnston
Pacific/Kanton
Pacific/Pohnpei
Pacific/Samoa
Pacific/Yap
Poland
Portugal
ROC
ROK
UCT
US/Alaska
US/Aleutian
US/Arizona
US/Central
US/East-Indiana
US/Eastern
US/Hawaii
US/Indiana-Starke
US/Michigan
US/Mountain
US/Pacific
US/Samoa
Universal
W-SU
WET
Zulu
```

</details>

core/native/src/internal/TimeZoneRules.kt
ilya-g: Ah, maybe I've not reimported the project in the branch.

core/windows/src/internal/WindowsZoneNames.kt
ilya-g: Do I get right that `America/Montreal` for example is now an alias to some other zone name, for which we have a mapping?

dkhalanskyjb: Yes, it's `America/Toronto`: https://github.com/eggert/tz/blob/7748036bace8562b9c047f368c8eba5f35e8c4b4/backward#L276

ilya-g: Ok, so if we can get these link mappings, we could allow obtaining zones by these link names from `TimeZone.of` without listing them in `availableZoneIds`. But I think this should be a separate PR.

core/windows/test/TimeZoneRulesCompleteTest.kt
```diff
@@ -0,0 +1,120 @@
...
+import kotlinx.datetime.internal.*
+import platform.windows.*
+import kotlin.test.*
+import kotlin.time.Duration.Companion.milliseconds
+
+class TimeZoneRulesCompleteTest {
+
+    /** Tests that all transitions that our system recognizes are actually there. */
+    @Test
+    fun iterateOverAllTimezones() {
```
ilya-g, line 20: I've tried to run tests locally and `iterateOverAllTimezones` has failed with
```
kotlin.AssertionError: in zone Iran Standard Time, at 2026-03-18T20:30:00Z (our guess at the offset is +04:30). Expected <2026-03-19T00:00>, actual <2026-03-19T01:00>
```

dkhalanskyjb: My guess is that your Windows isn't up-to-date, because currently, Iran doesn't seem to have DST transitions: both https://en.wikipedia.org/wiki/Iran_Standard_Time and the registry state in my Windows 10 VM think so.

This is probably the same issue as the one with `Morocco Standard Time` and `West Bank Standard Time`: there are some special cases in timezone handling that are not explained with just the registry contents. Could you add `Iran Standard Time` to the list of timezones to skip and see how it goes?

ilya-g: > Windows isn't up-to-date

Probably, but where is the discrepancy coming from? Doesn't this test compare the same information obtained with two different ways?

> Could you add Iran Standard Time to the list of timezones to skip and see how it goes?

```
in zone Syria Standard Time, at 2026-10-22T21:00:00Z (our guess at the offset is +02:00). Expected <2026-10-23T00:00>, actual <2026-10-22T23:00>.
```
If I add that one too, everything else passes successfully.

dkhalanskyjb: My guess is that Windows special-cases some time zones.

* Iran (https://en.wikipedia.org/wiki/Iran_Standard_Time) used to decide on its DST using the Solar Hijri calendar.
* Morocco (https://en.wikipedia.org/wiki/Daylight_saving_time_in_Morocco) used to decide on its DST using the Lunar Hijri calendar.
* The West Bank settlements (https://en.wikipedia.org/wiki/Time_in_the_State_of_Palestine): "IDT will begin on the Friday before the last Sunday of March".
* I don't know which rules were in Syria (https://en.wikipedia.org/wiki/Time_in_Syria) before they abolished the DST, but "sometimes" makes me think the rules were not consistent.

None of this is expressible using the Windows registry, which is why the "pre-existing" transitions are precalculated into the future.

ilya-g: ok, let's exclude them from the test

ilya-g (approved):

Issue #382, opened 2024-04-16
Consider moving `Instant` and `Clock` to the standard library


dkhalanskyjb:
Obtaining the current instant is often needed, as is doing some `Duration`-based operations on it. `Instant` is useful even when the rest of the library isn't: you don't need the timezone functionality to log human-readable timestamps (https://github.com/Kotlin/kotlinx-datetime/issues/121#issuecomment-877895270) or represent file access and modification times (okio), or sort events, etc.

With this change, the kotlinx-datetime library would provide calendar- and timezone-aware operations, whereas `Instant`, `Duration`, `Clock`, and `TimeSource` would work with just "the flow of time" as seen by the computer.

* The parsing and formatting facilities for `Instant` should probably stay in the `kotlinx-datetime` library, as they aren't very meaningful without the notion of a calendar.
* `InstantIso8601Serializer` and `InstantComponentSerializer` would probably enter the `kotlinx-serialization` library in some form.

Burtan:
Moving from 0.6.0 to 0.7.0 with migrating to kotlin.time.Instant causes errors, that there is no serializer.

PR #411, opened 2024-07-16
Implement parsing and formatting for Instant as separate functions

dkhalanskyjb:
Implement parsing and formatting of ISO strings for instant values separately from the rest of the library.

A stage of fixing #382

This is a draft PR because we probably don't need to actually merge it into the datetime library. However, the code is already written and can be reviewed.


core/native/src/Instant.kt
```diff
@@ -153,6 +154,261 @@ public actual class Instant internal constructor(public actual val epochSeconds:
...
+    fun toInstant(offsetSeconds: Int): Instant {
+        val epochSeconds = run {
+            // org.threeten.bp.LocalDate#toEpochDay
+            val epochDays = run {
+                val y = year
+                var total = 365 * y
+                if (y >= 0) {
+                    total += (y + 3) / 4 - (y + 99) / 100 + (y + 399) / 400
+                } else {
+                    total -= y / -4 - y / -100 + y / -400
```
ilya-g, line 175: Looks like a case for ceilDiv in stdlib :)

core/commonKotlin/test/InstantIsoStringsTest.kt
```diff
@@ -0,0 +1,420 @@
...
+
+    private fun parseInstant(isoString: String): Instant {
+        // return Instant.parse(isoString)
+        return parseIso(isoString)
+    }
+
+    private fun displayInstant(instant: Instant): String {
+        // return instant.toString()
+        return formatIso(instant)
+    }
```
dkhalanskyjb, line 409: @qurbonzoda, here's the logic behind this PR: the newly-added `formatIso` can replace `toString`, and the newly-added `parseIso` can replace `Instant.parse` in the case when no specific format is passed (which is the only case supported by `kotlin.time.Instant).

core/commonKotlin/src/Instant.kt
```diff
@@ -168,6 +169,266 @@ public actual class Instant internal constructor(public actual val epochSeconds:
...
+    while (i < s.length && s[i] in '0'..'9') {
+        absYear = absYear * 10 + (s[i] - '0')
+        ++i
+    }
+    val year = when {
+        i > yearStart + 10 -> {
+            parseFailure("Expected at most 10 digits for the year number, got ${i - yearStart}")
+        }
+        i == yearStart + 10 && s[yearStart] >= '2' -> {
+            parseFailure("Expected at most 9 digits for the year number or year 1000000000, got ${i - yearStart}")
```
ilya-g, line 289: Minor: can be unclear from the error message that `i - yearStart` is the actual number of digits

core/commonKotlin/src/Instant.kt
```diff
@@ -168,6 +169,266 @@ public actual class Instant internal constructor(public actual val epochSeconds:
...
+    appendTwoDigits(ldt.minute)
+    append(':')
+    appendTwoDigits(ldt.second)
+    if (ldt.nanosecond != 0) {
+        append('.')
+        var zerosToStrip = 0
+        while (ldt.nanosecond % POWERS_OF_TEN[zerosToStrip + 1] == 0) {
+            ++zerosToStrip
+        }
+        zerosToStrip -= (zerosToStrip.mod(3)) // rounding down to a multiple of 3
```
ilya-g, line 425: Minor: `%` would produce the same result for positive numbers.

core/commonKotlin/src/Instant.kt
```diff
@@ -168,6 +169,266 @@ public actual class Instant internal constructor(public actual val epochSeconds:
...
+    //                  0123456789012345 16 chars
+    if (s.length < i + 16) {
+        parseFailure("The input string is too short")
+    }
+    expect("'-'", i) { it == '-' }
+    expect("'-'", i + 3) { it == '-' }
+    expect("'T' or 't'", i + 6) { it == 'T' || it == 't' }
+    expect("':'", i + 9) { it == ':' }
+    expect("':'", i + 12) { it == ':' }
+    for (j in listOf(1, 2, 4, 5, 7, 8, 10, 11, 13, 14)) {
```
ilya-g, line 314: Better to extract such index lists to variables outside the function.

core/commonKotlin/src/Instant.kt
```diff
@@ -168,6 +169,270 @@ public actual class Instant internal constructor(public actual val epochSeconds:
...
+            throw DateTimeFormatException(
+                "The parsed date is outside the range representable by Instant (Unix epoch second $epochSeconds)"
+            )
+        return Instant.fromEpochSeconds(epochSeconds, nanosecond)
+    }
+
+    override fun toString(): String = "UnboundedLocalDateTime($year-$month-$day $hour:$minute:$second.$nanosecond)"
+
+    companion object {
+        fun fromInstant(instant: Instant, offsetSeconds: Int): UnboundedLocalDateTime {
```
qurbonzoda, line 217: There is only one call to this function, and it passes `0` for `offsetSeconds`.

core/commonKotlin/src/Instant.kt
```diff
@@ -168,6 +169,270 @@ public actual class Instant internal constructor(public actual val epochSeconds:
...
+            return UnboundedLocalDateTime(year, month, day, hours, minutes, second, instant.nanosecondsOfSecond)
+        }
+    }
+}
+
+internal fun parseIso(isoString: String): Instant {
+    fun parseFailure(error: String): Nothing {
+        throw IllegalArgumentException("$error when parsing an Instant from $isoString")
+    }
+    inline fun expect(what: String, where: Int, predicate: (Char) -> Boolean) {
```
qurbonzoda, line 265: > Local inline functions are not yet supported.

core/commonKotlin/src/Instant.kt
dkhalanskyjb: Yes. Do you see any problems with that?

PR #415, opened 2024-07-23
Rework the serializers using the DateTimeFormat API

### LINKED ISSUE #350: Introduce serializers corresponding to custom formats
Right now, we provide serializers that format the entities of our library as ISO 8601 strings, using `toString` and `parse` functions. However, now that `parse` and `format` can work with custom formats, we could provide things like `class LocalDateCustomSerializer(format: DateTimeFormat<LocalDate>): KSerializer<LocalDate>`.

### LINKED ISSUE #351: Fix the ISO serializers omitting seconds when zero and emitting fractional parts in groups of three signs

### LINKED ISSUE #416: The same serial descriptor name is used for several serializers
[SerialDescriptor](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.descriptors/-serial-descriptor/) must have a unique name, but we use one name per library entity, even when there are multiple serializers.

dkhalanskyjb:
* Default serializers are separated from the ISO serializers, which now mirror the behavior of `Formats.ISO` instead of `toString`/`parse`.
* New API: custom serializers parameterized with a `DateTimeFormat`.
* Serial names are properly distinct for incompatible formats.

Fixes #350
Fixes #351
Fixes #416

### DIFF
--- a/core/common/src/serializers/InstantSerializers.kt
+++ b/core/common/src/serializers/InstantSerializers.kt
@@ -6,6 +6,9 @@
 package kotlinx.datetime.serializers
 
 import kotlinx.datetime.Instant
+import kotlinx.datetime.format
+import kotlinx.datetime.format.DateTimeComponents
+import kotlinx.datetime.format.DateTimeFormat
 import kotlinx.serialization.*
 import kotlinx.serialization.descriptors.*
 import kotlinx.serialization.encoding.*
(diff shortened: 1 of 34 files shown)


core/common/src/serializers/InstantSerializers.kt
```diff
@@ -15,19 +17,18 @@ import kotlinx.serialization.encoding.*
...
  *
- * @see Instant.toString
- * @see Instant.parse
+ * @see DateTimeComponents.Formats.ISO_DATE_TIME_OFFSET
  */
 public object InstantIso8601Serializer : KSerializer<Instant> {
 
     override val descriptor: SerialDescriptor =
-        PrimitiveSerialDescriptor("kotlinx.datetime.Instant", PrimitiveKind.STRING)
+        PrimitiveSerialDescriptor("kotlinx.datetime.Instant ISO", PrimitiveKind.STRING)
```
ilya-g, line 25: While there are no apparent restrictions on what can be used as a serial name, and no recommendations how to name multiple alternate serializers for the same type, this name may be used in error messages without quotes, so I think it would be better not to use spaces, so it can be read as one word.
Perhaps a slash can be used? E.g. `kotlinx.datetime.Instant/ISO` and `kotlinx.datetime.Instant/components`
Or have we already come to a decision about the current serial descriptor naming scheme?

ilya-g (changes requested):

core/common/src/serializers/InstantSerializers.kt
dkhalanskyjb: I do remember that we have explicitly (but arbitrarily!) chosen spaces, yes, but the point about error messages was not raised at the time. The naming scheme you propose looks good.

core/common/src/serializers/InstantSerializers.kt
```diff
@@ -75,3 +76,70 @@ public object InstantComponentSerializer : KSerializer<Instant> {
...
+
+/**
+ * An abstract serializer for [Instant] values that uses
+ * a custom [DateTimeFormat] for serializing to and deserializing.
+ *
+ * [format] should be a format that includes enough components to unambiguously define a date, a time, and a UTC offset.
+ * See [Instant.parse] for details of how deserialization is performed.
+ *
+ * [name] is the name of the serializer.
+ * The [SerialDescriptor.serialName] of the resulting serializer is `kotlinx.datetime.Instant serializer `[name].
```
ilya-g, line 88: Please update in docs of abstract serializers as well

ilya-g (approved):

PR #419, opened 2024-08-02
Annotate Instant with @JsExport

lamba92:
Class `Instant` has been annotated with `@JsExport` to allow usage of it from JS code. Due to [**KT-62385** @JsExport doesn't play well with the WasmTarget](https://youtrack.jetbrains.com/issue/KT-62385), a workaround has been added as well.

### DIFF
--- a/core/commonJs/src/Instant.kt
+++ b/core/commonJs/src/Instant.kt
@@ -7,6 +7,10 @@
 
 package kotlinx.datetime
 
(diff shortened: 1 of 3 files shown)


lppedd:
This will affect everyone using datetime, even the ones not using `Instant`.  
Plus, it will possibly output an unwanted TypeScript declaration.

PR #438, opened 2024-09-23
Add export of kotlinx.datetime.format in module-info

cdelabou:
This allows usages of the introduced format api from java 9 modules, which currently fails with the following error:

```
Kotlin: Symbol is declared in module 'kotlinx.datetime' which does not export package 'kotlinx.datetime.format'.
```


dkhalanskyjb:
Thank you!

PR #441, opened 2024-09-27
Fix typo in hour property header doc

davertay:
Fixes a small but confusing typo for library consumers :)


dkhalanskyjb (approved):
Thank you!

PR #447, opened 2024-10-17
Fix some parsing bugs in default parsers

### LINKED ISSUE #443: A number of bugs and inconsistencies with Java
https://gist.github.com/ilma4/a5ad5b5c284969e24f4e2149a2f8b273 lists several bugs and inconsistencies.

- [ ] `kotlinx.datetime.LocalDate.parse(s) vs LocalDate.Formats.ISO.parse(s)` (and the same for `LocalDateTime`): Java's facilities don't support having too many leading zeroes, but our parsing mechanism does, and delegating to Java prevents us from reading strings like `+00000000000002022-07-16`.
- [ ] `DatePeriod.parse and DateTimePeriod.parse parsing "P"`: it is inconsistent with ISO 8601 to parse just `P` without any components, so we should probably fix that.
- [x] `kotlinx.datetime.DatePeriod is always normalized unlike `java.time.Duration`: this is intentional.
- [x] `kotlinx.datetime.Instant.parse`: intentional.
- [x] `Different periods between two dates in Java and Kotlin`: looks like Java's bug.
- [x] `kotlin.datetime.Duration max value is different from Java's`: intentional.
- [ ] `kotlin.time.Duration.parseIsoString with too many digits`: it is a bug that we think `PT0000000000000000000000000M` is an infinite duration. (Not this library's scope, though)
- [ ] `kotlin.time.Duration.parse("PT+-2H") successfully parses -2 hours`: not valid ISO 8601, we can fix that. (Not this library's scope, though)

dkhalanskyjb:
Fixes #443

### DIFF
--- a/core/common/src/internal/util.kt
+++ b/core/common/src/internal/util.kt
@@ -8,3 +8,32 @@ package kotlinx.datetime.internal
 internal fun Char.isAsciiDigit(): Boolean = this in '0'..'9'
 
 internal fun Char.asciiDigitToInt(): Int = this - '0'
+
+/** Working around the JSR-310 behavior of failing to parse long year numbers even when they start with leading zeros */
+private fun removeLeadingZerosFromLongYearForm(input: String, minStringLengthAfterYear: Int): String {
+    // the smallest string where the issue can occur is "+00000002024", its length is 12
+    val failingYearStringLength = 12
+    // happy path: the input is too short or the first character is not a sign, so the year is not in the long form
+    if (input.length < failingYearStringLength + minStringLengthAfterYear || input[0] !in "+-") return input
+    // the year is in the long form, so we need to remove the leading zeros
+    // find the `-` that separates the year from the month
+    val yearEnd = input.indexOf('-', 1).takeIf { it != -1 }
+    // if the year is too short, no need to remove the leading zeros, and if the string is malformed, just leave it
+    if (yearEnd == null || yearEnd < failingYearStringLength) return input
+    // how many leading zeroes are there?
+    var leadingZeros = 0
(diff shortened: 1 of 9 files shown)


core/common/src/internal/util.kt
```diff
@@ -8,3 +8,32 @@ package kotlinx.datetime.internal
...
+private fun removeLeadingZerosFromLongYearForm(input: String, minStringLengthAfterYear: Int): String {
+    // the smallest string where the issue can occur is "+00000002024", its length is 12
+    val failingYearStringLength = 12
+    // happy path: the input is too short or the first character is not a sign, so the year is not in the long form
+    if (input.length < failingYearStringLength + minStringLengthAfterYear || input[0] !in "+-") return input
+    // the year is in the long form, so we need to remove the leading zeros
+    // find the `-` that separates the year from the month
+    val yearEnd = input.indexOf('-', 1).takeIf { it != -1 }
+    // if the year is too short, no need to remove the leading zeros, and if the string is malformed, just leave it
+    if (yearEnd == null || yearEnd < failingYearStringLength) return input
```
ilya-g, line 22:
```suggestion
    val yearEnd = input.indexOf('-', 1)
    // if the year is too short, no need to remove the leading zeros, and if the string is malformed, just leave it
    if (yearEnd < 0 || yearEnd < failingYearStringLength) return input
```

ilya-g: Or, in this case the first condition can be removed, though it would be less clear.

ilya-g (approved):

core/common/src/internal/util.kt
dkhalanskyjb: Good call. With a clarifying comment, it seems fine.

Issue #372, opened 2024-03-21
Consider removing `Clock.asTimeSource`


dkhalanskyjb:
There are no known public usages of this function: https://grep.app/search?q=asTimeSource&words=true Also, `Clock.System.asTimeSource().measureTime { }` looks fairly benign, even though it doesn't do what you want, given that the system clock can jump in time by several minutes occasionally.

dkhalanskyjb:
What's the purpose of what? `Clock`? It's to show you the current time, according to the system. `2024-03-22T09:43:15Z`, as of writing.

hfhbd:
How is this issue linked to #382?

Also, isn't the (wall) clock based Timesource more precise than the monotonic due to ntp syncs/monotonic pauses, see https://github.com/Kotlin/kotlinx-datetime/pull/164#issuecomment-997859203?

dkhalanskyjb:
> How is this issue linked to https://github.com/Kotlin/kotlinx-datetime/issues/382?

If we decide this function is useful after all, I think that it, too, should migrate to the stdlib.

> Also, isn't the (wall) clock based Timesource more precise than the monotonic due to ntp syncs/monotonic pauses, see https://github.com/Kotlin/kotlinx-datetime/pull/164#issuecomment-997859203?

Wall clocks are neither more nor less precise than `TimeSource.MONOTONIC`, they simply serve different needs and fail in different scenarios.

* `TimeSource.MONOTONIC` measures the elapsed time from the point of view of a program. If we start now, let the program run for a couple of seconds, the time measured by `TimeSource.MONOTONIC` will be two seconds. It won't be `-2` seconds, it won't be `0`. You don't know what the wall clock will say, because someone could have adjusted it in the meantime, or because the computer didn't consider some amount of time due to being hibernated.
* `Clock.System.now()` finds out the current wall-clock time, according to the computer. We could look at the clock now, and two seconds *later* (in terms of the time that actually elapsed), the wall clock could stay the same, or even shift to an earlier time, or whatever, because someone adjusted the clock in the meantime.

dalewking:
Not a public usage of this, but our project uses this for one reason. We have a cache class that is supposed to expire. When we store a value we also store a TimeMark with it to know whether the value has expired. That time mark comes from an object that implements Clock so that we can control time in tests to tests things like the cache expiration.

I think the problem with Monotonic is that it doesn't have a good way to control time in tests. It could probably be made to do so, but I think it is not quite there.

[...]

PR #450, opened 2024-10-18
Deprecate Clock.asTimeSource

dkhalanskyjb:
Fixes https://github.com/Kotlin/kotlinx-datetime/issues/372

### DIFF
--- a/core/common/src/Clock.kt
+++ b/core/common/src/Clock.kt
@@ -77,9 +77,10 @@ public fun Clock.todayIn(timeZone: TimeZone): LocalDate =
  * because [Clock.System] is not well suited for measuring time intervals.
  * Please only use this conversion function on the [Clock] instances that are fully controlled programmatically.
  */
+@ExperimentalTime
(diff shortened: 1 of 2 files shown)


core/common/src/Clock.kt
```diff
@@ -77,8 +77,11 @@ public fun Clock.todayIn(timeZone: TimeZone): LocalDate =
  * because [Clock.System] is not well suited for measuring time intervals.
  * Please only use this conversion function on the [Clock] instances that are fully controlled programmatically.
  */
-@ExperimentalTime
+@Deprecated("This function is deprecated because Clock.System.asTimeSource " +
+        "can be confused with TimeSource.Monotonic, which are very different. " +
+        "See https://github.com/Kotlin/kotlinx-datetime/issues/372", level = DeprecationLevel.ERROR)
```
ilya-g, line 82: Better to keep ExperimentalTime and start with a deprecation warning.

ilya-g (approved):

PR #453, opened 2024-10-31
Implement the same ranges for all platforms

### LINKED ISSUE #432: Make admissible ranges the same on all platforms
Create a giant local date on a JVM server:
```kotlin
java.time.LocalDate.MAX.toKotlinLocalDate()
```
Send it over the network to a Native client. Parsing this date will fail.

Why? The reason is that `LocalDate`, `LocalDateTime`, and `Instant` have different platform-specific ranges on different platforms.

It may be surprising that some values produced, then serialized and deserialized purely by the code in our library fail to work on the same version of our library on a different platform.

The proposal: align all platform-specific ranges to be the same.

Cons:
* Slightly less efficient implementation.
* The enormous dates and instants are hard to come by: on Native, the range is generous `-1000000..1000000` years, enough for all realistic use cases. Java-specific sentinel values are the one exception.

dkhalanskyjb:
Fixes #432


core/common/src/LocalDate.kt
```diff
@@ -471,7 +473,7 @@ public expect fun LocalDate.plus(value: Int, unit: DateTimeUnit.DateBased): Loca
  * @throws DateTimeArithmeticException if the result exceeds the boundaries of [LocalDate].
  * @sample kotlinx.datetime.test.samples.LocalDateSamples.minus
  */
-public expect fun LocalDate.minus(value: Int, unit: DateTimeUnit.DateBased): LocalDate
+public fun LocalDate.minus(value: Int, unit: DateTimeUnit.DateBased): LocalDate = plus(-value.toLong(), unit)
```
lukellmann, line 476: it might be a good idea to use parenthesis here, because of this:
```kt
val value = Int.MIN_VALUE
println(  -value.toLong()    ) // 2147483648
println(  -(value.toLong())  ) // 2147483648
println(  (-value).toLong()  ) // -2147483648
```

core/common/src/LocalDate.kt
```diff
@@ -344,7 +350,7 @@ public operator fun LocalDate.minus(period: DatePeriod): LocalDate =
  * - Negative or zero if this date is later than the other.
  * - Exactly zero if this date is equal to the other.
  *
- * @throws DateTimeArithmeticException if the number of months between the two dates exceeds an Int (JVM only).
+ * @throws DateTimeArithmeticException if the number of months between the two dates exceeds an Int.
```
lukellmann, line 353: Is this still true now that `DatePeriod.totalMonths` is a `Long`?

core/common/src/LocalDate.kt
```diff
@@ -86,7 +86,15 @@ public expect class LocalDate : Comparable<LocalDate> {
         /**
          * Returns a [LocalDate] that is [epochDays] number of days from the epoch day `1970-01-01`.
          *
-         * @throws IllegalArgumentException if the result exceeds the platform-specific boundaries of [LocalDate].
+         * @throws IllegalArgumentException if the result exceeds the boundaries of [LocalDate].
```
qurbonzoda, line 89: Do I get it correctly that currently the boundaries of `LocalDate` are not documented, but it is guaranteed to support all dates within `Instant.DISTANT_PAST` and `Instant.DISTANT_FUTURE`?

What about the `Instant`?

dkhalanskyjb: The same goes for `Instant`: its boundaries are not documented, but `Instant.DISTANT_PAST` and `Instant.DISTANT_FUTURE` are guaraneteed to behave well.

The reason is that the only use cases for `Instant.MAX` and `Instant.MIN` we've encountered so far were:
* A sentinel "fake" value, in particular when looking for the maximum of the minimum value. `null` typically serves this purpose better.
* A shared concept of "a time so far away into the past/future that real data will certainly not be outside this boundary". `DISTANT_PAST` and `DISTANT_FUTURE` are ok for that.

core/common/src/LocalDate.kt
```diff
@@ -231,12 +239,10 @@ public expect class LocalDate : Comparable<LocalDate> {
     /**
      * Returns the number of days since the epoch day `1970-01-01`.
      *
-     * If the result does not fit in [Int], returns [Int.MAX_VALUE] for a positive result or [Int.MIN_VALUE] for a negative result.
-     *
      * @see LocalDate.fromEpochDays
      * @sample kotlinx.datetime.test.samples.LocalDateSamples.toEpochDays
      */
-    public fun toEpochDays(): Int
+    public fun toEpochDays(): Long
```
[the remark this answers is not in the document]
dkhalanskyjb: We didn't find a way to choose an overload based on its return type, so we can't restore source compatibility. Also, on Native, the binary compatibility can't be restored, as two signatures with different return types cause a compilation error.

Restored the binary compatibility for the JVM.

core/commonJs/src/internal/Platform.kt
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    fun List<Long>.partialSums(): List<Long> = scanWithoutInitial(0, Long::plus)
+
+    val zones = mutableMapOf<String, TimeZoneRules>()
+    val (zonesPacked, linksPacked) = readTzdb() ?: return@runCatching null
+    for (zone in zonesPacked) {
+        val components = zone.split('|')
+        val offsets = components[2].split(' ').map { unpackBase60(it) }
+        val indices = components[3].map { it - '0' }
```
qurbonzoda, line 74: I'm not sure if any time zone will ever have more than 10 different offsets, now or in the future.
But the documentation says it's in base 60, which can accommodate up to 60 values. Does it make sense to use `charCodeToInt(it)` just to be safe?

core/commonJs/src/internal/Platform.kt
```diff
@@ -0,0 +1,153 @@
...
+            transitionEpochSeconds = lengthsOfPeriodsWithOffsets.partialSums().take<Long>(indices.size - 1),
+            offsets = indices.map { UtcOffset(null, null, -(offsets[it] * 60).roundToInt()) },
+            recurringZoneRules = null
+        )
+    }
+    for (link in linksPacked) {
+        val components = link.split('|')
+        zones[components[0]]?.let { rules ->
+            zones[components[1]] = rules
+        }
```
qurbonzoda, line 89: Is the order of the two time zones specified? I mean is it possible that the second time zone has rules but the first doesn't?

core/commonJs/src/internal/Platform.kt
```diff
@@ -0,0 +1,153 @@
...
+
+    fun <T, R> List<T>.scanWithoutInitial(initial: R, operation: (acc: R, T) -> R): List<R> = buildList {
+        var accumulator = initial
+        for (element in this@scanWithoutInitial) {
+            accumulator = operation(accumulator, element)
+            add(accumulator)
+        }
+    }
+
+    fun List<Long>.partialSums(): List<Long> = scanWithoutInitial(0, Long::plus)
```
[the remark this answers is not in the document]
dkhalanskyjb: Yes, this is indeed outdated.

core/commonJs/src/internal/Platform.kt
dkhalanskyjb: Sure, if the document says so, let's do that.

core/commonJs/src/internal/Platform.kt
dkhalanskyjb: https://momentjs.com/timezone/docs/#/data-formats/link-format/ doesn't say anything either way, so the js-joda implementation is the best guidance we have on this, and they do assume this order.

core/commonJs/src/internal/Platform.kt
```diff
@@ -0,0 +1,153 @@
...
+    for (zone in zonesPacked) {
+        val components = zone.split('|')
+        val offsets = components[2].split(' ').map { unpackBase60(it) }
+        val indices = components[3].map { it - '0' }
+        val lengthsOfPeriodsWithOffsets = components[4].split(' ').map {
+            (unpackBase60(it) * SECONDS_PER_MINUTE * MILLIS_PER_ONE).roundToLong() / // minutes to milliseconds
+                    MILLIS_PER_ONE // but we only need seconds
+        }
+        zones[components[0]] = TimeZoneRules(
+            transitionEpochSeconds = lengthsOfPeriodsWithOffsets.partialSums().take<Long>(indices.size - 1),
```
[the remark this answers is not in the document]
dkhalanskyjb: I've looked into this, and there are two cases in the current tzdb:
* As you said, `lengthsOfPeriodsWithOffsets.size` is already `indices.size - 1`.
* Both `lengthsOfPeriodsWithOffsets` and `indices` are one-element lists.

core/common/src/LocalDate.kt
qurbonzoda: Thanks for the explanation!

core/commonKotlin/src/Instant.kt
```diff
@@ -200,7 +215,7 @@ public actual fun Instant.plus(value: Long, unit: DateTimeUnit, timeZone: TimeZo
         is DateTimeUnit.DateBased -> {
             if (value < Int.MIN_VALUE || value > Int.MAX_VALUE)
                 throw ArithmeticException("Can't add a Long date-based value, as it would cause an overflow")
```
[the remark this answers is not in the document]
dkhalanskyjb: Nope, it's not. Good catch, thanks!

core/commonJs/src/internal/Platform.kt
```diff
@@ -0,0 +1,153 @@
...
+
+    val zones = mutableMapOf<String, TimeZoneRules>()
+    val (zonesPacked, linksPacked) = readTzdb() ?: return@runCatching null
+    for (zone in zonesPacked) {
+        val components = zone.split('|')
+        val offsets = components[2].split(' ').map { unpackBase60(it) }
+        val indices = components[3].map { it - '0' }
+        val lengthsOfPeriodsWithOffsets = components[4].split(' ').map {
+            (unpackBase60(it) * SECONDS_PER_MINUTE * MILLIS_PER_ONE).roundToLong() / // minutes to milliseconds
+                    MILLIS_PER_ONE // but we only need seconds
```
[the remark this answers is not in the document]
dkhalanskyjb: I've removed rounding (and using Double values) altogether, so this question shouldn't be important anymore.

core/js/src/PlatformSpecifics.kt
```diff
@@ -4,13 +4,17 @@
  */
 package kotlinx.datetime.internal
 
-import kotlinx.datetime.internal.JSJoda.ZoneId
+import kotlinx.datetime.internal.JSJoda.ZoneRulesProvider
 
-internal actual fun getAvailableZoneIdsSet(): Set<String> =
-    ZoneId.getAvailableZoneIds().unsafeCast<Array<String>>().toSet()
+internal actual fun readTzdb(): Pair<List<String>, List<String>>? = try {
+    val tzdbData = ZoneRulesProvider.asDynamic().getTzdbData()
```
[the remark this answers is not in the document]
dkhalanskyjb: I did manage to find an incantation that seems like it would allow us to improve the lives of JS and Wasm/JS kotlinx-datetime users, but that's too difficult to be in the scope of this PR.

core/commonJs/src/internal/Platform.kt
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * References:
+     * - https://github.com/js-joda/js-joda/blob/8c1a7448db92ca014417346049fb64b55f7b1ac1/packages/timezone/src/MomentZoneRulesProvider.js#L78-L94
+     * - https://github.com/js-joda/js-joda/blob/8c1a7448db92ca014417346049fb64b55f7b1ac1/packages/timezone/src/unpack.js
+     * - <https://momentjs.com/timezone/docs/#/zone-object/>
+     */
+    fun unpackBase60(string: String): Double {
+        var i = 0
+        var parts = string.split('.')
+        var whole = parts[0]
```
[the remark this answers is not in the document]
dkhalanskyjb: Alright, there are no remaining `var`s in the file.

core/commonKotlin/src/LocalDate.kt
```diff
@@ -198,65 +204,57 @@ public actual class LocalDate actual constructor(public actual val year: Int, pu
...
-        when (unit) {
-            is DateTimeUnit.DayBased -> plusDays(safeMultiply(value, unit.days))
-            is DateTimeUnit.MonthBased -> plusMonths(safeMultiply(value, unit.months))
-        }
-    } catch (e: ArithmeticException) {
-        throw DateTimeArithmeticException("Arithmetic overflow when adding a value to a date", e)
-    } catch (e: IllegalArgumentException) {
-        throw DateTimeArithmeticException("Boundaries of LocalDate exceeded when adding a value", e)
-    }
+public actual fun LocalDate.plus(unit: DateTimeUnit.DateBased): LocalDate = plus(1, unit)
```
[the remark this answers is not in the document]
dkhalanskyjb: I don't think so. The JVM overload is in `@file:JvmName("LocalDateJvmKt")`, and moving it to `LocalDateKt` would be a breaking change. We could keep a `LowPriorityInOverloadResolution` `DeprecationLevel.HIDDEN` copy in `LocalDateJvmKt`, but then what's the point in moving it to common code? It's deprecated anyway.

core/js/src/PlatformSpecifics.kt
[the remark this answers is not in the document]
qurbonzoda: Thanks for the clarification!

core/commonKotlin/src/LocalDate.kt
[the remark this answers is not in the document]
qurbonzoda: Thanks for explanation!

settings.gradle.kts
```diff
@@ -20,5 +20,6 @@ include(":timezones/full")
 project(":timezones/full").name = "kotlinx-datetime-zoneinfo"
 include(":serialization")
 project(":serialization").name = "kotlinx-datetime-serialization"
+include(":js-without-timezones")
+project(":js-without-timezones").name = "kotlinx-datetime-js-without-timezones"
```
ilya-g, line 24:
```suggestion
project(":js-without-timezones").name = "kotlinx-datetime-js-test-without-timezones"
```

core/common/src/DateTimePeriod.kt
```diff
@@ -464,7 +465,7 @@ public class DatePeriod internal constructor(
 
     /** The number of nanoseconds in this period. Always equal to zero. */
     override val nanoseconds: Int get() = 0
-    internal override val totalNanoseconds: Long get() = 0
+    override val totalNanoseconds: Long get() = 0
```
ilya-g, line 468: Better to keep `internal` on overrides for clarity

core/common/src/LocalDate.kt
```diff
@@ -160,7 +168,7 @@ public expect class LocalDate : Comparable<LocalDate> {
      * The components [monthNumber] and [dayOfMonth] are 1-based.
      *
      * The supported ranges of components:
-     * - [year] the range is platform-dependent, but at least is enough to represent dates of all instants between
+     * - [year] the range is at least is enough to represent dates of all instants between
```
ilya-g, line 171: Double "is"

core/common/test/InstantTest.kt
```diff
@@ -313,7 +313,9 @@ class InstantTest {
                 val diff2 = date1.periodUntil(date2)
 
                 if (diff1 != diff2)
-                    println("start: $instant1, end: $instant2, diff by instants: $diff1, diff by dates: $diff2")
+                    throw AssertionError(
```
ilya-g, line 316:
```suggestion
                    fail(
```

core/common/test/InstantTest.kt
```diff
@@ -574,6 +559,12 @@ class InstantRangeTest {
             assertArithmeticFails("$instant") { instant.plus(Long.MAX_VALUE, DateTimeUnit.YEAR, UTC) }
             assertArithmeticFails("$instant") { instant.plus(Long.MIN_VALUE, DateTimeUnit.YEAR, UTC) }
         }
+        for (instant in smallInstants) {
+            instant.plus(2 * Int.MAX_VALUE.toLong(), DateTimeUnit.DAY, UTC)
+            instant.plus(2 * Int.MIN_VALUE.toLong(), DateTimeUnit.DAY, UTC)
+            instant.plus(2 * Int.MAX_VALUE.toLong(), DateTimeUnit.MONTH, UTC)
+            instant.plus(2 * Int.MIN_VALUE.toLong(), DateTimeUnit.MONTH, UTC)
```
ilya-g, line 566: Perhaps worth checking if subtracting the same value produces the same instant without clamping in between.

core/commonKotlin/src/Instant.kt
```diff
@@ -106,13 +119,15 @@ public actual class Instant internal constructor(public actual val epochSeconds:
...
-            else Instant(
-                epochMilliseconds.floorDiv(MILLIS_PER_ONE.toLong()),
-                (epochMilliseconds.mod(MILLIS_PER_ONE.toLong()) * NANOS_PER_MILLI).toInt()
-            )
+        public actual fun fromEpochMilliseconds(epochMilliseconds: Long): Instant {
+            val epochSeconds = epochMilliseconds.floorDiv(MILLIS_PER_ONE.toLong())
+            val nanosecondsOfSecond = (epochMilliseconds.mod(MILLIS_PER_ONE.toLong()) * NANOS_PER_MILLI).toInt()
+            return when {
+                epochSeconds < MIN_SECOND -> MIN
+                epochSeconds > MAX_SECOND -> MAX
```
ilya-g, line 127: Is it still possible? If the Instant range doesn't potentially fit in the Long range of milliseconds, doesn't it mean that the Long range of milliseconds always fit in the Instant range?

ilya-g (approved):

PR #457, opened 2024-11-13
Implement YearMonth

### LINKED ISSUE #168: Substitute for YearMonth?
java.time has a YearMonth class for referring to a particular month in the year, what's the kotlinx-datetime equivalent? LocalDate also stores a day of the month which isn't ideal

### LINKED ISSUE #184: Getting days in Month
Is there currently any way of getting the number of total days in a month?

dkhalanskyjb:
Fixes #168
Fixes #184

Related PRs:

* After https://github.com/Kotlin/kotlinx-datetime/pull/189, we can implement `YearMonth.days: LocalDateRange` and ranges of `YearMonth` values.
* Without https://github.com/Kotlin/kotlinx-datetime/pull/453, `YearMonth.until` returning `Long` is inconsistent with `LocalDate.until: Int`. Also, this PR removes the js-joda-based JS implementation, which makes it easier to make `YearMonth` an `expect` class if we want that.
* https://github.com/Kotlin/kotlinx-datetime/pull/415 affects which serializers are available and what they are called.
* https://github.com/Kotlin/kotlinx-datetime/pull/387 proposes to remove `monthNumber` and rename the `monthNumber: Int` constructor parameter to `month: Int`, which affects `YearMonth` as well.
* https://github.com/Kotlin/kotlinx-datetime/pull/373 could also implement `java.io.Serializable` for `YearMonth`.


dalewking:
About time we get this added. You need to resolve merge conflicts

dkhalanskyjb:
You are right! Thanks for the advice!

core/common/src/YearMonth.kt
```diff
@@ -0,0 +1,396 @@
...
+     * @sample kotlinx.datetime.test.samples.YearMonthSamples.numberOfDays
+     */
+    public val numberOfDays: Int get() = monthNumber.monthLength(isLeapYear(year))
+
+    /**
+     * Returns the range of days in the year-month.
+     *
+     * @sample kotlinx.datetime.test.samples.YearMonthSamples.days
+     */
+    // val days: LocalDateRange get() = firstDay..lastDay // no ranges yet
```
dalewking, line 130: What is the reason for commenting this out? Seems useful to me.

dalewking, line 130: Another useful range would be an open end range of LocalDateTime that is midnight on the first day of the given month until midnight on the first day of the next month

core/common/src/YearMonth.kt
```diff
@@ -0,0 +1,396 @@
...
+ */
+public fun YearMonth.plusMonth(): YearMonth = plus(1, DateTimeUnit.MONTH)
+
+/**
+ * The [YearMonth] one month earlier.
+ *
+ * @throws DateTimeArithmeticException if the result exceeds the boundaries of [YearMonth].
+ * @sample kotlinx.datetime.test.samples.YearMonthSamples.minusMonth
+ */
+public fun YearMonth.minusMonth(): YearMonth = minus(1, DateTimeUnit.MONTH)
```
dalewking, line 325: These 4 functions should be plural and accept an int parameter that defaults to 1 as in:
```
public fun YearMonth.minusMonths(months: Int = 1): YearMonth = minus(months, DateTimeUnit.MONTH)
```

core/common/src/YearMonth.kt
```diff
@@ -0,0 +1,396 @@
...
+     * @sample kotlinx.datetime.test.samples.YearMonthSamples.firstAndLastDay
+     */
+    public val firstDay: LocalDate get() = onDay(1)
+
+    /**
+     * Returns the last day of the year-month.
+     *
+     * @sample kotlinx.datetime.test.samples.YearMonthSamples.firstAndLastDay
+     */
+    public val lastDay: LocalDate get() = onDay(numberOfDays)
```
dalewking, line 116: You should add contains methods for LocalDate and LocalDateTime as in:
```
    public fun contains(date: LocalDate): Boolean = date in firstDay..lastDay
```

Also helpful would be intersects functions for ranges to determine if this year month has any overlap with another range. This should include the cases of `ClosedRange<LocalDate>` and `OpenEndRange<LocalDateTime>`. 

```
    public fun intersects(dateRange: ClosedRange<LocalDate>): Boolean = ...
    public fun intersects(dateTimeRange: OpenEndRange<LocalDateTime>): Boolean = ...
```

core/common/src/YearMonth.kt
dkhalanskyjb: The reason is that `LocalDateRange` is not in `master`. See the description of the pull request.

core/common/src/YearMonth.kt
dkhalanskyjb: In our research, we haven't found cases where the value was anything other than `1`, so the extra flexibility is pointless. You have a more general `plus` for this case if `1` doesn't suit your needs.

core/common/src/YearMonth.kt
dkhalanskyjb: There are no ranges of `LocalDateTime` semantically, because due to DST transitions, sometimes, "earlier" local date-times follow the "later" ones, and some local date-times don't exist at all.

dalewking: LocalDateTime implements Comparable so technically anything Comparable can be a range. Your argument is really that LocalDateTime should not be Comparable

dalewking: While LocalDateRange may not be in master, you can for now implement this as:
```
val days: ClosedRange<LocalDate> get() = firstDay..lastDay
```

dalewking: Actually, as of 2 days ago it looks like LocalDateRange is in master

core/common/src/YearMonth.kt
```diff
@@ -0,0 +1,387 @@
...
+ * is `2020-09` everywhere).
+ *
+ * Operations with [DateTimeUnit.MonthBased] are provided for [YearMonth]:
+ * - [YearMonth.plus] and [YearMonth.minus] allow expressing concepts like "two months later".
+ * - [YearMonth.until] and its shortcuts [YearMonth.monthsUntil] and [YearMonth.yearsUntil]
+ *   can be used to find the number of months or years between two dates.
+ *
+ * ### Platform specifics
+ *
+ * The range of supported years is platform-dependent, but at least is enough to represent year-months of all instants
```
ilya-g, line 32: Is it still true after we unified ranges across platforms?

PR #465, opened 2024-12-10
Remove kotlinx.datetime.Instant (branch 1: binary compatibility)

dkhalanskyjb:
The contents of `kotlinx.time` are the same as what `kotlin.time` will probably have, discounting the changes that may happen in https://github.com/Kotlin/kotlinx-datetime/pull/453 and https://github.com/Kotlin/kotlinx-datetime/pull/411 because of code review.


dkhalanskyjb:
Superseded by #506. The historical information of how this PR evolved may be interesting at some point, but this format is likely to make the review very complicated.

PR #475, opened 2025-01-23
Upgrade to Kotlin 2.1.0

dkhalanskyjb:
* Avoid using deprecated API when grouping source sets together
* Do not explicitly set `metaInfo = true` for NPM, it's the default
* Upgrade the Yarn lock file

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -6,6 +6,7 @@ import java.io.ByteArrayOutputStream
 import java.io.PrintWriter
 import org.jetbrains.dokka.gradle.AbstractDokkaLeafTask
 import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
+import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
 import org.jetbrains.kotlin.konan.target.Family
 
 plugins {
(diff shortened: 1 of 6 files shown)


core/build.gradle.kts
```diff
@@ -161,7 +157,26 @@ kotlin {
             }
         }
     }
+
     sourceSets {
+        groupSourceSets("linux", listOf("linuxX64", "linuxArm64", "linuxArm32Hfp"))
```
whyoleg, line 162: just in case: probably it could be easier to use KGP hierarchy API for this: [example from kotlinx.atomicfu](https://github.com/Kotlin/kotlinx-atomicfu/blob/b5db8fc8c9f6a85d2ccee8cdf58c68d8a368f7d4/atomicfu/build.gradle.kts#L131-L147), [example from ktor](https://github.com/ktorio/ktor/blob/4fd22824f87ccf30d294675e36168c92e2a97675/build-logic/src/main/kotlin/ktorbuild/targets/KtorTargets.kt#L114-L161)

dkhalanskyjb: I was wrong. After clearing all caches, `wasmJs` and `wasmWasi` are actually present.

ilya-g:
I wonder, does it still make sense to use `-trw` compiler argument for native test compilations with the current memory model?

FenstonSingel:
JiC: Kotlin 2.0.0+ uses `.kotlin/` instead of `.gradle/kotlin/` for dumping various intermediate/temporary/debug data, so it might be worth adding the former to `.gitignore` in advance.

core/build.gradle.kts
```diff
@@ -35,54 +36,78 @@ java {
...
-                        // Tier 2
-                        target("watchosX64")
-                        target("watchosArm32")
-                        target("watchosArm64")
-                        target("tvosX64")
-                        target("tvosArm64")
-                        target("iosArm64")
-                        // Tier 3
-                        target("watchosDeviceArm64")
+    // Tiers are in accordance with <https://kotlinlang.org/docs/native-target-support.html>
```
ilya-g, line 39: Minor: I'd move this after non-native targets

ilya-g (approved):

Issue #461, opened 2024-11-19
Add unambiguous LocalDateTime.toInstant


simonolander:
As the documentation states, the function `LocalDateTime.toInstant(timeZone: TimeZone)` can be ambiguous.

> Returns an instant that corresponds to this civil date/time value in the specified timeZone.
>
> Note that the conversion is not always well-defined. There can be the following possible situations:
> * There's only one instant that has this date/time value in the timeZone. In this case, the conversion is unambiguous.
> * There's no instant that has this date/time value in the timeZone. Such a situation appears when the time zone experiences a transition from a lesser to a greater offset. In this case, the conversion is performed with the lesser (earlier) offset, as if the time gap didn't occur yet.
> * There are two possible instants that can have this date/time components in the timeZone. In this case, the earlier instant is returned.

I would be greatly helped by a function that returns _all_ the instants that correspond to the `LocalDateTime` in the specified time zone. For example something with the following signature:
```kotlin
fun LocalDateTime.toInstants(timeZone: TimeZone): List<Instant>
```
If there is no such instant, the resulting list would be empty.

Would you consider adding this to the library?

dkhalanskyjb:
Sure, we're open to adding this. Could you share the business logic you need this for?

simonolander:
> Could you share the business logic you need this for?

I have recurring events that happen during certain time windows, e.g. between 12:00 and 18:00, or 02:00 and 02:30, local time. My need is to compute the start and end of each of these windows as pairs of instants for a given set of days. 

The first example is straight forward, but the second example event happens twice or never certain days, and I need to be able to account for that.

simonolander:
My use cases are the following:

* Display the time remaining until the start of the next event occurrence. This duration may be larger if the next occurrence doesn't happen due to a gap.
* If an event is ongoing, display the time remaining until it isn't. In case of an overlap, this duration may be larger than normal.
* For a given week or month, display the number of event occurrences, their duration, and when they occur. Overlaps may add occurrences, or prolong them. Gaps may remove occurrences, or shorten them.

I hope this answers your question!

As you say, there are many approaches that are ultimately equivalent. I think that my suggestion would be most ergonomic for my use case, but I could likely work with all the alternatives you proposed.

dkhalanskyjb:
Let's double-check my understanding using specific examples.

> the next occurrence doesn't happen due to a gap.

[...]

PR #491, opened 2025-02-20
unambiguos-instant

AlexCue987:
fix for https://github.com/Kotlin/kotlinx-datetime/issues/461

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,6 +393,39 @@ val localDateTimeTwoDaysLater = instantTwoDaysLater.toLocalDateTime(timeZone)
 // 2021-03-29T02:16:20
 ```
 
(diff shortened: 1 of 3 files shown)


dkhalanskyjb:
Hi! I don't understand how this implementation solves the use case described in #461, as it turns out an unambiguous `LocalDateTime.toInstant` wasn't actually enough for the author of the issue ticket. So, it's not clear that we need this functionality at all.

Even if we do need it, it's not clear in which form we should provide it: see https://github.com/Kotlin/kotlinx-datetime/issues/461#issuecomment-2487809642

We are not going to simply add an implementation without even discussing the general approach first. If you want to meaningfully contribute to the solution, a better approach is to describe your use cases in detail in a comment under #461 if you also require this API.

PR #472, opened 2025-01-08
Refer to DateTimeComponents field names in errors for null values

dkhalanskyjb:
Fixes #471

PR #500, opened 2025-03-05
fix: missing isoDayOfWeek

alfonsocv12:
Reading to changes on PR [472](https://github.com/Kotlin/kotlinx-datetime/pull/472) I thinks this were missing


dkhalanskyjb:
I'd prefer to keep the internal variable named `isoDayOfWeek`: it's a numerical value, and there are several ways of assigning numbers to days, so it's nice to have it unambiguous.
The name of the `dayOfWeek` variable in the date container does get exposed to the user, so we are limited in what we can call it.

PR #503, opened 2025-03-20
Support obtaining the system timezone on old Debian-based distributions

DmitryNekrasov:
## Issue Background

On older Debian-based distributions (between Etch 4.0 and Jessie 8.0), the system timezone configuration is stored differently than on most modern Linux distributions. Instead of using a symbolic link for `/etc/localtime`, these Debian versions used a regular file copied from the zoneinfo directory.

The current implementation in `kotlinx-datetime` only supports finding the timezone via the symlink method, which fails on these systems. Even though these Debian versions are past their support date, they remain in use in various environments, making this fix valuable for broader compatibility.

## Solution

I've modified the `currentSystemDefaultZone()` method to implement a fallback mechanism when the symlink approach fails:

1. First, attempt to determine the timezone via the symbolic link path (existing method)
2. If that fails, read the contents of the Debian-specific `/etc/timezone` file
3. Verify that the timezone is valid by comparing the contents of `/etc/localtime` with the corresponding zoneinfo file
4. If all attempts fail, throw an appropriate exception with a detailed message

This implementation gracefully handles all possible configurations while maintaining robust validation to prevent silent failures and incorrect timezone reporting.

## Testing Approach

To thoroughly test this fix, I created a comprehensive test suite using testcontainers that verifies correct behavior across different system configurations:

### Modern Systems (Ubuntu Noble)
- Correctly identifies timezone when properly configured
- Falls back to UTC when no timezone is configured
- Throws appropriate exception when symlinks are incorrect

### Legacy Systems (Debian Jessie)
- Reads timezone from `/etc/timezone` when configured correctly
- Falls back to "Universal" in default configuration
- Falls back to UTC when `/etc/localtime` is missing
- Throws appropriate exceptions for inconsistent or incorrect configurations

The testcontainers approach provides several advantages:
- Enables testing against real Linux distributions without modifying the developer's environment
- Creates isolated, reproducible test environments for each configuration scenario
- Allows validation against both legacy and modern systems
- Ensures the fix works correctly in real-world environments, not just simulated ones

Additionally, all standard timezone tests from the common package also pass in the correctly configured containers, confirming that core functionality remains intact.

## Summary

This fix significantly improves the robustness of the `kotlinx-datetime` library by extending timezone detection support to older Debian-based systems while maintaining strict validation. The comprehensive test suite ensures reliability across diverse system configurations.


core/jvm/test/testcontainers/TimezoneTestContainer.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    fun execDebianCopyTimeZoneTest(): ExecResult {
+        exec("rm /etc/localtime")
+        exec("cp /usr/share/zoneinfo/Europe/Berlin /etc/localtime")
+        exec("echo 'Europe/Berlin' > /etc/timezone")
+        return execTest("kotlinx.datetime.test.TimeZoneLinuxNativeTest.debianCopyTimeZoneTest")
+    }
+
+    fun execTimezoneMismatchTest(): ExecResult {
```
fzhinkin, line 34: Can we achieve the same result in terms of testing by performing all actions listed below in a temporary directory, let's say `/tmp/test-12345/` and then using this directory as a "root" inside `getTimezoneFromEtcTimezone`?
I mean, there will be a `/tmp/test-12345/etc/localtime` and `getTimezoneFromEtcTimezone`, once we kindly ask it, will check that file instead of the `/etc/localtime`.

core/linux/src/internal/TimeZoneNative.kt
```diff
@@ -16,10 +18,40 @@ internal actual fun getAvailableZoneIds(): Set<String> =
...
-internal actual fun currentSystemDefaultZone(): Pair<String, TimeZone?> {
-    // according to https://www.man7.org/linux/man-pages/man5/localtime.5.html, when there is no symlink, UTC is used
-    val zonePath = currentSystemTimeZonePath ?: return "Z" to null
-    val zoneId = zonePath.splitTimeZonePath()?.second?.toString()
-        ?: throw IllegalTimeZoneException("Could not determine the timezone ID that `$zonePath` corresponds to")
-    return zoneId to null
+@OptIn(ExperimentalForeignApi::class)
+private fun getTimezoneFromEtcTimezone(): String? {
+    val timezoneContent = Path.fromString("/etc/timezone").readBytes()?.toKString()?.trim() ?: return null
+    val zoneId = chaseSymlinks("/usr/share/zoneinfo/$timezoneContent")
```
fzhinkin, line 24: We may need to validate both `timezoneContent` and `zoneId` values to prevent potential path traversal attacks.

core/jvm/test/testcontainers/TimezoneTestContainer.kt
dkhalanskyjb: I'm not sure that this even needs to be a *temporary* directory, though. We are not modifying anything, after all, just reading. So, something like `kotlinx-datetime/core/linux/test-data/misconfigured-debian/` can be checked into the repo and used as the root.

core/linux/src/internal/TimeZoneNative.kt
DmitryNekrasov: I appreciate the suggestions for alternative testing approaches, but I'd like to make a comprehensive case for keeping the testcontainers implementation.

## Cross-Platform Development Reality

The timezone issue we're addressing is Linux-specific, but our development team works across various platforms. Testcontainers provides a consistent testing environment regardless of the host OS:

- Developers on macOS or Windows can test Linux-specific functionality without managing VMs
- No need to maintain different testing workflows for different operating systems
- Tests run the same way in local development and CI environments

A directory-based approach would require platform-specific setup instructions and potentially different implementations for different development environments.

## File System Modifications

The tests do more than just read files - they actively modify the filesystem to simulate different configurations:

- Removing files (`rm -f /etc/localtime`)
- Creating new files with specific content (`echo 'Europe/Berlin' > /etc/timezone`)
- Changing symlinks (`ln -sf /usr/share/zoneinfo/Europe/Paris /etc/localtime`)
- Copying files between locations (`cp /usr/share/zoneinfo/Europe/Berlin /etc/localtime`)

These operations would require either root permissions or a complex filesystem overlay mechanism if done outside a container. Using chroot would require sudo privileges, creating potential security and permission issues in development and CI environments.

## True Isolation and Reproducibility

Testcontainers provide complete isolation that's difficult to achieve with directory-based approaches:

- Each test runs in a fresh environment with known initial state
- Tests can't interfere with each other or with the host system
- No risk of permissions issues or unexpected side effects
- Containers can be configured with specific OS versions to match target environments

A directory-based approach would require careful cleanup between tests and couldn't guarantee the same level of isolation.

## Industry Standard Solution

Testcontainers is widely adopted for this exact use case across the industry:

- Well-maintained library with extensive documentation and community support
- Integrates seamlessly with test frameworks
- Developers familiar with containerization can easily understand and extend the tests
- Aligns with modern DevOps practices

Implementing a custom directory-based solution would require maintaining proprietary testing infrastructure rather than leveraging a battle-tested solution.

## Performance Considerations

The performance impact is minimal:

Screenshot 2025-03-20 at 23 31 30

- All testcontainer tests complete in approximately 12 seconds on a local machine
- This represents less than 1% of the total CI pipeline time (~20 minutes)
- The slight performance cost is outweighed by reliability and maintainability benefits

## Integration with Current Workflow

The testcontainers approach integrates well with the existing development workflow:

- Tests run as part of the normal test suite
- No special setup required beyond Docker being available
- Clear distinction between regular unit tests and environment-dependent tests
- Self-contained tests that document their own requirements and setup

While using a fixed directory in the repo might seem simpler, it would likely require additional scripts and documentation to explain how to use them properly, especially for developers unfamiliar with the codebase.

I believe the benefits of using testcontainers significantly outweigh the minimal performance cost, providing a more robust, maintainable, and developer-friendly testing solution for this platform-specific functionality.

core/jvm/test/testcontainers/TimezoneTestContainer.kt
dkhalanskyjb: > Cross-Platform Development Reality

Unfortunately, no amount of Docker containers will solve this issue. The Linux implementation is quite straightforward compared to the Windows one or even the macOS one. If we could reliably emulate Apple devices and the Windows registry on all platforms, this could be worth looking into. As is, relying on VMs and having various CI configurations is already unavoidable, so we are not winning anything: some bugs will only be caught by the CI.

Some actual issues that could be solved with virtualization (but unfortunately do not seem as straightforward):
* We don't test our `androidNative` implementation automatically at all, we only check our code manually when we change something Android-Native-specific, as this requires spinning up an Android emulator.
* This PR *could* significantly benefit from being tested in Docker (though not necessarily on every run of the test suite), if the image was some OS version actually suffering from the problem at hand. As is, it looks like (please correct me if I'm wrong) the Docker file is some new Ubuntu version that isn't actually affected until we manually introduce an issue; this means that we don't know whether for sure whether we are actually fixing the *existing problem* or *a fake problem emulated in our tests and unrelated to what our users could experience*. The fix matches my understanding of the problem, but of course, my understanding could be wrong. I'd need to grab a VM *affected by the problem* anyway to test this properly.

> File System Modifications

> The tests do more than just read files - they actively modify the filesystem to simulate different configurations

As written, tests consist of two phases: 1) preparing the test directory, and 2) executing the test. Phase 2) does  not modify any files. The results of phase 1) can be placed into the repository as test data.

> Using chroot would require sudo privileges

Testing does not require a chroot.
* Option 1 (can be tricky to implement, but is ultimately a nice solution) includes using https://github.com/dex4er/fakechroot: we don't believe that our own tests will attempt to hack our computers, so true isolation is not needed, and hopefully, we are not relying on any fancy `/proc` or `/sys` behaviors, so it's plausible that `fakechroot` is enough.
* Option 2 (recommended) is the pure library solution: introducing an `internal var fileSystemRoot = "/"` and (maybe) `internal fun resetTzdb()` and manipulating those should be enough to test that the code is correct.
* Option 3 (hardcore, not recommended) would introduce a new layer like `internal interface FileSystem` along with `internal var currentFilesystem: FileSystem` through which all filesystem access would go.

> Industry Standard Solution

There are many industry standard solutions, each useful in its own contexts, and it is necessary to make an informed choice. Mocking or dependency injection are other standard ways of testing effectful code. Having unit tests for small pieces of functionality and resorting to containers for integration tests are typical approaches, too.

> Integration with Current Workflow

I fail to see how other, lighter-weight solution disrupt our current workflow. That said, if a test for Linux doesn't integrate with our workflow, we have the right not to include it, according to https://kotlinlang.org/docs/native-target-support.html#tier-2.

core/linux/src/internal/TimeZoneNative.kt
fzhinkin: Yep, you got the idea. `timezoneContent` may contain some specially prepared path, that'll allow reading some data outside `/usr/share/zoneinfo` and then that data may leak through an exception message.

fzhinkin: Sorry, my bad: checked how `chaseSymlinks` works and now longer convinced that there's an easy way to exploit the concurrent implementation.

core/jvm/test/testcontainers/TimezoneTestContainer.kt
DmitryNekrasov: > This PR _could_ significantly benefit from being tested in Docker (though not necessarily on every run of the test suite), if the image was some OS version actually suffering from the problem at hand.

I agree with the suggestion and have already implemented exactly this approach. Rather than attempting to emulate the problematic behavior on modern Ubuntu 24.04, I've created test scenarios using actual Debian Jessie Docker images where this issue naturally occurs. The test suite now uses both Jessie-based and Noble-based containers to comprehensively cover all possible scenarios for the `currentSystemDefaultZone()` method. This testcontainer approach provides much higher confidence in the solution since we're testing against the actual affected system configuration rather than a simulation. The tests systematically verify each edge case in isolated, reproducible environments that genuinely reflect the diverse real-world scenarios our library needs to support.

core/linux/src/internal/TimeZoneNative.kt
```diff
@@ -16,10 +18,40 @@ internal actual fun getAvailableZoneIds(): Set<String> =
...
+            "Timezone mismatch: /etc/timezone specifies $displayTimezone " +
+                    "but /etc/localtime content differs from /usr/share/zoneinfo/$zoneId"
+        )
+    }
+
+    return zoneId
+}
+
 internal actual fun currentSystemDefaultZone(): Pair<String, TimeZone?> {
-    // according to https://www.man7.org/linux/man-pages/man5/localtime.5.html, when there is no symlink, UTC is used
```
dkhalanskyjb, line 20: This comment is still accurate, though?

core/jvm/test/testcontainers/TimeZoneConfigurationTest.kt
```diff
@@ -0,0 +1,148 @@
...
+
+    @Test
+    fun commonTimeZoneNobleTests() {
+        assertExecSuccess(nobleCorrectConfigContainer.execCommonTimeZoneTests())
+    }
+
+    private fun assertExecSuccess(execResult: ExecResult) {
+        logger.info("Container stdout:\n${execResult.stdout}")
+        logger.info("Container stderr:\n${execResult.stderr}")
+        logger.info("Container exit code: ${execResult.exitCode}")
```
dkhalanskyjb, line 106: Is there any need to duplicate what's already in the error message?

core/jvm/test/testcontainers/debian-jessie/incorrect-timezone.dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM --platform=linux/arm64 debian/eol:jessie
+
+ENV INSIDE_TESTCONTAINERS=true
+
+RUN echo incorrect/data > /etc/timezone
+
+WORKDIR /app
```
dkhalanskyjb, line 7: Why does the workdir need to be specified if tests are started using the absolute path?

core/jvm/test/testcontainers/TimeZoneConfigurationTest.kt
```diff
@@ -0,0 +1,148 @@
...
+    companion object {
+        private val logger = LoggerFactory.getLogger(TimeZoneConfigurationTest::class.java)
+
+        @JvmStatic
+        @BeforeAll
+        fun buildTestBinary() {
+            logger.info("Building test binary...")
+
+            val process = ProcessBuilder()
+                .command("../gradlew", "linkDebugTestLinuxArm64")
```
dkhalanskyjb, line 130: So, the process is:
* Gradle runs our tests.
* Linux tests include the per-container checks, but if they detect they are not running inside the development container, they will not run.
* JVM tests call Gradle to build Linux executables.
* Having built Linux executables, JVM tests launch a container per test and run the created binaries there, specifying that only a single test should run.

This feels way too complicated to me:
1. I believe tests that should run in a container should be in a separate binary from normal tests. This is the better separation of responsibilities: this way, it will be possible to compile the compact, to-the-point test suite for containers without also compiling all Linux tests, and vice versa, so there is less warning cross-pollution, better parallelization of tests, etc.
2. Running Gradle from the JVM (and then using hardcoded paths to access the result) seems suboptimal: the logging logic is custom; I don't know how build timeout logic will work with this indirection; this is reinventing the wheel. Why not introduce a proper dependency in Gradle instead? There can be a separate test run with its own dependencies that depends on the binary compilation task outputs and (probably) accepts the path to the resulting binary as a system property or something like that. This way, the build logic stays in Gradle, without leaking into tests; no need to introduce any loggers and such.
3. Whenever we change the JVM implementation, it makes sense to run all JVM tests in the core module to check that the implementation is still correct. However, with the containers being part of the JVM tests, this will no longer be as quick and straightforward, as it will take some extra time to run *Linux* tests for some reason. The test containers should be a separate test run, probably in another Gradle module even. We already have several modules purely for testing some specialized scenarios.

In essence, I propose the following:
* All container-related test logic is moved into a separate Gradle module.
* The logic of running the containers (be it `test-containers` or just pure Gradle code) depends (in the Gradle sense) on the binary containing the actual test logic. The logic of running the containers receives the path to the binary as a parameter.

The binary doesn't even necessarily have to use `kotlin.test`, it can be as simple as a `fun main(args: Array<String>)` that checks its argument to see which check to perform.

What we win:
* Less intricate code interactions, better readability, improved code locality.
* The CI is one step closer to our build logic and can control/analyze our tests better.
* Flexibility: if this is a separate module, we can easily conditionally enable or disable container-based tests altogether.
* Improved build parallelization.
* The dependency on loggers can be removed.
* No need for an environment variable check.

dkhalanskyjb (changes requested):

core/linux/src/internal/TimeZoneNative.kt
DmitryNekrasov: Not really, we return UTC if there is no `/etc/localtime` file, whether it is a symbolic link or a regular file. If there is a file but it is not a symlink, we suspect that the file is a copy of the timezone file and check for that further down the code. I suggest the next comment:
```kotlin
// According to https://www.man7.org/linux/man-pages/man5/localtime.5.html, UTC is used when /etc/localtime is missing.
// If /etc/localtime exists but isn't a symlink, we check if it's a copy of a timezone file by examining /etc/timezone
// (which is a Debian-specific approach used in older distributions).
```

core/jvm/test/testcontainers/debian-jessie/incorrect-timezone.dockerfile
DmitryNekrasov: Remove WORKDIR

core/linux/test/TimeZoneNativeTest.kt
```diff
@@ -0,0 +1,87 @@
...
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.assertTrue
+
+class TimeZoneNativeTest {
+
+    @Test
+    fun correctSymlinkTest() {
+        root = "${RESOURCES}correct-symlink/"
```
dkhalanskyjb, line 20: This way, if other tests trying to acquire the system timezone execute after this set of tests does, they will use the incorrect root. This needs either a `try`/`finally` or an `@AfterTest` restoring the value, and I'd introduce a function like `fun <T> withFakeRoot(root: String, action: () -> T): T` that would abstract this logic.

DmitryNekrasov: Done

core/linux/src/internal/TimeZoneNative.kt
```diff
@@ -16,10 +18,43 @@ internal actual fun getAvailableZoneIds(): Set<String> =
 
 private val tzdb = runCatching { TzdbOnFilesystem() }
 
+@OptIn(ExperimentalForeignApi::class)
+private fun getTimezoneFromEtcTimezone(): String? {
```
dkhalanskyjb, line 22: The logic in this function does not make it obvious why this is needed. Please add a short comment (like "this is needed in such and such versions of Debian") or a link to the issue that this function solves.

core/tzdbOnFilesystem/src/internal/TzdbOnFilesystem.kt
```diff
@@ -47,7 +47,9 @@ internal fun tzdbPaths(defaultTzdbPath: Path?) = sequence {
     currentSystemTimeZonePath?.splitTimeZonePath()?.first?.let { yield(it) }
 }
 
-internal val currentSystemTimeZonePath get() = chaseSymlinks("/etc/localtime")
+internal var root: String = "/"
```
dkhalanskyjb, line 50: This variable should either have a descriptive name explaining the limited scope it's applied to (that is, something like `systemTimezoneSearchRoot`?) or be applied consistently to all filesystem access (for example, to the list of locations to search for a timezone database in this file).

core/linux/test/TimeZoneNativeTest.kt
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Test
+    fun correctLocaltimeCopyTest() = withFakeRoot("${RESOURCES}correct-localtime-copy/") {
+        val tz = TimeZone.currentSystemDefault()
+        assertEquals(TimeZone.of("Europe/Oslo"), tz)
+    }
+
+    @Test
+    fun fallsBackToUTC() = withFakeRoot("${RESOURCES}falls-back-to-utc/") {
```
dkhalanskyjb, line 31: 1. I'd still create this directory and add an empty `.keep` file to git.
2. The name is not very descriptive. `falls-back-to-utc` in what scenarios? The name of the test, the name of the directory, a comment for this test, or all of the above should explain the specific logic: falling back to UTC when there is no `/etc/localtime`.

core/linux/test/TimeZoneNativeTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        val tz = TimeZone.currentSystemDefault()
+        assertEquals(TimeZone.UTC, tz)
+    }
+
+    @Test
+    fun missingTimezoneTest() = withFakeRoot("${RESOURCES}missing-timezone/") {
```
dkhalanskyjb, line 37: Ditto w.r.t. clearly describing what the test checks: it's normal that `/etc/timezone` doesn't exist, so what this test checks is what happens if `/etc/localtime` is a not a symlink, *but* the `/etc/timezone` file doesn't exist.

core/linux/test/TimeZoneNativeTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        val exception = assertFailsWith<IllegalTimeZoneException> {
+            TimeZone.currentSystemDefault()
+        }
+
+        assertTrue(
+            exception.message?.startsWith("Could not determine the timezone ID") == true,
+            "Exception message did not match"
+        )
```
dkhalanskyjb, line 45: I'm not sure this is worth checking. If we decide to reword the exception message, this check will be a nuisance, but I can't imagine it helping us catch a regression.

core/linux/test/TimeZoneNativeTest.kt
```diff
@@ -0,0 +1,85 @@
...
+class TimeZoneNativeTest {
+
+    @Test
+    fun correctSymlinkTest() = withFakeRoot("${RESOURCES}correct-symlink/") {
+        val tz = TimeZone.currentSystemDefault()
+        assertEquals(TimeZone.of("Europe/Oslo"), tz)
+    }
+
+    @Test
+    fun correctLocaltimeCopyTest() = withFakeRoot("${RESOURCES}correct-localtime-copy/") {
```
dkhalanskyjb, line 25: A more descriptive name is something like "timezone file agrees with localtime contents".

core/linux/test/TimeZoneNativeTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        assertTrue(
+            exception.message?.startsWith("Could not determine the timezone ID") == true,
+            "Exception message did not match"
+        )
+    }
+
+    @Test
+    fun incorrectTimezoneTest() = withFakeRoot("${RESOURCES}incorrect-timezone/") {
```
dkhalanskyjb, line 49: "non-existent timezone in timezone file"?

dkhalanskyjb (changes requested):

core/linux/test/TimeZoneNativeTest.kt
```diff
@@ -0,0 +1,75 @@
...
+    @Test
+    fun timezoneFileDisagreesWithLocaltimeContentsTest() = withFakeRoot("${RESOURCES}timezone-file-disagrees-with-localtime-contents/") {
+        val exception = assertFailsWith<IllegalTimeZoneException> {
+            TimeZone.currentSystemDefault()
+        }
+
+        assertTrue(
+            exception.message?.startsWith("Timezone mismatch") == true,
+            "Exception message did not start with 'Timezone mismatch'"
+        )
```
dkhalanskyjb, line 59: What *is* worth checking here is that the error message includes the contents of `/etc/timezone`. This should be true regardless of how we word the exception message.

dkhalanskyjb: I meant the *contents* of `/etc/timezone`, so `Europe/Oslo`.

DmitryNekrasov: I got it, thanks

dkhalanskyjb (approved):
Good job! I believe it's good to go. I've checked this on a Jessie VM, and the problem is both 1) real and 2) solved by this PR.

PR #509, opened 2025-04-02
Add `orNull` functions for non-throwing construction of datetime entities

### LINKED ISSUE #68: Add non-throwing LocalDate(Time) construction
There should be `null`-returning variants for creating `LocalDate(Time)` instances. It shouldn't be necessary to catch an exception. This is especially useful when creating instances from user input.

```kotlin
val date = runCatching { LocalDate(userInput.year, userInput.month, userInput.day) }.getOrNull()
```

dkhalanskyjb:
Fixes #68


core/common/test/LocalDateTimeTest.kt
```diff
@@ -131,6 +131,40 @@ class LocalDateTimeTest {
         assertFailsWith<IllegalArgumentException> { localTime(0, 0, 0, 1_000_000_000) }
     }
 
+    @Test
+    fun orNull() {// Test orNull with month number
```
ilya-g, line 136: Minor: comment on a new line

ilya-g (approved):

PR #518, opened 2025-04-13
Added LocalTime.Formats.ISO_BASIC (ISO 8601 basic format)

lucgirardin:


dkhalanskyjb:
ISO 8601 specifies that the format for time-of-day includes a leading `T`. The extended format can omit it, but a *basic* representation of time-of-day *must* always include `T`. Therefore, the format you're proposing is not a valid basic ISO time-of-day, and should not be named `ISO_BASIC`.

lucgirardin:
You are right, thanks for spotting this! I believe the same should also apply to LocalTime.Formats.ISO. Should I attempt to change it as well?

dkhalanskyjb:
> I believe the same should also apply to LocalTime.Formats.ISO.

The extended ISO format is allowed to omit the `T`, and since the format without the `T` is vastly more useful, we use that, so no.

In fact, regarding usefulness: is `LocalTime.Formats.ISO_BASIC` useful at all? When researching how people used date-time formats, I haven't seen anyone defining this format on their own.

lucgirardin:
I find it very useful, for example if one need to have a timestamp stored in a file name, in my case `weather-20250415T065500Z.parquet`.

core/common/src/format/LocalTimeFormat.kt
```diff
@@ -279,6 +279,23 @@ internal class LocalTimeFormat(override val actualFormat: CachedFormatStructure<
 }
 
 // these are constants so that the formats are not recreated every time they are used
+internal val ISO_TIME_BASIC by lazy {
+    LocalTimeFormat.build {
+        char('T')
```
dkhalanskyjb, line 284: Please see `internal val ISO_DATETIME`: our format is case-insensitive.

core/common/src/LocalTime.kt
```diff
@@ -197,6 +197,18 @@ public expect class LocalTime : Comparable<LocalTime> {
...
+         * ISO 8601 basic format.
+         *
+         * Examples: `1234`, `123456`, `123456.789`, `123456.1234`.
+         *
+         * When formatting, seconds are always included, even if they are zero.
+         * Fractional parts of the second are included if non-zero.
+         *
+         * @sample kotlinx.datetime.test.samples.LocalTimeSamples.Formats.isoBasic
+         */
+        public val ISO_BASIC: DateTimeFormat<LocalTime>
```
dkhalanskyjb, line 210: Please also add tests. Examples: `LocalDateTimeFormatTest#testIso`, `LocalDateFormatTest#testBasicIso`, etc.

core/common/src/LocalTime.kt
```diff
@@ -197,6 +197,18 @@ public expect class LocalTime : Comparable<LocalTime> {
      * [kotlinx.datetime.format.DateTimeFormat] for [LocalTime] values.
      */
     public object Formats {
+        /**
+         * ISO 8601 basic format.
+         *
+         * Examples: `1234`, `123456`, `123456.789`, `123456.1234`.
```
dkhalanskyjb, line 203: These examples are not valid for this format.

dkhalanskyjb (changes requested):
Got it! I don't see any problems with adding a `LocalTime.Formats.ISO_BASIC` if it is indeed useful, but by that logic, it also makes sense to add `LocalDateTime.Formats.ISO_BASIC` (which is the `20250415T065500` part in your message).

lucgirardin:
Should be complete now!

core/common/src/LocalDateTime.kt
```diff
@@ -194,6 +194,25 @@ public expect class LocalDateTime : Comparable<LocalDateTime> {
...
+         * Examples of datetime in ISO 8601 format:
+         * - `20200830T1843`
+         * - `+120200830T184300`
+         * - `00000830T184300.5`
+         * - `-00010830T184300.123456789`
+         *
+         * When formatting, seconds are included, only if they are non-zero.
+         * Fractional parts of the second are included if non-zero.
+         *
+         * See ISO-8601-1:2019, 5.4.2.1b), the version without the offset, together with
```
dkhalanskyjb, line 210:
```suggestion
         * See ISO-8601-1:2019, 5.4.2.1a), the version without the offset, together with
```
`b)` is the extended format.

core/common/src/LocalTime.kt
```diff
@@ -197,6 +197,18 @@ public expect class LocalTime : Comparable<LocalTime> {
...
      */
     public object Formats {
+        /**
+         * ISO 8601 basic format.
+         *
+         * Examples: `T1234`, `T123456`, `T123456.789`, `T123456.1234`.
+         *
+         * When formatting, seconds are always included, even if they are zero.
+         * Fractional parts of the second are included if non-zero.
+         *
```
dkhalanskyjb, line 207: I think `LocalDateTime.ISO_BASIC` is also worth mentioning (given that you personally were interested in `LocalDateTime.Formats.ISO_BASIC`, for example).

core/common/src/format/LocalTimeFormat.kt
```diff
@@ -279,6 +279,25 @@ internal class LocalTimeFormat(override val actualFormat: CachedFormatStructure<
...
+            // intentionally empty
+        }) {
+            optional {
+                second()
+            }
+            optional {
+                char('.')
+                secondFraction(1, 9)
+            }
+        }
```
dkhalanskyjb, line 297:
```suggestion
            optional {
                second()
                optional {
                    char('.')
                    secondFraction(1, 9)
                }
            }
        }
```
As the tests show, otherwise, `12:30:00.123` gets formatted as `T1230.123`, which is not valid: fractions of a second are only emitted together with the second value itself, even if it is zero.

dkhalanskyjb (changes requested):

lucgirardin:
Thanks for your careful review!

core/common/src/LocalDateTime.kt
```diff
@@ -194,6 +194,25 @@ public expect class LocalDateTime : Comparable<LocalDateTime> {
...
+        /**
+         * ISO 8601 basic format.
+         *
+         * Examples of datetime in ISO 8601 format:
+         * - `20200830T1843`
+         * - `+120200830T184300`
+         * - `00000830T184300.5`
+         * - `-00010830T184300.123456789`
+         *
+         * When formatting, seconds always included.
```
dkhalanskyjb, line 207:
```suggestion
         * When formatting, seconds are always included.
```

core/common/src/LocalDateTime.kt
```diff
@@ -194,6 +194,25 @@ public expect class LocalDateTime : Comparable<LocalDateTime> {
...
+         * Fractional parts of the second are included if non-zero.
+         *
+         * See ISO-8601-1:2019, 5.4.2.1a), the version without the offset, together with
+         * [LocalDate.Formats.ISO_BASIC] and [LocalTime.Formats.ISO_BASIC].
+         *
+         * @sample kotlinx.datetime.test.samples.LocalDateTimeSamples.Formats.basicIso
```
dkhalanskyjb, line 213: `LocalDateTimeSamples` needs to include the actual sample. Also, please use consistent naming: for `LocalDate` and `LocalTime`, this sample is called `isoBasic`.

core/common/test/format/LocalDateTimeFormatTest.kt
```diff
@@ -224,6 +224,41 @@ class LocalDateTimeFormatTest {
         test(dateTimes, LocalDateTime.Formats.ISO)
     }
 
+    @Test
+    fun testBasicIso() {
```
dkhalanskyjb, line 228: Please also add the corresponding test for `LocalTime.Formats.ISO_BASIC`.

core/common/test/format/LocalDateTimeFormatTest.kt
```diff
@@ -224,6 +224,41 @@ class LocalDateTimeFormatTest {
...
+        val dateTimes = buildMap<LocalDateTime, Pair<String, Set<String>>> {
+            put(LocalDateTime(2008, 7, 5, 0, 0, 0, 0), ("20080705T000000" to setOf("20080705t000000")))
```
dkhalanskyjb, line 230: Please also test the scenarios of parsing a string which does not include the seconds.

dkhalanskyjb (changes requested):
Another thing that I noticed just now: please also add the new constants to this list: https://github.com/Kotlin/kotlinx-datetime/blob/90be731a9ae47ecbed4f037e5a277550955a8ea3/core/common/src/format/DateTimeFormat.kt#L154-L171

PR #521, opened 2025-04-17
Use the "UTC" identifier for the TimeZone.UTC time zone

### LINKED ISSUE #474: TimeZone..of("UTC") isn't equal to TimeZone.UTC
It's very unintuitive that this assert fails.
```kotlin
assertEquals(TimeZone.UTC, TimeZone.of("UTC"))
```
I've re-discovered this fact multiple times when writing tests around datetime logic with time zones.

Presumably this is the same root cause as #222

dkhalanskyjb:
Fixes #474

### DIFF
--- a/core/common/test/TimeZoneTest.kt
+++ b/core/common/test/TimeZoneTest.kt
@@ -17,6 +17,7 @@ class TimeZoneTest {
         val utc: FixedOffsetTimeZone = TimeZone.UTC
         println(utc)
         assertEquals("UTC", utc.id)
+        assertEquals(utc, TimeZone.of("UTC"))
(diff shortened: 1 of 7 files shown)


core/common/test/TimeZoneTest.kt
```diff
@@ -16,7 +16,7 @@ class TimeZoneTest {
     fun utc() {
         val utc: FixedOffsetTimeZone = TimeZone.UTC
         println(utc)
-        assertEquals("Z", utc.id)
+        assertEquals("UTC", utc.id)
```
ilya-g, line 19: Can you add a check that TimeZone.UTC equals TimeZone.of("UTC")?

PR #522, opened 2025-04-26
Defend against attempts to bypass JVM serial proxy

lukellmann:


core/jvm/test/MaliciousJvmSerializationTest.kt
```diff
@@ -0,0 +1,182 @@
...
+import java.io.Serializable
+import kotlin.reflect.KClass
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.fail
+
+// TODO investigate other null stream (it's different from the one I got) from this comment:
+//  https://github.com/Kotlin/kotlinx-datetime/pull/373#discussion_r2009789491
+//  aced00057372001a6b6f746c696e782e6461746574696d652e4c6f63616c44617465618443f17dae33e70200014c00097472756556616c75657400154c6a6176612f74696d652f4c6f63616c446174653b787070
```
lukellmann, line 20: @dkhalanskyjb do you remember how you constructed that stream?

lukellmann: Using https://github.com/NickstaDB/SerializationDumper (seems to be a very helpful tool btw) I determined that the only difference is that you used a different field name in that stream (you used `trueValue`, I used `value`).

```java
package kotlinx.datetime;

import java.io.*;
import java.util.*;

public class LocalDate implements Serializable {
    private final java.time.LocalDate trueValue = null;

    @Serial
    private static final long serialVersionUID = 7026816023079564263L;

    public static void main(String[] args) throws IOException {
        var bos = new ByteArrayOutputStream();
        try (var oos = new ObjectOutputStream(bos)) {
            oos.writeObject(new LocalDate());
        }
        // prints your stream:
        // aced00057372001a6b6f746c696e782e6461746574696d652e4c6f63616c44617465618443f17dae33e70200014c00097472756556616c75657400154c6a6176612f74696d652f4c6f63616c446174653b787070
        System.out.println(HexFormat.of().formatHex(bos.toByteArray()));
    }
}
```

core/jvm/src/LocalDate.kt
```diff
@@ -52,6 +52,11 @@ public actual class LocalDate internal constructor(
         @Suppress("FunctionName")
         public actual fun Format(block: DateTimeFormatBuilder.WithDate.() -> Unit): DateTimeFormat<LocalDate> =
             LocalDateFormat.build(block)
+
+        // even though this class uses writeReplace (so serialVersionUID is not needed for a stable serialized form), a
+        // stable serialVersionUID means exceptions caused by deserialization of malicious streams will be consistent
+        // (InvalidClassException vs. InvalidObjectException, see MaliciousJvmSerializationTest)
```
ilya-g, line 58: Is this really needed? As soon as some runtime serialization exception is thrown on a malicious stream, does it really matter which one?

core/jvm/test/MaliciousJvmSerializationTest.kt
```diff
@@ -0,0 +1,207 @@
+/*
+ * Copyright 2019-2025 JetBrains s.r.o. and contributors.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.datetime
```
ilya-g, line 6: Better to use `kotlinx.datetime.test` package.

ilya-g (changes requested):

core/jvm/src/LocalDate.kt
lukellmann: I realized that explicitly declaring the `serialVersionUID` is needed to properly do the tests in `MaliciousJvmSerializationTest`. The malicious serial streams in that test are designed to bypass the proxy given no defensive `readObject` method. But that only works if they have the correct `serialVersionUID` in them. So they would have to be regenerated each time the `serialVersionUID` changes (which e.g. happened for `LocalDate` in d15ec21e1d53944168357bf70934fec46107730f due to new methods in that class, it is now `-9141474295923275006L`). Otherwise we would just test that `ObjectInputStream` detects and rejects invalid `serialVersionUID`s.

lukellmann:
For 89fd8182e94bdde3453e2015b2e708a8d38bbc8a serialver now outputs this (`LocalDate` has a different `serialVersionUID`):
```
$ serialver -classpath .m2/repository/org/jetbrains/kotlinx/kotlinx-datetime-jvm/0.6.2-SNAPSHOT/kotlinx-datetime-jvm-0.6.2-SNAPSHOT.jar:kotlin-stdlib-2.1.20.jar kotlinx.datetime.LocalDate kotlinx.datetime.LocalDateTime kotlinx.datetime.LocalTime kotlinx.datetime.UtcOffset kotlinx.datetime.DateTimeArithmeticException kotlinx.datetime.IllegalTimeZoneException kotlinx.datetime.DateTimeFormatException kotlinx.datetime.internal.format.parser.ParseException
kotlinx.datetime.LocalDate:    private static final long serialVersionUID = -9141474295923275006L;
kotlinx.datetime.LocalDateTime:    private static final long serialVersionUID = -4261744960416354711L;
kotlinx.datetime.LocalTime:    private static final long serialVersionUID = -352249606036216323L;
kotlinx.datetime.UtcOffset:    private static final long serialVersionUID = -6636773355667981618L;
kotlinx.datetime.DateTimeArithmeticException:    private static final long serialVersionUID = -3207806170214997982L;
kotlinx.datetime.IllegalTimeZoneException:    private static final long serialVersionUID = 1159315966274264801L;
kotlinx.datetime.DateTimeFormatException:    private static final long serialVersionUID = 4231196759387994100L;
kotlinx.datetime.internal.format.parser.ParseException:    private static final long serialVersionUID = 5691186997393344103L;
```
I won't update it here though. The actual value doesn't matter, non-malicious serialization will not use it since it goes through `kotlinx.datetime.Ser`. It's just important that it stays stable for `MaliciousJvmSerializationTest` after this PR is merged, see https://github.com/Kotlin/kotlinx-datetime/pull/522#discussion_r2118502411.

lukellmann:
I've extracted the changes unrelated to the bypassing the proxy into #533.

ilya-g (approved):

lukellmann:
This PR shouldn' t be merged until `YearMonth` is included here. I hope I can find some time for this in the next few days.

lukellmann:
Done.

Issue #444, opened 2024-10-16
DateTimeComponents.Format parse throws an exception when timeZoneId() is UTC


LotuxPunk:
I have this piece of code to handle a Long date (20241016153100) to an ISO8106 date, it works for most of the cases, but if `tz` is UTC, the `id` will be `Z`, and throw an exception.

```kotlin
DateTimeComponents.Format {
    year()
    monthNumber()
    dayOfMonth()
    hour()
    minute()
    second()
    char('|')
    timeZoneId()
}.parse("$date|${(tz.id)}").let {
    buildString {
        append(it.year.toString().padStart(2, '0'))
        append('-')
        append(it.monthNumber.toString().padStart(2, '0'))
        append('-')
        append(it.dayOfMonth.toString().padStart(2, '0'))
        append('T')
        append(it.hour.toString().padStart(2, '0'))
        append(':')
        append(it.minute.toString().padStart(2, '0'))
        append(':')
        append(it.second.toString().padStart(2, '0'))
        append(tz.offsetAt(it.toInstantUsingOffset()).asTimeZone().offset.toString())
    }
}
```

```
Errors: position 15: 'Expected timeZoneId but got ', position 0: 'Expected - but got 2', position 0: 'Expected + but got 2'kotlinx.datetime.DateTimeFormatException: Failed to parse value from '20231013000000|Z'
```

The workaround I have now is to change `Z` to `Europe/London` and it seems to works, but it would seems fair that timeZoneId could recognise `Z` as a valid timeZoneId.

dkhalanskyjb:
Would this work instead?
```kotlin
import kotlinx.datetime.*
import kotlinx.datetime.format.*

fun main() {
    val tz = TimeZone.UTC
    val date = "20231013000000"
    val ldt = LocalDateTime.Format {
        year()
        monthNumber()
        dayOfMonth()
        hour()
        minute()
        second()
    }.parse(date)
    val instant = ldt.toInstant(tz)
    println(instant.format(DateTimeComponents.Formats.ISO_DATE_TIME_OFFSET, tz.offsetAt(instant)))
}
```
Runnable version: https://pl.kotl.in/wJEhu6YbS

* Why put the timezone into a string just to immediately parse it back?
* `.asTimeZone().offset` shouldn't do anything.
[...]

PR #528, opened 2025-05-16
DateTimeComponents.Format parse throws an exception when timeZoneId() is UTC

DmitryNekrasov:


DmitryNekrasov:
@dkhalanskyjb Dmitry, hello! Could you please look at the added tests, do they correctly document the expected behavior and fully cover possible scenarios? Thank you!

core/common/test/TimeZoneTest.kt
```diff
@@ -246,6 +248,66 @@ class TimeZoneTest {
...
+            )
+        )
+    }
+
+    private fun assertTimeZoneIdCanBeParsed(zoneIds: List<String>) {
+        zoneIds.forEach { zoneId ->
+            try {
+                val tz = TimeZone.of(zoneId)
+                val result = DateTimeComponents.Format { timeZoneId() }.parse(zoneId)
+                assertEquals(tz.id, result.timeZoneId)
```
dkhalanskyjb, line 304: I don't think `parse` should do any normalization the way `of` does. So, I'd expect `UTC+0800` to be parsed into the string `UTC+0800`, but the timezone `UTC+08:00`.

dkhalanskyjb (changes requested):

core/common/src/internal/format/parser/ParserOperation.kt
```diff
@@ -139,6 +139,113 @@ internal class UnconditionalModification<Output>(
...
+                index += length
+                lastValidIndex = index
+                return true
+            }
+
+            var state = State.START
+            while (index < input.length) {
+                state = when (state) {
+                    State.START -> when {
+                        input[index] == 'Z' -> {
```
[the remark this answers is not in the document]
DmitryNekrasov: I wrote the next simple test:
```kotlin
@Test
fun test() {
    val zoneId = "z"
    val tz = TimeZone.of(zoneId)
}
```
, and got the exception:
```
Invalid ID for ZoneOffset, invalid format: z
java.time.DateTimeException: Invalid ID for ZoneOffset, invalid format: z
	at java.time.ZoneOffset.of(ZoneOffset.java:241)
	at java.time.ZoneId.of(ZoneId.java:405)
	at java.time.ZoneId.of(ZoneId.java:359)
	at kotlinx.datetime.TimeZone$Companion.of(TimeZoneJvm.kt:39)
	at kotlinx.datetime.test.TimeZoneTest.test(TimeZoneTest.kt:254)
```
, but with capital 'Z' it passes.

core/common/src/internal/format/parser/ParserOperation.kt
```diff
@@ -139,6 +139,130 @@ internal class UnconditionalModification<Output>(
...
+                val value = input.substring(index, index + length).toInt()
+                if (value > maxValue || !additionalValidation(value)) return INVALID_TIME
+                index += length
+                lastValidIndex = index
+                return value
+            }
+
+            fun validateHH() = validateTimeComponent(2, 18)
+            fun validateH() = validateTimeComponent(1, 9)
+            fun validateMM() = validateTimeComponent(2, 59) { minutes -> hours < 18 || minutes == 0 }
```
dkhalanskyjb, line 200: There are also seconds (`UTC+011530`, `UTC+01:15:30` can both be passed to `TimeZone.of`), but before implementing their support, let's have a discussion tomorrow about which behaviors we want.

core/common/src/internal/format/parser/ParserOperation.kt
```diff
@@ -139,6 +139,130 @@ internal class UnconditionalModification<Output>(
...
+        val lastMatch = validateTimezone(input, startIndex)
+        if (lastMatch > startIndex) {
+            setter.setWithoutReassigning(storage, input.substring(startIndex, lastMatch), startIndex, lastMatch)
+            return ParseResult.Ok(lastMatch)
+        }
+        return ParseResult.Error(startIndex) { "Invalid timezone format" }
+    }
+
+    companion object {
+        enum class State {
```
dkhalanskyjb, line 157: Can't it be `private`?

core/common/src/internal/format/parser/ParserOperation.kt
```diff
@@ -139,6 +139,130 @@ internal class UnconditionalModification<Output>(
...
+                if (hasEnoughChars(length) && input.substring(index, index + length) in validValues) {
+                    index += length
+                    lastValidIndex = index
+                    return true
+                }
+                return false
+            }
+
+            fun validateTimeComponent(length: Int): Boolean {
+                if (hasEnoughChars(length) && input.slice(index..<(index + length)).all { it.isDigit() }) {
```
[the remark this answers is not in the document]
dkhalanskyjb: Another simplifications:
```suggestion
                if ((index..<(index + length)).all { input.getOrNull(it)?.isAsciiDigit() ?: false }) {
```

core/common/src/internal/format/parser/ParserOperation.kt
```diff
@@ -139,6 +139,130 @@ internal class UnconditionalModification<Output>(
...
+        }
+
+        private fun validateTimezone(input: CharSequence, startIndex: Int): Int {
+            var index = startIndex
+            var lastValidIndex = startIndex
+
+            fun hasEnoughChars(length: Int) = index + length <= input.length
+
+            fun validatePrefix(length: Int, validValues: List<String>): Boolean {
+                if (hasEnoughChars(length) && input.substring(index, index + length) in validValues) {
```
dkhalanskyjb, line 176:
```suggestion
                if (validValues.any { input.startsWith(it) }) {
```
With this and another simplification, `hasEnoughChars` is no longer needed.

core/common/src/internal/format/parser/ParserOperation.kt
```diff
@@ -139,6 +139,130 @@ internal class UnconditionalModification<Output>(
...
+                if (hasEnoughChars(length) && input.slice(index..<(index + length)).all { it.isDigit() }) {
+                    index += length
+                    lastValidIndex = index
+                    return true
+                }
+                return false
+            }
+
+            var state = State.START
+            while (index < input.length) {
```
dkhalanskyjb, line 194: This state machine accepts `+12:3456` and `+1234:56`, none of which is in a format that `TimeZone.of` accepts.

dkhalanskyjb (changes requested):

core/common/test/format/DateTimeComponentsFormatTest.kt
```diff
@@ -268,4 +267,127 @@ class DateTimeComponentsFormatTest {
...
+                }
+
+                ParseExpectation.SHOULD_FAIL_TO_PARSE -> {
+                    assertFailsWith<DateTimeFormatException> {
+                        DateTimeComponents.Format { timeZoneId() }.parse(zoneId)
+                    }
+                }
+            }
+        }
+    }
```
dkhalanskyjb, line 392: I don't see the benefit of this being one function instead of being three functions. In all usages, this is called with a constant value for `expectation`, and
```kotlin
assertTimezoneParsingBehavior(TimezoneTestData.invalidTimezoneIds, ParseExpectation.SHOULD_FAIL_TO_PARSE)
```
is longer than
```kotlin
assertNonParseableAsTimeZone(TimezoneTestData.invalidTimezoneIds)
```

This is not much shorter than
```kotlin
TimezoneTestData.invalidTimezoneIds.forEach(::assertNonParseableAsTimeZone)
```
which I believe to be the clearest version, given that we also pass one-element lists to `assertTimezoneParsingBehavior` in some places.

core/common/test/format/DateTimeComponentsFormatTest.kt
```diff
@@ -268,4 +267,127 @@ class DateTimeComponentsFormatTest {
...
+        )
+
+        val invalidTimezoneIds = listOf("INVALID", "XYZ", "ABC/DEF", "NOT_A_TIMEZONE", "SYSTEM")
+    }
+
+    @Test
+    fun testZuluTimeZone() {
+        // Replace it to:
+        // assertTimezoneParsingBehavior(listOf("Z", "z"), ParseExpectation.SHOULD_PARSE_CORRECTLY)
+        // when TimeZone.of("z") will work correctly
```
dkhalanskyjb, line 316:
```suggestion
        // when TimeZone.of("z") works correctly
```

core/common/test/format/DateTimeComponentsFormatTest.kt
```diff
@@ -268,4 +267,127 @@ class DateTimeComponentsFormatTest {
...
+        val incorrectUnparsableOffsets = listOf(
+            "a", "_", "+",                                      // Single non-digit characters
+            "a9", "y!", "1#",                                   // Two characters: letter+digit, letter+symbol, digit+symbol
+            "110", "020",                                       // Three digits (invalid length - not 2 or 4 digits)
+            "18000", "02300",                                   // Five digits (invalid length - not 4 or 6 digits)
+            "3:10", "2:70", "99:", ":20",                       // HH:MM format violations: single digit hour, missing minute, missing hour
+            "12:3456", "1234:56",                               // Invalid colon-separated formats: too many digits in an hour/minute component
+            "1:00:00", "00:6:00", "09:99:9",                    // HH:MM:SS format violations: single digit hour, single digit minute, single digit second
+            ":00:00", "00::00", "09:99:", "::00", "00::", "::", // Colon placement errors
+            "180:00:00", "00:610:00", "99:99:199"               // HH:MM:SS format violations: 3-digit hour, 3-digit minute, 3-digit second
```
dkhalanskyjb, line 298: These lines are a bit too long (we usually stick to about 120 characters per line), and this can be easily avoided like this:
```kotlin
// comment1
"group1, item 1", "group1, item 2",
// comment 2
"group2, item 1", "group2, item 2",
```

core/common/src/internal/format/parser/ParserOperation.kt
```diff
@@ -139,6 +139,118 @@ internal class UnconditionalModification<Output>(
     }
 }
 
+internal class TimeZoneParserOperation<Output>(
+    private val setter: AssignableField<Output, String>
+) : ParserOperation<Output> {
+
+    override fun consume(storage: Output, input: CharSequence, startIndex: Int): ParseResult {
+        if (startIndex >= input.length) return ParseResult.Error(startIndex) { "Unexpected end of input" }
```
dkhalanskyjb, line 147: I think that this line should either mention that a timezone is expected, or be removed altogether (as the error case of empty input is covered later anyway).

core/common/src/internal/format/parser/ParserOperation.kt
```diff
@@ -139,6 +139,118 @@ internal class UnconditionalModification<Output>(
...
+        val lastMatch = validateTimezone(input, startIndex)
+        if (lastMatch > startIndex) {
+            setter.setWithoutReassigning(storage, input.substring(startIndex, lastMatch), startIndex, lastMatch)
+            return ParseResult.Ok(lastMatch)
+        }
+        return ParseResult.Error(startIndex) { "Invalid timezone format" }
```
dkhalanskyjb, line 153:
```suggestion
        return if (lastMatch > startIndex) {
            setter.setWithoutReassigning(storage, input.substring(startIndex, lastMatch), startIndex, lastMatch)
            ParseResult.Ok(lastMatch)
        } else {
            ParseResult.Error(startIndex) { "Invalid timezone format" }
        }
```

dkhalanskyjb (approved):
Yes, this should do the trick, I only have minor suggestions. Please squash when merging and mention `Fixes #444` in the full commit message.

PR #534, opened 2025-06-04
Fix various issues with datetime arithmetics

### LINKED ISSUE #484: Broken guarantee in Instant.periodUntil
Quoting https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/period-until.html:
> All components of the `DateTimePeriod` returned are [...] positive or zero if this instant is earlier than the other.

```kotlin
val tz = TimeZone.of("Europe/Berlin")
println(LocalDateTime(2025, 3, 29, 2, 30).toInstant(tz).periodUntil(LocalDateTime(2025, 3, 30, 3, 10).toInstant(tz), tz))
// P1DT-20M
```
Runnable: https://pl.kotl.in/LedZ6Njqf

### LINKED ISSUE #487: Adding `DateTimePeriod` to `Instant` is not fully ISO-8601-compliant
ISO 8601 defines the period of one month as a fixed number of calendar days, specified by the calendar.

In `DateTimePeriod` + `Instant` arithmetics, we treat a month as a separate time scale unit, namely, we resolve the `LocalDateTime` on which we perform arithmetics internally after adding the months but before adding the days.

Adding months and days should instead be done as a single change, only consulting the calendar, and the time zone should only be used once all calendar-based operations are performed.

Earlier, I attempted to implement a solution in (https://github.com/Kotlin/kotlinx-datetime/pull/330) and didn't create an issue ticket, hoping for a quick resolution, but at this point, I should file it so that this doesn't get forgotten about.

### LINKED ISSUE #488: Broken guarantee in Instant.until
Almost equivalent to https://github.com/Kotlin/kotlinx-datetime/issues/484, but more pronounced.

From the documentation:

> Returns the whole number of the specified date or time [units](https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/until.html) between this and [other](https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/until.html) instants in the specified [timeZone](https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/until.html).

This is not true: "the whole number" implies "the number rounded towards zero to the nearest whole number", but that's not what we get:

```kotlin
import kotlinx.datetime.*

fun main() {
    val tz = TimeZone.of("Europe/Berlin")
    val i1 = LocalDateTime(2025, 3, 29, 2, 30).toInstant(tz)
    val i2 = LocalDateTime(2025, 3, 30, 3, 10).toInstant(tz)
    println(i2 - i1.plus(i1.until(i2, DateTimeUnit.DAY, tz), DateTimeUnit.DAY, tz)) // -20m
}
```

`b - (a + floor(b - a))` should be non-negative for positive `b` and `a`, but here, it's not.

dkhalanskyjb:
Fixes #488
Fixes #487
Fixes #484

### DIFF
--- a/core/common/src/Instant.kt
+++ b/core/common/src/Instant.kt
@@ -66,27 +66,10 @@ public fun Instant.Companion.parse(
 public fun Instant.plus(period: DateTimePeriod, timeZone: TimeZone): Instant = try {
     with(period) {
         val initialOffset = offsetIn(timeZone)
-        val initialLdt = toLocalDateTimeFailing(initialOffset)
-        val instantAfterMonths: Instant
-        val offsetAfterMonths: UtcOffset
-        val ldtAfterMonths: LocalDateTime
-        if (totalMonths != 0L) {
-            val unresolvedLdtWithMonths = initialLdt.plus(totalMonths, DateTimeUnit.MONTH)
-            instantAfterMonths = localDateTimeToInstant(unresolvedLdtWithMonths, timeZone, preferred = initialOffset)
-            offsetAfterMonths = instantAfterMonths.offsetIn(timeZone)
-            ldtAfterMonths = instantAfterMonths.toLocalDateTime(offsetAfterMonths)
-        } else {
-            instantAfterMonths = this@plus
-            offsetAfterMonths = initialOffset
-            ldtAfterMonths = initialLdt
-        }
-        val instantAfterMonthsAndDays = if (days != 0) {
-            val unresolvedLdtWithDays = ldtAfterMonths.plus(days, DateTimeUnit.DAY)
-            localDateTimeToInstant(unresolvedLdtWithDays, timeZone, preferred = offsetAfterMonths)
-        } else {
-            instantAfterMonths
-        }
-        instantAfterMonthsAndDays
+        val ldtPlusDate = toLocalDateTimeFailing(initialOffset)
+            .run { if (totalMonths != 0L) { plus(totalMonths, DateTimeUnit.MONTH) } else { this } }
+            .run { if (days != 0) { this.plus(days, DateTimeUnit.DAY) } else { this } }
+        localDateTimeToInstant(ldtPlusDate, timeZone, preferred = initialOffset)
             .run { if (totalNanoseconds != 0L) plus(totalNanoseconds.nanoseconds).check(timeZone) else this }
     }.check(timeZone)
 } catch (e: ArithmeticException) {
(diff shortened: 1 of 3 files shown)


core/common/src/Instant.kt
```diff
@@ -171,8 +154,18 @@ public fun Instant.periodUntil(other: Instant, timeZone: TimeZone): DateTimePeri
  */
 public fun Instant.until(other: Instant, unit: DateTimeUnit, timeZone: TimeZone): Long =
     when (unit) {
-        is DateTimeUnit.DateBased ->
-            toLocalDateTimeFailing(offsetIn(timeZone)).until(other.toLocalDateTimeFailing(other.offsetIn(timeZone)), unit)
```
ilya-g, line 175: LDT.until seems to be unused after that

core/common/src/Instant.kt
```diff
@@ -171,8 +154,18 @@ public fun Instant.periodUntil(other: Instant, timeZone: TimeZone): DateTimePeri
...
+        is DateTimeUnit.DateBased -> {
+            val start = toLocalDateTimeFailing(timeZone)
+            val end = other.toLocalDateTimeFailing(timeZone)
+            val timeAfterAddingDate =
+                localDateTimeToInstant(end.date.atTime(start.time), timeZone, preferred = this.offsetIn(timeZone))
```
ilya-g, line 161: What do you think about another way: calculating the difference using LD.until, than adding it back to the this instant and correcting it by one if we have overshot?

dkhalanskyjb: Like so?
```kotlin
val start = toLocalDateTimeFailing(timeZone)
val end = other.toLocalDateTimeFailing(timeZone)
val differenceEstimate = start.date.until(end.date, unit)
val otherEnd = plus(differenceEstimate, unit, timeZone)
val delta = when {
    other > this && otherEnd > other -> -1 // addition won't throw: end date - date >= 1
    other < this && otherEnd < other -> 1 // addition won't throw: date - end date >= 1
    else -> 0
}
differenceEstimate + delta
```
This should work.

For `periodUntil`, the change is not so straightforward. For example, if we overshoot by a day and end up with `DatePeriod(months = 1, days = 0)`, the correction is timezone-dependent and can result in `DatePeriod(days = 27|28|29|30)`. Quite a mouthful.

I like symmetry between `until` and `periodUntil`, so I'd prefer to keep the current approach.

ilya-g: I'm not sure that this symmetry remains after introduction of overloads with resolvers, but time will show.
In general I agree with the behavior in the tests introduced here, so the actual implementation can be either  👍

ilya-g (approved):

PR #538, opened 2025-06-18
fix: Handle DST transitions correctly in Instant.periodUntil

sakaguchi-yuki:


dkhalanskyjb:
The code does not even compile, and even if it did, it does not solve the problem. Here's another PR that does: https://github.com/Kotlin/kotlinx-datetime/pull/534

Issue #485, opened 2025-02-14
currentSystemDefault() iOS crash with "Exception: Invalid zone ID: Europe/Paris"


stefanhaustein:
```
Terminating app due to uncaught exception 'kotlinx.datetime.IllegalTimeZoneException', reason: 'Invalid zone ID: Europe/Paris'

(TimeZone.kt:78) |   | kfun:kotlinx.datetime.TimeZone.Companion#of(kotlin.String){}kotlinx.datetime.TimeZone
(TimeZone.kt:31) |   | kfun:kotlinx.datetime.TimeZone.Companion#currentSystemDefault(){}kotlinx.datetime.TimeZone
``` 

Device data for an example instance

- iPad11,6
- iOS 17.6.1 21G93

Kotlinx DateTime Version: 0.6.1

dkhalanskyjb:
I've double-checked that `TimeZone.currentSystemDefault()` works on kotlinx-datetime 0.6.2 with both iOS 17.1.1 (before an update) and on 18.3.1 (after an update) on a physical (non-Simulator) iPhone 12 mini. I don't know where the difference between the two devices could come from or what custom settings your device has that causes this.

stefanhaustein:
P.S: Sorry, the hex string is probably a red herring -- it might be added by our system to create a unique identifier.

stefanhaustein:
Any ideas how this could happen? Europe/Paris seems to be in your region database...

dkhalanskyjb:
If the timezone is `Europe/Paris-9a420787`, then it means that this invocation https://github.com/Kotlin/kotlinx-datetime/blob/e69f92fbf51866459be6b12710784a45c1063483/core/darwin/src/internal/TimeZoneNative.kt#L68 returned such a string, which is clearly an incorrect time zone name.

If the timezone is `Europe/Paris`, which is clearly a valid time zone name, it means we can no longer rely on the location of the timezone database in https://github.com/Kotlin/kotlinx-datetime/blob/e69f92fbf51866459be6b12710784a45c1063483/core/darwinDevices/src/internal/TimeZoneNative.kt

We can work around both, probably, but we do need to understand where the problem lies first.

stefanhaustein:
Sorry about the red herring. The Time Zone was "Europe/Paris", and we see this for other time zones, too:

- Invalid zone ID: America/Mexico_City
- Invalid zone ID: Asia/Ho_Chi_Minh
- Invalid zone ID: Europe/Rome
[...]

PR #539, opened 2025-06-19
Implement Foundation-based time zone support for Darwin systems with comprehensive DST handling

DmitryNekrasov:
## Summary

This PR addresses issue #485 by leveraging Apple’s native Foundation time zone data if tzdb is not available. The goal is to make time zone computations work correctly on Darwin devices if tzdb is not available for some reasons, while ensuring the behavior remains consistent with other platforms.

## Solution: Abstract Time Zone Rules and Introduce Foundation Implementation

The core of the solution is to abstract time zone logic behind a new internal interface, `TimeZoneRules`. This interface (introduced in this PR) defines the essential operations needed for time zone conversions: retrieving the offset at a given `Instant` (`infoAtInstant`) and obtaining detailed offset transition info for a given `LocalDateTime` (`infoAtDatetime`). By extracting this interface, we can provide multiple implementations of time zone rules. The existing tzdb-based logic has been refactored into an implementation called `TimeZoneRulesCommon` (which reads from tzdb as before), and a new implementation called `TimeZoneRulesFoundation` has been added for Darwin platforms. The TimeZone API now uses these abstractions under the hood, selecting the appropriate rules implementation based on the platform. For example, on iOS or macOS, `TimeZone.of(zoneId)` will use the Foundation-backed rules (if tzdb is not available), whereas on other targets it continues to use tzdb data.

## Implementation Details of TimeZoneRulesFoundation (Darwin)

The `TimeZoneRulesFoundation` class uses Apple’s Foundation framework (`NSTimeZone`, `NSCalendar`, `NSDate`) to obtain time zone information. Upon initialization, it looks up the `NSTimeZone` by the zone identifier. The `infoAtInstant(instant)` function is straightforward: it converts the Instant to an `NSDate` and calls `NSTimeZone.secondsFromGMT(for: date)` to get the raw UTC offset in seconds, returning it as a `UtcOffset`. The more complex logic resides in `infoAtDatetime(localDateTime)`, which must handle situations like Daylight Saving Time transitions (gaps and overlaps in local time). This method uses an `NSCalendar` set to the specific time zone to convert a Kotlin `LocalDateTime` to an `NSDate`. If the local datetime does not exist (`!components.isValidDateInCalendar(calendar)`, this happens during a spring-forward DST gap when clocks jump forward) - the code identifies it as a gap. In a gap scenario, `TimeZoneRulesFoundation` finds the next DST transition before the missing hour and constructs an `OffsetInfo.Gap` object, providing the transition instant (start of the gap), as well as the offsets immediately before and after the gap. On the other hand, if the `components` is valid date in calendar (meaning the local time exists), the code then checks for a potential fall-back DST overlap. It does so by comparing the UTC offset of that time with the offset 24 hours later. If the offset 24 hours later is smaller (indicating that clocks were turned back, creating an overlap of local times), the code further pinpoints the exact transition moment by adding the offset difference to the NSDate. If it confirms that the local time repeats (the offset after adding the difference is still less than the original), it returns an `OffsetInfo.Overlap` with the overlap’s start instant and the two different offsets before/after the overlap. If neither a gap nor overlap is detected, it returns a regular offset info (`OffsetInfo.Regular`). This careful calculation ensures that `TimeZoneRulesFoundation.infoAtDatetime` can correctly distinguish normal times from “missing” hours and “repeated” hours, paralleling the behavior of tzdb. All exceptions and edge cases are handled. In summary, on Darwin the library now uses the system’s time zone database via Foundation for all conversions, meaning it will recognize all valid zone IDs and compute offsets correctly without requiring the tzdb files. Additionally, the mechanism for retrieving available time zone IDs (`TimeZone.availableZoneIds` internally) was updated: on Darwin it falls back to using `NSTimeZone.knownTimeZoneNames` (through a new internal `getAvailableZoneIdsFoundation()`), ensuring the set of zone IDs is complete and up-to-date with the system’s zones.

## Testing and Verification

This PR includes an extensive test suite (`TimeZoneNativeTest`) to verify that the Foundation-based implementation behaves identically to the tzdb-based implementation across a wide range of scenarios. The tests create pairs of `TimeZone` objects for the same zone ID - one using the regular tzdb rules (`timeZoneById` / `TimeZoneRulesCommon`) and one using the Foundation rules (`timeZoneByIdFoundation` / `TimeZoneRulesFoundation`) - and compare their results. Key aspects tested are:

- **DST Transitions**: The tests check that offsets around DST boundaries match. For example, for `America/New_York` the suite verifies the spring-forward transition in 2025 (clocks jumping from 1:59:59 to 3:00:00 on Mar 9, 2025) and the fall-back transition (clocks repeating 1:00 - 1:59 on Nov 2, 2025). It asserts that before a DST change, both implementations give the same offset, and after the change they still agree, and that the offset actually changes as expected (ensuring both detect the transition). Similar checks are done for other zones like `Europe/London` (which has a gap at 1 AM when DST starts, and an overlap in fall) and `Australia/Sydney` (opposite hemisphere DST dates). The tests confirm that `OffsetInfo.Gap` and `OffsetInfo.Overlap` produced by `TimeZoneRulesFoundation` exactly match those from the tzdb rules in each case (including the start instant of the gap/overlap and the offsets before/after).
- **Stable Times (No Transition)**: The tests pick ordinary dates (e.g. a random summer day mid-year, or a week span with no DST changes) and verify that for multiple sample instants the offsets remain constant and equal between the two implementations. For instance, for New York in July 2025 (well outside any DST change), the offset is consistently UTC-04 for all samples, and Foundation and tzdb agree on every query. This ensures no divergence in normal scenarios.
- **Comprehensive Edge Cases**: A variety of historical and edge-case time zone scenarios are covered to ensure robustness. The test suite includes zones that have permanent offset changes or skipped dates:
    - **Asia/Pyongyang**: North Korea’s time zone shifted from UTC+9 to UTC+8:30 on August 14, 2015, then back to UTC+9 on May 5, 2018. The tests verify that a local time that fell in the gap when the clocks moved (e.g. 23:30 on May 4, 2018, which did not exist due to the jump) is handled as a gap, and that times before and after yield the correct offsets (and that Foundation and tzdb agree on those offsets).
    - **America/Caracas**: Venezuela used UTC-4:30 for several years and switched to UTC-4 on May 1, 2016 (eliminating the half-hour offset). The tests ensure that the overlap/transition at that date is recognized consistently.
    - **Pacific/Apia** (Samoa) and **Pacific/Kwajalein**: These zones famously skipped entire days when they moved across the International Date Line (Samoa skipped December 30, 2011, and Kwajalein skipped August 21, 1993, jumping a whole day ahead). The Foundation-based implementation is tested to handle such extreme gaps correctly. For example, it verifies that 2011-12-30 simply doesn’t exist in Apia’s local time line and is reported as a gap, with the transition to 2011-12-31 identified properly, matching tzdb results.
    - **Asia/Magadan**: Changed from UTC+12 to UTC+10 on October 26, 2014. Creating a 2-hour overlap.
- **API Consistency**: Beyond just offset calculations, the tests also verify higher-level API consistency. They compare the results of `TimeZone.atZone(LocalDateTime)` (producing a `ZonedDateTime`) and `TimeZone.atStartOfDay(LocalDate)` for Foundation vs tzdb implementations across many sample dates and zones. In every case, the resulting instants or zoned date-times are equal, showing that the overall `TimeZone` behavior is unchanged from the user’s perspective. 

All these tests passed with the final implementation, confirming that `TimeZoneRulesFoundation` produces identical results to the standard tzdb-based approach. The few discrepancies initially found (for example, in handling overlaps like Venezuela’s offset change) were fixed during development - the overlap detection logic was adjusted and now aligns with tzdb’s data. The extensive coverage of edge cases gives confidence that the Foundation-based strategy is robust and accurate.

## Outcome

With this PR, Darwin (macOS/iOS) platforms now can use the native system time zone data for all date-time conversions, solving the tzdb access issues #485. The `TimeZoneRules` abstraction cleanly separates platform-specific implementations. On Apple devices, the library will seamlessly fall back to Foundation’s `NSTimeZone` for zone offsets and transitions, while other platforms continue using the embedded tzdb as before - all behind the scenes. This change ensures that developers can rely on kotlinx-datetime on iOS/macOS without encountering missing zone IDs or crashes, and without any API changes. The solution is validated by comprehensive tests demonstrating that the new implementation’s behavior is indistinguishable from the previous one, even in tricky scenarios. In summary, this PR provides a robust fix for time zone handling on Darwin systems and maintains consistency of results across all platforms.


core/darwin/src/internal/TimeZoneRulesFoundation.kt
```diff
@@ -0,0 +1,72 @@
...
+
+internal class TimeZoneRulesFoundation(zoneId: String) : TimeZoneRules {
+    private val nsTimeZone: NSTimeZone = NSTimeZone.timeZoneWithName(zoneId)
+        ?: throw IllegalArgumentException("Unknown timezone: $zoneId")
+
+    override fun infoAtInstant(instant: Instant): UtcOffset =
+        infoAtNsDate(instant.toNSDate())
+
+    override fun infoAtDatetime(localDateTime: LocalDateTime): OffsetInfo {
+        val calendar = NSCalendar.currentCalendar().apply { timeZone = nsTimeZone }
```
dkhalanskyjb, line 29: Using `currentCalendar` will mean that the behavior will be different depending on the user's system configuration. This is an approach we are explicitly avoiding in `kotlinx-datetime`: for example, in java.time, the system timezone or locale is occasionally queried implicitly, and we don't do that.
Instead, this calendar can be used: https://developer.apple.com/documentation/foundation/nscalendar/identifier/iso8601

core/darwin/src/internal/TimeZoneRulesFoundation.kt
```diff
@@ -0,0 +1,72 @@
...
+        val components = localDateTime.toNSDateComponents()
+        val nsDate = requireNotNull(calendar.dateFromComponents(components)) {
```
dkhalanskyjb, line 31: When can this happen? In any case, it doesn't look like this could result from a user's error, so `IllegalArgumentException` doesn't seem suitable here. Instead, `IllegalStateException` (via `check`) seems better suited, and if we don't expect this to ever happen, adding more diagnostic information (`zoneId`, for one) could help us debug this if it does happen.

core/darwin/src/internal/TimeZoneRulesFoundation.kt
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+
+        return OffsetInfo.Regular(currentOffset)
+    }
+
+    @OptIn(UnsafeNumber::class)
+    private fun infoAtNsDate(nsDate: NSDate): UtcOffset {
+        val offsetSeconds = nsTimeZone.secondsFromGMTForDate(nsDate)
+        return UtcOffset.ofSeconds(offsetSeconds.toInt())
```
dkhalanskyjb, line 66: Minor: `ofSeconds` is an internal implementation detail of `UtcOffset`. Coupling can be reduced by using the user-visible `UtcOffset(seconds = offsetSeconds.convert())` form.

`.convert()` instead of `toInt()` demonstrates that we only want to deal with platform-dependent types, not actually lose some information on conversion.

core/darwin/src/internal/TimeZoneRulesFoundation.kt
```diff
@@ -0,0 +1,72 @@
...
+            val deltaSecondsLater = nsDate.addTimeInterval(delta.toDouble()) as NSDate
+            val deltaSecondsLaterOffset = infoAtNsDate(deltaSecondsLater)
+            if (currentOffset.totalSeconds > deltaSecondsLaterOffset.totalSeconds) {
+                return OffsetInfo.Overlap(
+                    start = nsTimeZone.nextDaylightSavingTimeTransitionAfterDate(nsDate)!!.toKotlinInstant(),
+                    offsetBefore = currentOffset,
+                    offsetAfter = deltaSecondsLaterOffset
+                )
+            }
+        }
```
dkhalanskyjb, line 58: Have you considered reusing the logic from `fun infoAtLocalDateTime(localDateTime: LocalDateTime, offsetAtYearStart: UtcOffset): OffsetInfo`? That is, find some point in time earlier than the given one (say, by a month), iterate through rules (in this case, using `nextDaylightSavingTimeTransitionAfterDate`), find the rules that were in effect at the given `localDateTime`). Looks more straightforward to me, but I may be missing some potential complications, and the implementation proposed here does seem correct at a glance.

core/darwin/src/internal/TimeZoneNative.kt
```diff
@@ -12,10 +12,18 @@ import kotlinx.datetime.internal.*
 import platform.Foundation.*
 
 internal actual fun timeZoneById(zoneId: String): TimeZone =
-    RegionTimeZone(tzdb.getOrThrow().rulesForId(zoneId), zoneId)
+    runCatching { RegionTimeZone(tzdb.getOrThrow().rulesForId(zoneId), zoneId) }
+        .getOrElse { timeZoneByIdFoundation(zoneId) }
```
dkhalanskyjb, line 16: With this implementations, the system can enter a weird state where one part of the data comes from the filesystem, whereas the other comes from the fallback. I think it makes more sense to only check once whether to use a filesystem or the fallback, and then consistently stick to that.

core/darwin/src/internal/TimeZoneNative.kt
```diff
@@ -12,10 +12,18 @@ import kotlinx.datetime.internal.*
...
+
+internal fun timeZoneByIdFoundation(zoneId: String): TimeZone =
+    RegionTimeZone(TimeZoneRulesFoundation(zoneId), zoneId)
 
 internal actual fun getAvailableZoneIds(): Set<String> =
-    tzdb.getOrThrow().availableTimeZoneIds()
+    runCatching { tzdb.getOrThrow().availableTimeZoneIds() }
+        .getOrElse { getAvailableZoneIdsFoundation() }
```
dkhalanskyjb, line 23: Can `availableTimeZoneIds` even reasonably fail? I think it will just return broken/incomplete data, even if the fallback contains some timezone identifiers. One more reason to consistently stick to either the filesystem or the fallback.

core/darwin/src/internal/TimeZoneRulesFoundation.kt
dkhalanskyjb: I've taken a more careful look, and here are the cases when this code may return an incorrect result:
* If any transition (either a gap or an overlap) is followed by a gap in the next 24 hours, and the `LocalDateTime` lands in that gap, then the gap will be considered to have started at the moment that the earlier transition has, and the UTC offsets will be wrong.
* If an overlap is followed by at least some gap in the next 24 hours, the gap may be misclassified as `OffsetInfo.Regular`. In particular, if the gap is the same duration or larger than the overlap, the overlap will always be misclassified as `OffsetInfo.Regular`.
* If there are several overlaps in the span of 24 hours and the `LocalDateTime` falls at the start of an earlier overlap, the `OffsetInfo.Regular` before an earlier overlap can be misclassified as an overlap itself. Example: a one-hour overlap on 02:00 → 01:00, then one-hour overlap on 15:00 → 14:00; we're interested in 00:30. `delta` = 2 hours; `deltaSecondsLater` = the second occurrence of 01:30, `currentOffset > deltaSecondsLaterOffset`.

This means that in any scenario, if several transitions happen in the span of 24 hours, some information around that time will be wrong. I haven't seen this ever happening historically, but currently, our implementations are working even with these cases, and it would be nice to support this if it doesn't complicate the code.

dkhalanskyjb (changes requested):
It troubles me that this implementation strongly relies on a contract that I haven't found stated anywhere. Have I missed some note in the Apple docs?

DmitryNekrasov:
I've rewritten the `TimeZoneRulesFoundation.infoAtDatetime(LocalDateTime)` implementation so that it is now similar to `RecurringZoneRules.infoAtLocalDateTime()` and no longer speculates on how far apart the overlaps and gaps are in time.

core/darwin/test/TimeZoneNativeTest.kt
```diff
@@ -0,0 +1,433 @@
...
+
+        val firstOffset = regularTz.offsetAt(testPoints.first().toInstant(regularTz))
+        for (dateTime in testPoints.drop(1)) {
+            val offset = regularTz.offsetAt(dateTime.toInstant(regularTz))
+            assertEquals(
+                firstOffset,
+                offset,
+                "Offset should remain constant during stable summer period"
+            )
+        }
```
dkhalanskyjb, line 146:
```suggestion
        val summerTime = LocalDateTime(2025, 7, 15, 12, 30, 45)
        val zoneId = "America/New_York"

        val regularTz = timeZoneById(zoneId)
        val foundationTz = timeZoneByIdFoundation(zoneId)

        val summerTimeInstant = summerTime.toInstant(regularTz)
        val firstOffset = regularTz.offsetAt(summerTimeInstant)

        for (period in listOf(
            DateTimePeriod(),
            DateTimePeriod(hours = 1),
            DateTimePeriod(days = 1),
            DateTimePeriod(days = 7),
        )) {
            val instant = summerTimeInstant.plus(period, regularTz)
            val regularOffset = regularTz.offsetAt(instant)
            val foundationOffset = foundationTz.offsetAt(instant)

            assertEquals(
                regularOffset,
                foundationOffset,
                "Regular and Foundation implementations should have same offset during stable period at $instant"
            )
            assertEquals(
                firstOffset,
                regularOffset,
                "Regular implementation should have same offset as at summer time for period $period"
            )
        }
```
Looks like the same logic to me, but somewhat simpler.

dkhalanskyjb (changes requested):
The one significant problem here is that the fallback does not report `"UTC"` as an available timezone identifier. Other than that, all I have are small stylistic suggestions. Thanks!

core/darwin/test/TimeZoneNativeTest.kt
dkhalanskyjb: Here it is: https://github.com/Kotlin/kotlinx-datetime/blob/5a2a6e17fc357f4c8e09cfc1bdb668720cf38bfa/core/common/src/Instant.kt#L64

core/darwin/src/internal/TimeZoneRulesFoundation.kt
```diff
@@ -0,0 +1,76 @@
...
+
+            val yearOfNextDate = calendar.component(NSCalendarUnitYear.convert(), fromDate = transitionDateTime)
+            val transitionDateTimeInstant = transitionDateTime.toKotlinInstant()
+
+            val offsetBefore = infoAtNsDate(currentDate)
+            val ldtBefore = transitionDateTimeInstant.toLocalDateTime(offsetBefore)
+            val offsetAfter = infoAtNsDate(transitionDateTime)
+            val ldtAfter = transitionDateTimeInstant.toLocalDateTime(offsetAfter)
+
+            return if (localDateTime < ldtBefore && localDateTime < ldtAfter) {
```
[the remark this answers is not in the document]
dkhalanskyjb: Sure, fair point. Right now, nothing connects the two implementations in the code, though. Could you add a comment to both implementations, mentioning that if one is changed, the other should be changed, too? Then, we could indeed utilize the similarities.

dkhalanskyjb (approved):
Thanks, this looks good!

PR #551, opened 2025-07-06
Move a local function outside of the containing `inline` function

lunakoly:
Local functions inside inlines were never expected to work and were allowed in K2 by accident. Until they are explicitly allowed, they should be avoided.

^KT-77986

### DIFF
--- a/core/commonKotlin/src/internal/Tzfile.kt
+++ b/core/commonKotlin/src/internal/Tzfile.kt
@@ -88,6 +88,9 @@ internal fun readTzFile(data: ByteArray): TzFile {
         override fun toString(): String = "Ttinfo(utoff=$utoff, isdst=$isdst, abbrind=$abbrind)"
     }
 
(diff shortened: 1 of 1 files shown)


dkhalanskyjb (approved):
The change makes sense. `kotlin-community/k2/dev` doesn't need an approval from a maintainer, though. Instead, the release team is responsible for that branch.

qwwdfsad:
Aren't the changes supposed to go into `develop` and then merged into `kotlin-community/k2/dev` to avoid branches divergence?

FenstonSingel:
> Instead, the release team is responsible for that branch.

> Aren't the changes supposed to go into `develop` [first]?

FTR JiC: kotlinx-datetime is, in fact, the only kotlinx project that is not being maintained as a KUP by either the Release team or the Libraries team; that honor is mine. Unfortunately, it means two things: 1) I'm not extensively privy to the Release team's processes and commitments regarding kotlinx-as-KUPs (nor do I follow them to a tee as a result of that); 2) my current KF/KLE QA workload has pushed most of my KUP maintenance tasks to the backburner either way.

PR #554, opened 2025-07-07
Support new maven central publication scheme

fzhinkin:
It is not longer possible to publish artifacts to sonatype staging repositories and the central publishing portal has to be used instead.
The PR adds TC jobs publishing all the artifacts into intermediate maven repo and then collect all the artifacts into a single deployment uploaded to the portal.

### DIFF
--- a/.teamcity/CentralDeployment.kt
+++ b/.teamcity/CentralDeployment.kt
@@ -45,12 +45,6 @@ fun Project.startDeployment() = BuildType {
             "",
             display = ParameterDisplay.PROMPT
         )
-        select(
-            "Repository",
-            "central",
-            options = listOf("central", "sonatype"),
-            display = ParameterDisplay.PROMPT
-        )
         text(
             "ZoneInfoVersion",
             "",
@@ -81,7 +75,6 @@ fun Project.deployToCentral(startDeployment: BuildType) = buildType("DeployCentr
     params {
         param(versionSuffixParameter, "${startDeployment.depParamRefs["VersionSuffix"]}")
         param(releaseVersionParameter, "${startDeployment.depParamRefs["Version"]}")
-        param("system.publication_repository", "${startDeployment.depParamRefs["Repository"]}")
     }
 
     vcs {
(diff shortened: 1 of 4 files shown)


.teamcity/CentralDeployment.kt
```diff
@@ -0,0 +1,214 @@
...
+        when (platform) {
+            Platform.Linux -> {
+                addAll(
+                    listOf(
+                        "publishAndroidNativeArm32PublicationToCentralRepository",
+                        "publishAndroidNativeArm64PublicationToCentralRepository",
+                        "publishAndroidNativeX64PublicationToCentralRepository",
+                        "publishAndroidNativeX86PublicationToCentralRepository",
+                        "publishLinuxArm64PublicationToCentralRepository",
+                        "publishLinuxX64PublicationToCentralRepository"
```
dkhalanskyjb, line 108: Where do all these tasks come from? `/gradlew tasks | grep publish | grep Central` returns nothing for me. The old `Deploy` builds only have `clean publish`, and I do have `publish` in my list of tasks.

Also, if this conditional publication depending on the platform is an optimization here, then can't `kotlinx-datetime` be optimized to just `clean publish` on just `MacOS`?

fzhinkin: > Where do all these tasks come from? 

You have to add following Gradle properties to get those tasks:
`-Pteamcity=on -Plibs.repo.url=https://localhost -Plibs.repo.user=user -Plibs.repo.password=password`

> Also, if this conditional publication depending on the platform is an optimization here, then can't `kotlinx-datetime` be optimized to just `clean publish` on just `MacOS`?

Sure, let's do that!

fzhinkin: Done

.teamcity/CentralDeployment.kt
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    requirements {
+        doesNotMatch("teamcity.agent.jvm.os.name", "Windows")
+    }
+
+    dependsOnSnapshot(startDeployment)
+
+    triggers {
+        finishBuildTrigger {
```
ilya-g, line 137: I wonder why `finishBuildTrigger` is used here instead, i.e. why the first build in chain triggers subsequent ones instead of an aggregate build configuration running dependencies, like it was before?
Will it be possible not to publish TZ info for `-compat` versions, for example, with this setup?

fzhinkin:
Obsolete in favor of https://github.com/Kotlin/kotlinx-datetime/pull/619

PR #560, opened 2025-07-18
Backport kxDT-as-KUP infrastructure and migration changes

FenstonSingel:


FenstonSingel:
(corrected a faulty commit description)

buildSrc/src/main/kotlin/CommunityProjectsBuild.kt
```diff
@@ -1,50 +0,0 @@
...
-/**
- * The version of Kotlin compiler to use in the Kotlin aggregate build.
- */
-// unused, but may be useful in the future
-val Project.kotlinVersion: String
-    get() = if (rootProject.properties["build_snapshot_train"]?.toString()?.toBoolean() == true) {
-        rootProject.properties["kotlin_snapshot_version"] as? String ?: error("kotlin_snapshot_version must be specified")
-    } else {
-        rootProject.properties["defaultKotlinVersion"] as String
-    }
```
dkhalanskyjb, line 50: I like this file: it looks like it could some day become a Gradle plugin, a single one for all projects, reducing the maintenance load and providing a unified experience. For example, here's the corresponding file in `kotlinx.coroutines`: https://github.com/Kotlin/kotlinx.coroutines/blob/f4f519b36734238ec686dfaec1e174086691781e/buildSrc/src/main/kotlin/CommunityProjectsBuild.kt It's fairly similar, and one can compare them to see the discrepancies/omissions.

In fact, even if just for conveniently searching for discrepancies, this file should not only exist but encompass more of the logic: if `Xpartial-linkage-loglevel=error` truly is a KUP requirement, `kotlinx.coroutines` doesn't fulfill it currently, and it would have been easier to notice were this configuration part of this file.

Still, since you're the one who has to interact with this part of the project the most, if you think this file is unnecessary, to be it. I'm not really asking for this file to be restored but am simply sharing an opinion.

benchmarks/build.gradle.kts
```diff
@@ -1,3 +1,5 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask
+
 /*
```
dkhalanskyjb, line 3: Please move the `import` below the copyright header so that it stays a header.

dkhalanskyjb (approved):

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,8 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionSha256Sum=ff7bf6a86f09b9b2c40bb8f48b25fc19cf2b2664fd1d220cd7ab833ec758d0d7
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.0.2-bin.zip
+distributionSha256Sum=9d926787066a081739e8200858338b4a69e837c3a821a33aca9db09dd4a41026
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.5-bin.zip
 networkTimeout=10000
+validateDistributionUrl=true
```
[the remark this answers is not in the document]
FenstonSingel: AFAIR, this line was added automatically by `./gradlew wrapper --gradle-version 8.5.
I decided that it knww better than me. 😅

buildSrc/src/main/kotlin/CommunityProjectsBuild.kt
FenstonSingel: Done. PTAL.

PR #562, opened 2025-07-29
Describe the release procedures

### LINKED ISSUE #524: Implement and document a procedure for publishing timezone database information separately from the main release
The Wasm/WASI implementation of `kotlinx-datetime` uses the timezone database supplied by us (https://github.com/Kotlin/kotlinx-datetime?tab=readme-ov-file#note-about-time-zones-in-wasmwasi), as the platform itself does not expose the necessary data.

Someone may need to use newer versions of the timezone database while staying on the old version of `kotlinx-datetime`: for example, new versions may introduce unwanted behavior or remove the deprecated APIs, and the adaptation of the client code to the new version may be postponed. To support this, we need to be able to publish just the timezone database artifacts separately from the releases of the library itself.

dkhalanskyjb:
Depends on #554
Fixes #524 (but does not allow updating the Windows name mappings without releasing the whole library).

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,220 @@
+# kotlinx-datetime release checklist
+
+## Full release
+
+To release `kotlinx-datetime`:
+
+1. Make sure there are no updates to either Windows/IANA timezone name mapping
+   or to the IANA timezone database that we publish.
+   * <https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxDatetime_Check_Updates>
+     checks this automatically every night.
+   * Follow the procedure in
+     [UPDATE_TIMEZONE_DATABASE.md](UPDATE_TIMEZONE_DATABASE.md)
+     if there are updates or to check for updates manually.
+
+2. Check out the `master` branch:
+   > `git checkout master`
+
+3. Retrieve the most recent version of it:
+   > `git pull`
+
+4. Search & replace `<old-version>` with `<version>` in
+   * Documentation: [README.md](README.md)
+   * [gradle.properties](gradle.properties)
+   * **NOT** [CHANGES.md](CHANGES.md)!
+
+5. Write release notes in [CHANGES.md](CHANGES.md):
+   * Follow the established style from the earlier release notes.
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -0,0 +1,220 @@
...
+   > `git diff version-<version>-compat version-<version>-normal`
+
+16. In TeamCity, start deployment of `version-<version>-compat` by running the
+   `Deployment/Deploy [RUN THIS ONE]` configuration:
+   <https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxDatetime_Deploy_All>.
+   Use the `Run custom build` button.
+   - In the `Changes` tab, select the build branch `version-<version>-compat`.
+   - In the `Parameters` tab, set the parameters:
+     * `Version` to `<version>-0.6.x-compat`.
+     * `Artifacts to publish` to `zoneinfo only`.
```
fzhinkin: Should it be `all` here?

fzhinkin: Or `core only`, if zone info artifacts are not expected for compat?

dkhalanskyjb: `core only`, yes, good find.

RELEASE.md
```diff
@@ -0,0 +1,226 @@
...
+When we no longer have the compatibility artifact, we'll remove the subsection.
+
+9. Create a new branch from `version-<version>`:
+   > `git checkout -b version-<version>-compat version-<version>`
+
+10. Replace `<version>` with `<version>-0.6.x-compat`
+   in [gradle.properties](gradle.properties).
+
+11. Commit and push the changes:
+   > `git commit -a -m 'Version <version>, compatibility artifact`
```
fzhinkin:
```suggestion
   > `git commit -a -m 'Version <version>, compatibility artifact'`
```

Also, GH's markdown renderer collapses this and the following line into a single line (it applies to all snippets with `>` in consecutive lines):

Image

Perhaps, adding additional `>` in between or wrapping both lines into a single "\`\`\`sh ... \`\`\`"  code block will do a trick.

fzhinkin: Otherwise, it all looks good to me.

PR #577, opened 2025-10-20
Deprecate TimeZone serialization

### LINKED ISSUE #576: Remove `TimeZone` and `FixedOffsetTimeZone` serialization
It looks like having `TimeZone` and `FixedOffsetTimeZone` be `@Serializable` is more trouble than it's worth, so we plan to deprecate it.

Currently, they get serialized by simply taking their `id: String`. This approach is not fully reliable.
- New timezone identifiers occasionally get introduced to the timezone database (and sometimes, removed). If some system serializes a new timezone identifier and sends it over to a system that doesn't have an up-to-date timezone database, this will result in an error during deserialization.
- Even if both systems know a specific timezone identifier, they don't necessarily associate it with the same timezone rules. This means that an object serialized on one system will behave differently on another one once deserialized.

Providing the `Serializable` implementation for `TimeZone` and `FixedOffsetTimeZone` hides these issues behind an innocent-looking interface. We would like to have client code implicitly manipulate the identifier strings instead. This is less convenient, but it shifts the place where problems occur from the hard-to-tweak serialization internals to fully controlled client code, which allows flexibly reacting to the problem.

An elegant alternative would be to add a `@Serializable` timezone identifier as a separate entity to separate identifier syntax validation from querying the timezone database. Despite being entirely reasonable, this approach looks seems non-viable, as the naming rules of timezone identifiers are too lenient to allow us to implement useful validity checks. The closest to a timezone identifier grammar that someone got is https://tc39.es/proposal-temporal/#prod-TimeZoneIANAName, and it allows clearly broken identifiers like `dkhalanskyjb`, `nice_orange_chicken.2009`, and `Yes/no+maybe`.

dkhalanskyjb:
* Remove the `@kotlinx.serialization.Serializable` annotation from `TimeZone` and `FixedOffsetTimeZone`.
* Deprecate `TimeZoneSerializer` and `FixedOffsetTimeZoneSerializer` with a warning.
* Introduce and deprecate the `.serializer()` functions in their companion objects with a warning. Previously, this function used to be generated by the `kotlinx.serialization` compiler plugin.

Fixes #576

### DIFF
--- a/core/common/src/TimeZone.kt
+++ b/core/common/src/TimeZone.kt
@@ -167,7 +167,7 @@ public expect open class TimeZone {
         @Deprecated(
             "Serializing TimeZone is discouraged, " +
                     "as deserialization can fail depending on the configuration. " +
-                    "Please serialize the id String instead.",
+                    "Please serialize the string id instead.",
             level = DeprecationLevel.WARNING,
         )
         public fun serializer(): kotlinx.serialization.KSerializer<TimeZone>
@@ -267,7 +267,7 @@ public expect class FixedOffsetTimeZone : TimeZone {
         @Deprecated(
             "Serializing FixedOffsetTimeZone is discouraged, " +
                     "as deserialization can fail or return a non-fixed-offset zone depending on the configuration. " +
-                    "Please serialize the id String instead.",
+                    "Please serialize the string id instead.",
             level = DeprecationLevel.WARNING,
         )
         public fun serializer(): kotlinx.serialization.KSerializer<FixedOffsetTimeZone>
--- a/core/common/src/TimeZone.kt
+++ b/core/common/src/TimeZone.kt
@@ -33,7 +33,6 @@ import kotlin.time.Instant
  *
  * @sample kotlinx.datetime.test.samples.TimeZoneSamples.usage
  */
-@Serializable(with = TimeZoneSerializer::class)
 public expect open class TimeZone {
     /**
      * Returns the identifier string of the time zone.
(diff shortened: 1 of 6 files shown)


ilya-g (approved):

core/common/src/serializers/TimeZoneSerializers.kt
```diff
@@ -16,6 +15,10 @@ import kotlinx.serialization.encoding.*
  *
  * JSON example: `"Europe/Berlin"`
  */
+@Deprecated(
+    "Serializing TimeZone is discouraged. Please serialize the id String instead.",
```
sandwwraith: `id String` sounds weird. Maybe you want `id: String`? Or since this is more of a text message, `string id` sounds better IMO

core/common/src/TimeZone.kt
```diff
@@ -33,7 +33,6 @@ import kotlin.time.Instant
  *
  * @sample kotlinx.datetime.test.samples.TimeZoneSamples.usage
  */
-@Serializable(with = TimeZoneSerializer::class)
 public expect open class TimeZone {
```
sandwwraith: Note that the moment you remove `@Serializable(with=...)` annotation from type, it is shown as red with SERIALIZER_NOT_FOUND where it was used in any `@Serializable` user class. Sadly, we do not have any kind of deprecation migration for that; so I suggest raising all `@Deprecated` here to ERROR just to speedup this process

core/common/src/serializers/TimeZoneSerializers.kt
```diff
@@ -1,12 +1,11 @@
 /*
- * Copyright 2019-2021 JetBrains s.r.o.
+ * Copyright 2025 JetBrains s.r.o.
```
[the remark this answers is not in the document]
dkhalanskyjb: We can remove the copyright notices completely: https://github.com/Kotlin/kotlinx.coroutines/pull/4016

core/common/src/TimeZone.kt
sandwwraith: I think most of the people wouldn't be able to guess that red code with deprecation forces them to use `@Serializable(TimeZoneSerializer::class)` (also with deprecation). But let's try, there's hardly another approach if you do not want to skip WARNING stage.

sandwwraith: Although I would still advocate going directly to ERROR as kotlinx-datetime is still 0.x and ~~we are breaking binary compatibility anyway (missing @Serializable would lead to linkage errors, as we try to access `TimeZone.Companion.serializer()` which would be gone)~~ (upd: I saw you addressed this)

sandwwraith (approved):

PR #581, opened 2025-10-22
Enable Gradle build caches

dkhalanskyjb:
Proposed by @adam-enko


fzhinkin:
Does it actually improve build times (or anything else)?
If there are no arguments why enabling it could be detrimental, then there should be nothing stopping us from enable it :)

PR #583, opened 2025-10-29
On the mingw64 target, use a fixed-offset timezone name if the user disabled DST transitions

DmitryNekrasov:
This PR aligns our behavior with Java's approach: when DST is disabled, return a proper fixed-offset timezone with an appropriate name instead of a region timezone.

### Changes

1. Extracted `asTimeZone(prefix)` utility function 
2. Refactored timezone ID parsing: uses the new `asTimeZone(prefix)` utility for consistent fixed-offset timezone creation
3. Updated Windows DST handling
    - Extracted `currentSystemDefaultFromDtzi` helper function for testability
    - Changed behavior: when `DynamicDaylightTimeDisabled` is set, returns `UtcOffset.asTimeZone("GMT")` instead of a `RegionTimeZone` 
4. Added comprehensive test
    - Verifies DST enabled → `RegionTimeZone`
    - Verifies DST disabled → `FixedOffsetTimeZone` with `GMT` prefix

### Impact

This is a behavior change for Windows users who have disabled DST. Instead of receiving a `RegionTimeZone` with a confusing name, they will receive a properly-typed `FixedOffsetTimeZone` with a name like `GMT+01:00`.

### DIFF
--- a/core/windows/test/TimeZoneRulesCompleteTest.kt
+++ b/core/windows/test/TimeZoneRulesCompleteTest.kt
@@ -19,7 +19,6 @@ import kotlin.time.Clock
 
 class TimeZoneRulesCompleteTest {
 
-    @OptIn(ExperimentalStdlibApi::class)
     @Test
     fun dynamicDaylightTimeDisabled() {
         memScoped {


core/windows/test/TimeZoneRulesCompleteTest.kt
```diff
@@ -19,6 +19,27 @@ import kotlin.time.Clock
 
 class TimeZoneRulesCompleteTest {
 
+    @OptIn(ExperimentalStdlibApi::class)
```
dkhalanskyjb: Does this line have an effect?

PR #587, opened 2025-11-10
Add contribution guidelines

dkhalanskyjb:


fzhinkin:
@dkhalanskyjb, WDYT about https://github.com/Kotlin/kotlinx-datetime/pull/587#issuecomment-3513835994?

dkhalanskyjb:
Honestly, nothing comes to mind. `kotlinx.coroutines` has some opaque rules (naming conventions for tests, using `develop` as the target most of the time but not always, etc.), but `kotlinx-datetime` is very lenient. Tests are named reasonably but without a strong convention, the default branch is always the PR target, the style is also mostly just https://kotlinlang.org/docs/coding-conventions.html.

We could try listing even these aspects, but I'd prefer to avoid adding text with obvious information, to decrease the risk that the contributors get intimidated and skip reading the guidelines entirely. If we witness a contributor making a mistake, we can always update the guidelines later in reaction to that.

fzhinkin (approved):

PR #590, opened 2025-12-02
Document the parser concatenation function

dkhalanskyjb:


dkhalanskyjb:
Ok, then please merge your PR when you're ready, and we'll put this documentation as a separate commit in `master` (since it's not a trivial change).

PR #610, opened 2026-02-26
Introduce APIs for dependency injection of TimeZone-related information

### LINKED ISSUE #17: TimeZone.SYSTEM encourages static dependency

### LINKED ISSUE #201: Provide the means for a non-default timezone database
When developing for Android, developers can use desugaring to enable `java.time` APIs that will be used for kotlinx-datetime implementation on JVM. However, Android doesn't allow developers to provide their own time zone database to be used with those APIs. On APIs 26-28 this means devices will only use time zones provided with the system update (so mostly outdated). On APIs 29+, time zone updates are provided via project mainline, which is supposed to happen regardless of OEM support. However, project mainline depends on Google Play Services infrastructure, which a lot of devices (like Huawei or custom ROMs) don't have. That prevents developers who want to use most recent TZ database from using java.time APIs, and right now that includes `kotlinx-datetime`.

For that reason, it'd be great to be able to use ThreeTenBP backend with kotlinx-datetime on JVM, ensuring that we can provide most recent time zone database with the app, instead of relying on OEMs or the devices having Google services installed.

See also:
 - https://source.android.com/devices/architecture/modular-system/timezone
 - https://issuetracker.google.com/issues/159421054
 - https://www.zacsweers.dev/ticktock-desugaring-timezones/

### LINKED ISSUE #435: Consider a new name for `TimeZone.of`
It doesn't seem idiomatic. `of` is just visual noise that doesn't tell anything about what argument to expect. It's unclear right now what it should be called instead: it's a bit too effectful for a constructor function (as it queries the timezone database), it's far from a pure computation.

dkhalanskyjb:
Fixes #17
Fixes #201
Fixes #435

**https://github.com/Kotlin/kotlinx-datetime/discussions/611 is a space for comments, questions, and suggestions.** Since the PR is at the conceptual stage, some discussion threads may end up being quite long, so the PR format is not particularly well-suited for it.

The most interesting places in the code are:
* https://github.com/Kotlin/kotlinx-datetime/blob/dkhalanskyjb/deprecate-datetime-singletons/core/common/src/TimeZoneContext.kt the new APIs.
* https://github.com/Kotlin/kotlinx-datetime/blob/dkhalanskyjb/deprecate-datetime-singletons/core/common/src/TimeZone.kt how `TimeZone` will look like with its static functions now deprecated. Note the updated semantics of `==`, which now takes the `TimeZone`'s origin timezone databases into account.
* https://github.com/Kotlin/kotlinx-datetime/blob/dkhalanskyjb/deprecate-datetime-singletons/timezones/full/common/src/BundledTimeZoneContext.kt the API of a separate artifact published alongside `kotlinx-datetime` one will be able to depend on to obtain an up-to-date timezone database.

It's important to note that this is *not* a full-fledged solution for providing custom timezone databases, though! Constructing a custom `TimeZone` is still going to be non-viable, even with the proposed API, as there's no way to construct a custom `TimeZone`. To properly implement that, we would (most likely) need to expose the structure of a `TimeZone` object (historical data, recurring rules) and the additional logic that is likely going to be needed when implementing realistic timezone databases (see https://github.com/Kotlin/kotlinx-datetime/blob/e5505568c51b7165f683d740d3789d1c505a3957/timezones/full/common/src/BundledTimeZoneContext.kt#L84-L96). We wanted to start small for now and watch the community response to at least the dependency injection API as a whole. Then, we would like to tweak it if necessary, and only then collect the less common use cases and expose the rest of the pieces.


swankjesse (commented):
This is fantastic and I appreciate how much design effort went into this.

When I set up servers and CI environments and such I always set an environment variable `TZ=UTC0` so the host machine’s zone doesn’t accidentally leak through to the application layer. For example, I don’t need to worry that payment scheduling will change due to DST.

Would you consider exposing a symbol similar to `TimeZoneContext.System` like `TimeZoneContext.Universal` that is locked to UTC?

dkhalanskyjb:
`TimeZoneContext.Universal` looks very specific, so probably not in the form you propose. My guess is that the API would look something like `TimeZoneContext.System.withFixedCurrentTimeZone(TimeZone.UTC)`, though we'd need to carefully study the use cases for custom `TimeZoneContext`s to arrive at a good orthogonal API.

core/api/kotlinx-datetime.api
```diff
@@ -680,6 +680,42 @@ public final class kotlinx/datetime/TimeZone$Companion {
...
 
+public abstract interface class kotlinx/datetime/TimeZoneContext : kotlinx/datetime/TimeZoneDatabase, kotlinx/datetime/TimeZoneIdProvider {
+	public static final field Companion Lkotlinx/datetime/TimeZoneContext$Companion;
+	public fun currentTimeZone ()Lkotlinx/datetime/TimeZone;
+}
+
+public final class kotlinx/datetime/TimeZoneContext$Companion {
+}
+
+public final class kotlinx/datetime/TimeZoneContext$DefaultImpls {
```
JakeWharton: Since this is a new API, can it use the new mechanism of default function compilation which relies on JVM default methods? This old trampoline-based version is only needed for existing APIs targeted by old code.

dkhalanskyjb: Good call, thanks!

core/common/src/TimeZoneContext.kt
```diff
@@ -0,0 +1,257 @@
...
+    /**
+     * The system's timezone context.
+     *
+     * It is encouraged to avoid direct calls to [System] methods in your code
+     * in favor of dependency injection of a [TimeZoneContext] instance,
+     * or if only a part of that functionality is needed,
+     * of either a [TimeZoneDatabase] or a [TimeZoneIdProvider] instance.
+     * When writing library code, this is a strict requirement.
+     */
+    public object System: TimeZoneContext {
```
qwwdfsad, line 109:
```suggestion
    public object System : TimeZoneContext {
```

core/common/src/TimeZoneContext.kt
```diff
@@ -0,0 +1,257 @@
...
+         * @throws RuntimeException if a system misconfiguration prevents obtaining
+         * the default system time zone.
+         *
+         * @sample kotlinx.datetime.test.samples.TimeZoneSamples.currentSystemDefault
+         */
+        override fun currentTimeZone(): TimeZone = currentSystemDefaultTimeZone()
+    }
+
+    /** @suppress */
+    public companion object
```
qwwdfsad, line 257: Is it left here by accident, or is it a place to put extensions on?

core/jvm/src/TimeZoneJvm.kt
```diff
@@ -114,35 +126,87 @@ internal constructor(public actual val offset: UtcOffset, zoneId: ZoneId): TimeZ
...
+            java.time.ZonedDateTime.ofLocal(
+                dateTime.value, actualZoneId, preferred?.zoneOffset
+            ).toInstant().toKotlinInstant()
+
+        override fun equals(other: Any?): Boolean = other is ActualZoneId && actualZoneId == other.actualZoneId
+        override fun hashCode(): Int = actualZoneId.hashCode()
+        override fun toString(): String = actualZoneId.toString()
+    }
+
+    class RuleBasedZoneId(val zoneRules: RuleBasedTimeZoneCalculations): ZoneIdLike {
```
qwwdfsad:
```suggestion
    class RuleBasedZoneId(private val zoneRules: RuleBasedTimeZoneCalculations): ZoneIdLike {
```

core/common/src/internal/tz/FixedOffsetTimeZoneDatabase.kt
```diff
@@ -0,0 +1,75 @@
...
+        } else {
+            onParsed(parsedOffset.asTimeZone())
+        }
+    }
+    return when {
+        id == "UTC" -> onParsed(TimeZone.UTC)
+        id == "Z" || id == "z" -> onParsed(UtcOffset.ZERO.asTimeZone())
+        id.length == 1 -> null
+        id.startsWith("+") || id.startsWith("-") -> parse(id, null)
+        id == "UTC" || id == "GMT" || id == "UT" -> onParsed(FixedOffsetTimeZone.withSpecificName(UtcOffset.ZERO, id))
```
qwwdfsad: `UTC` is handled a few lines above

timezones/full/jvm/java9/module-info.java
```diff
@@ -0,0 +1,6 @@
+module kotlinx.datetime.zoneinfo {
+    requires kotlin.stdlib;
+    requires kotlinx.datetime;
```
qwwdfsad, line 3: Shouldn't it be `requires transitive` to export timezones themselves as well?

core/jvm/src/TimeZoneJvm.kt
```diff
@@ -114,35 +126,87 @@ internal constructor(public actual val offset: UtcOffset, zoneId: ZoneId): TimeZ
...
+        override fun atStartOfDay(date: LocalDate): Instant =
+            zoneRules.atStartOfDay(date)
+
+        override fun localDateTimeToInstant(dateTime: LocalDateTime, preferred: UtcOffset?): Instant =
+            zoneRules.localDateTimeToInstant(dateTime, preferred)
+
+        override fun instantToLocalDateTime(instant: Instant): LocalDateTime =
+            instant.toLocalDateTime(offsetAt(instant))
+
+        override fun equals(other: Any?): Boolean = other is RuleBasedZoneId && zoneRules.id == other.zoneRules.id
```
qwwdfsad: Equals compares `id`s only, but `hashCode` is computed over `zoneRules`, which also includes `origin`.

Not sure if it is possible to trigger though, but that's a contract violation

core/jvm/src/Converters.kt
```diff
@@ -66,8 +68,16 @@ public fun java.time.Period.toKotlinDatePeriod(): DatePeriod = DatePeriod(this.y
 
 /**
  * Converts this [kotlinx.datetime.TimeZone][TimeZone] value to a [java.time.ZoneId][java.time.ZoneId] value.
- */
-public fun TimeZone.toJavaZoneId(): java.time.ZoneId = this.zoneId
+ *
+ * If the [TimeZone] value was obtained from a timezone database other than the system one,
+ * a direct conversion is not possible.
+ * In that case, the [ZoneId.of] will be queried with the timezone identifier of the [TimeZone]
+ * and may throw an [IllegalArgumentException] if the timezone identifier is invalid or not supported.
```
qwwdfsad: `ZoneId.of` throws something else (`ZoneRulesException`/`DTE`)

core/jvm/src/Converters.kt
```diff
@@ -66,8 +68,16 @@ public fun java.time.Period.toKotlinDatePeriod(): DatePeriod = DatePeriod(this.y
...
+ */
+public fun TimeZone.toJavaZoneId(): java.time.ZoneId = when (this.zoneId) {
+    is ZoneIdLike.ActualZoneId -> this.zoneId.actualZoneId
+    is ZoneIdLike.RuleBasedZoneId -> java.time.ZoneId.of(this.id)
```
qwwdfsad, line 79: On a second look, I am not sure about the semantics/use-case.

If I get a timezone from `BundledTimeZoneContext`, invoke `toJavaZoneId`, and it resolves, but with different offsets -- would it be the expected behaviour? 
Taking the subtlety of the conversion, maybe straightforward rejection would be a better option?

core/common/src/TimeZoneContext.kt
```diff
@@ -0,0 +1,257 @@
...
+/**
+ * Provides the identifier of the current time zone.
+ *
+ * This interface only provides a single method called [currentTimeZoneId].
+ * The meaning of the returned identifier is intentionally unspecified and application-specific.
+ *
+ * @see TimeZoneContext.System for the implementation where the system timezone configured on the host machine
+ * is taken as the current one.
+ */
+public interface TimeZoneIdProvider {
```
qwwdfsad, line 66: This one probably should be `fun`

core/common/src/TimeZoneContext.kt
qwwdfsad: (in many other places as well. ktfmt should address that en masse, feel free to ignore)

core/tzdbOnFilesystem/src/internal/TzdbOnFilesystem.kt
```diff
@@ -5,25 +5,40 @@
...
+        if (idAsPath.components.any { it == ".." }) {
+            throw IllegalTimeZoneException("'$idAsPath' must not contain '..' as a component")
+        }
+        val file = Path(tzdbPath.isAbsolute, tzdbPath.components + idAsPath.components)
+        val contents = file.readBytes() ?: throw IllegalTimeZoneException("File '$file' not found")
+        return readTzFile(contents).toTimeZoneRules()
+    }
+
+    override fun rulesForIdOrNull(id: String): TimeZoneRulesCommon? {
         val idAsPath = Path.fromString(id)
```
qwwdfsad, line 29: Looks a bit suspicious -- `rulesForId` checks for `if (id.length <= 1)` and throws, but here we don't do it.

Something like `"."` will reach `readBytes` and probably (?) will fail.
Just eye-balling, haven't tried to actually run it

timezones/full/common/src/BundledTimeZoneContext.kt
```diff
@@ -0,0 +1,106 @@
...
+     */
+    override fun get(id: String): TimeZone = impl.get(id)
+
+    /**
+     * Returns the time zone identified by the provided [id], or `null` if it's not found.
+     *
+     * This is optimized shorthand for `try { get(id) } catch (_: IllegalTimeZoneException) { null }`.
+     * Please see the documentation for [get].
+     *
+     * It is guaranteed that this function never throws an exception for [BundledTimeZoneContext].
```
qwwdfsad, line 53: Never throws or never throws `IllegalTimeZoneException`?
If the former, then I'd suggest a catch-all

core/common/src/TimeZoneContext.kt
```diff
@@ -0,0 +1,257 @@
...
+public interface TimeZoneDatabase {
+    /**
+     * Returns the time zone identified by the provided [id].
+     *
+     * @throws IllegalTimeZoneException if [id] has an unsupported format
+     * or the time zone with the given [id] is not found.
+     * @throws IllegalStateException if the [id] corresponds to a time zone, but its internal state
+     * could not be obtained, e.g., because the timezone database is corrupted.
+     */
+    public fun get(id: String): TimeZone =
```
qwwdfsad, line 30: Slightly tempted to make it `operator fun`.
It will be the opposite of maps tho (`get: T` + `getOrNull: T?` vs `get: T?` + `getValue: T`)

core/common/src/TimeZoneContext.kt
```diff
@@ -0,0 +1,257 @@
...
+ *   the timezone database (the [TimeZoneDatabase] interface).
+ *
+ * In addition to combining these two interfaces,
+ * an additional [currentTimeZone] function is provided that
+ * is in most cases equivalent to `get(currentTimeZoneId())`,
+ * but can additionally return a useful time zone
+ * even when the [TimeZoneDatabase] does not recognize the timezone identifier.
+ */
+@MyJvmDefaultWithoutCompatibility
+public interface TimeZoneContext: TimeZoneDatabase, TimeZoneIdProvider {
```
qwwdfsad, line 89: About the naming (I've looked into the design doc) -- have you considered calling this entity a provider as well? Seems both less ambiguous, nicely explains the purpose and one less term to consider

buildSrc/src/main/kotlin/zoneInfosResourcesGenerator.kt
```diff
@@ -27,17 +28,25 @@ private fun generateByteArrayProperty(tzData: TzData, header: String, propertyNa
 }
 
 private class TzData(val data: ByteArray, val fullTzNames: MutableList<String>)
+
+private fun listDirRecursively(currentRoot: File, prefix: String? = null): Sequence<String> = sequence {
+    require(currentRoot.isDirectory)
```
qwwdfsad, line 33: I wanted to write that `isDirectory` is presumably checked in `listFiles()` but then noticed it's my favourite `java.io.File` API :sigh:

timezones/full/common/src/BundledTimeZoneContext.kt
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * A synonym for [BundledTimeZoneContext].
+ */
+public val TimeZoneContext.Companion.Bundled: BundledTimeZoneContext
+    get() = BundledTimeZoneContext
+
+/**
+ * A [TimeZoneContext] that uses the timezone database from the `kotlinx-datetime-zoneinfo` artifact.
+ *
+ * This [TimeZoneContext] is a drop-in replacement for [TimeZoneContext.System] for cases where
```
qwwdfsad, line 20: Taking it's a drop-in replacement, the implementation is a bit concerning -- any `get*` (and if we ended up using BTZC, we probably expect a lot of usages) reads from a resource/JAR on the filesystem (also: going to be slow on Android and will cause all kinds of ANRs), then parse byte array every time.

I would expect to have some kind of thread-safe read-through cache between BTZC and the underlying resource

timezones/full/build.gradle.kts
```diff
@@ -21,20 +21,25 @@ node {
     nodeProjectDir.set(layout.buildDirectory.dir("node"))
 }
 
-val tzdbVersion: String by rootProject.properties
+val tzdbVersion = rootProject.property("tzdbVersion") as String
```
qwwdfsad, line 25: Could you please configure Kover for this project as well?
I typically look at coverage reporst as part of review to see what I've missed

timezones/full/jvm/src/BundledTimeZoneContextJvm.kt
```diff
@@ -0,0 +1,45 @@
...
+package kotlinx.datetime.zoneinfo
+
+import java.io.InputStream
+
+internal actual fun zoneDataByNameOrNull(name: String): ByteArray? {
+    // Forbid reading unrelated resources to avoid traversing arbitrary resources +
+    // improve performance for clearly faulty timezone names
+    if (name !in kotlinx.datetime.timezones.tzData.timeZones)
+        return null
+    (BundledTimeZoneContext.javaClass.classLoader.tzResourceByName(name) ?: return null).use { resource ->
```
qwwdfsad, line 15: (context: https://github.com/Kotlin/kotlinx-datetime/pull/610/changes#r3623058277)

Especially if we want older Android devices to be one of the main consumers here, I'd expect that even with read-through cache, reading a lot of small files from the resources might be detrimental.

To workaround a similar issue, ThreeTenABP used a single Android asset: https://github.com/JakeWharton/ThreeTenABP#why-not-use-threetenbp

The discussion is [quite old, though](https://github.com/JakeWharton/ThreeTenABP/issues/1), and probably things have changed since then.

qwwdfsad: Potential options:

* Do nothing, not an issue because we don't care or because `getResourceAsStream` is fast enough
* Have a single cacheable resource. Read/parse once outside of Main thread, enjoy the rest
* Built-in asset
* Embed it as a `ByteArray` (See also: [Deconstructing OkHttp](https://youtu.be/dmOrYzS_AKM?t=1777), the trick with [IDNA mappings](https://github.com/lysine-dev/okhttp/blob/b2cbae722dee8222ba3897f4aa657011a0548ecb/okhttp-idna-mapping-table/src/main/kotlin/okhttp3/internal/idn/GenerateIdnaMappingTableCode.kt#L47))

core/common/src/TimeZoneContext.kt
```diff
@@ -0,0 +1,257 @@
...
+
+/**
+ * A mapping from time zone identifiers to [TimeZone] objects.
+ *
+ * The interface does not impose any restrictions on the identifiers.
+ * Typically, the identifiers follow the IANA time zone conventions (https://www.iana.org/time-zones):
+ * for example, `Europe/Berlin` or `America/Los_Angeles`.
+ */
+@MyJvmDefaultWithoutCompatibility
+public interface TimeZoneDatabase {
```
qwwdfsad, line 21: Taking this interface is not externally implementable. I suggest making it so as to avoid compatibility burden in the future

core/common/src/TimeZoneContext.kt
```diff
@@ -0,0 +1,257 @@
...
+ */
+@MyJvmDefaultWithoutCompatibility
+public interface TimeZoneContext: TimeZoneDatabase, TimeZoneIdProvider {
+    /**
+     * Get the current time zone.
+     *
+     * In most cases, this is equivalent to `get(currentTimeZoneId())`.
+     * However, it is designed to also work in scenarios
+     * when the timezone database does not recognize
+     * the timezone identifier returned by [currentTimeZoneId] if possible.
```
qwwdfsad, line 96: Could you please elaborate on what you mean here?
That it is allowed for an implementation to return the timezone for which `zone.id == currentTimeZoneId()` yet `get(currentTimeZoneId())` throws?

core/common/src/TimeZoneContext.kt
qwwdfsad: I also implicitly assume that versioning of the database is a topic of further discussion (even though it can be done without this interface being publicly implemnetable)

qwwdfsad (changes requested):

core/common/src/TimeZoneContext.kt
dkhalanskyjb: It's for extensions, yes.

timezones/full/jvm/java9/module-info.java
dkhalanskyjb: I thought about it, but then, you'd get `LocalDateTime` + `UtcOffset` + `DateTimePeriod` just by requiring `kotlinx.datetime.zoneinfo`. To me, that would look weird. I don't know what the idiomatic approach to this is, though. We're okay with such dependency transitivity in Gradle, but this is subtly different.

If someone has strong opinions, sure, let's make it `transitive`, I have only a weak opinion against it.

core/jvm/src/Converters.kt
dkhalanskyjb: We decided on this behavior earlier when discussing this function: https://github.com/Kotlin/kotlinx-datetime/blob/fa7a190b778bd0733cfa947f5686d94428d99a0d/core/darwin/src/Converters.kt#L55-L74 If we disagree with the current behavior, we should break that converter as well. (By the way, that converter can be improved by short-circuiting on time zones that wrap an `NSTimeZone`).

The deciding factor is: the use cases we expect for such converters simply don't have a better/less buggy solution. `ZoneId` can not be manually constructed, so any function accepting a `ZoneId` is stuck with the JVM timezone database. If you already must provide a `ZoneId`, what else can you do but `ZoneId.of(myTimeZone.id)`? Nothing—so you will write that, either aware or unaware of the possibility of getting different UTC offsets.

Also note the `to` in the name: it implies a conversion, and a potentially lossy one. If this were `asJavaZoneId`, this behavior would be a clear bug. The `to` in the name was chosen intentionally, to signify the weak guarantees we provide.

core/common/src/TimeZoneContext.kt
dkhalanskyjb: We considered this, as well as consistency with `Map`, and decided that `get` both looks clearer *and* gives us a better way to revert the decision: if there's a demand for this being an `operator`, we can easily accommodate that later. If we make it `operator` now, we can't easily undo that.

core/common/src/TimeZoneContext.kt
dkhalanskyjb: Do you mean, `TimeZoneProvider`? I'm not sure if we considered that exact identifier, but it has the same problems that caused us to reject many other options in our prolonged discussion of this topic.

The main criterion that most naming attempts didn't fulfill was the inability to distinguish between `TimeZoneDatabase` and this mysterious entity at a glance. If you see a `TimeZoneProvider` in your codebase—would you understand that it's *not* simply a timezone database but something with extra information on top of it?

`TimeZoneContext` is vague and weird, but it's clearly wider than just the timezone database—`Context` implies an ambiguous, wide scope. That was the main deciding factor when choosing the current name.

If there are any suggestions of a more specific name than `TimeZoneContext` that won't be easily mixed up with `TimeZoneDatabase`, we will be genuinely thrilled to adopt them! It was very unsatisfying to settle for the least worst option here.

timezones/full/common/src/BundledTimeZoneContext.kt
dkhalanskyjb: It's "never throws, unless the runtime environment decides to exercise its right to make an arbitrary line of code throw" (e.g., stack overflows, out-of-memory errors, etc.). That's a mouthful, and also generally implied. catch-all would break the expectation that VM-level errors propagate upwards untouched. Do you have a suggestion for better wording here?

core/tzdbOnFilesystem/src/internal/TzdbOnFilesystem.kt
dkhalanskyjb: Yep, `.` will fail in `readBytes`. In `rulesForIdOrNull`, we don't return error messages, so we are only interested in security and receiving `null` for erroneous inputs.

core/common/src/TimeZoneContext.kt
dkhalanskyjb: The interface is intended to be fully externally implementable eventually, once we gather some feedback on the implementation and collect more requirements for custom timezone databases. It's unclear so far in which form we should expose the fixed-offset-timezone-database wrapper, or delegation to other timezone databases in general, etc., but once we figure that out, `TimeZoneDatabase` is 

Even right now, implementing `TimeZoneDatabase` can already be useful:

```kotlin
val TimeZoneDatabaseWithoutDstInBerlin = object: TimeZoneDatabase {
    override fun getOrNull(id: String): TimeZone? =
        if (id == "Europe/Berlin") TimeZoneContext.System.get("UTC+01:00")
        else TimeZoneContext.System.get(id)

    override fun availableZoneIds(): Set<String> = TimeZoneContext.System.availableZoneIds()
}
```

This way, you can test how your code reacts to expected changes to timezone rules.

As for versioning, I'm not sure what you are referring to. Not all timezone databases have a clear version—only IANA databases do. For example, AFAIK, the Windows registry is not versioned. Fully custom timezone databases may also lack a meaningful notion of a version. If we decide to expose IANA timezone databases as a separate entity, yes, it will make sense to require providing their versions in all implementations and to implement a mechanism for combining IANA timezone databases with partial information. Right now, it's not obvious that anyone would be interested in that.

core/common/src/TimeZoneContext.kt
dkhalanskyjb: Not necessarily. For example, on JS, `currentTimeZoneId()` is the actual time zone id, but `currentTimeZone()` is `SYSTEM` when no timezone database is provided.

core/common/src/TimeZoneContext.kt
dkhalanskyjb: Most single-function interfaces shouldn't be `fun`. `fun interface` is great for glorified lambdas or for callbacks whose contract needs to be specified in some central location. `TimeZoneIdProvider` doesn't fall into these buckets, and I believe it in particular isn't `fun` for the same reasons `Clock` isn't: https://github.com/Kotlin/kotlinx-datetime/issues/273#issuecomment-4176261589

core/common/src/TimeZoneContext.kt
dkhalanskyjb: This is a weird stylistic inconsistency between functions and classes out of nowhere. Or should `fun f(): Int` also become `fun f() : Int`? In any case, yes, the entire `kotlinx-datetime` codebase is in need of reformatting, and I'd like to avoid hunting down all extra/missing whitespace characters by hand, it just doesn't seem worth it.

timezones/full/build.gradle.kts
dkhalanskyjb: Done.

timezones/full/common/src/BundledTimeZoneContext.kt
dkhalanskyjb: You are right that for repeated calls, the performance of the bundled timezone database is clearly worse than that of `TimeZoneContext.System` on the JVM (on the other targets, I'm assuming it should be on par). The initial call for `System` and `Bundled` is in the same ballpark (and so the ANRs are equally possible in `System` and `Bundled` at the app start).

Our general recommendation is to create `TimeZone` objects once and then keep a reference to them, though. If people are generally doing that in mature codebases, we can expect any cache that we introduce not to get too many hits. That would make our internal caching for `Bundled` strictly detrimental: it is not free, as it requires some upkeep on every call, and it can't hope to match the tailor-made caching that applications themselves introduce.

So, it's not obvious to me that caching is the right call here. We can't make it unbounded—that's too much data—so we'd need to introduce cache eviction, which would cause opaque performance spikes, require adding control knobs like system properties, etc., and still be at risk of being useless anyway.

If we measure the impact on some real-world codebases and determine that caching would be an improvement, I'll be happy to implement/accept a PR with caching. If your intuition tells you it's so obvious we don't even need measurements, I'm ready to listen, but so far, you haven't said that.

timezones/full/jvm/src/BundledTimeZoneContextJvm.kt
dkhalanskyjb: Since it's possible to have a separate Android source set, we could implement a separate mechanism that is more efficient on Android. I agree this is very much worth investigating. Thank you for linking to several relevant resources and presenting a list of alternatives!

I'd like to argue that this is out of scope of this pull request. Indeed, 10+ years is a long time. Who knows—maybe Android development tooling got better at handling resources, and now they are just as efficient as the other methods? I have no idea. We'd need to ask an expert or benchmark the different approaches, but this can wait: all of this can be hidden behind the exact API this PR introduces. Let's stabilize the API first and research the performance improvements later.

core/common/src/TimeZoneContext.kt
qwwdfsad: [opt] Don't remember if we have a consistent approach to that, maybe worth leaving a KDoc?

timezones/full/jvm/java9/module-info.java
ilya-g: I would rather not add "transitive" to make clients that use kotlinx.datetime with bundled zoneinfo require both modules instead of just the latter.

core/common/src/internal/tz/FixedOffsetTimeZoneDatabase.kt
```diff
@@ -0,0 +1,75 @@
...
+ * Copyright 2019-2025 JetBrains s.r.o. and contributors.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.datetime.internal
+
+import kotlinx.datetime.*
+import kotlinx.datetime.format.*
+
+internal class FixedOffsetTimeZoneDatabase(val inner: TimeZoneDatabase): TimeZoneDatabase {
```
ilya-g: Maybe name it `TimeZoneDatabaseWithAddedFixedOffsets` or `FixedOffsetTimeZoneDatabaseWrapper`? Because the current name makes an impression that it somehow converts all timezones of the inner to fixed ones.
Or perhaps an internal kdoc can help.

timezones/full/build.gradle.kts
```diff
@@ -71,23 +182,109 @@ kotlin {
...
+        commonTest {
             dependencies {
                 runtimeOnly(project(":kotlinx-datetime"))
                 implementation(kotlin("test"))
             }
         }
 
-        val wasmWasiMain by getting {
+        jvmMain {
+            resources.srcDir(copiedTzdbDirectory)
```
ilya-g: Do we particularly need to copy these files?
Can we use instead a task that populates tzdbDirectory itself?

core/common/src/internal/MyJvmDefaultWithoutCompatibility.kt
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright 2019-2026 JetBrains s.r.o. and contributors.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.datetime
+
+@OptIn(ExperimentalMultiplatform::class)
+@OptionalExpectation
+internal expect annotation class MyJvmDefaultWithoutCompatibility()
```
ilya-g, line 10: I'd rather put it in `internal` directory so it doesn't get in sight when working with public API files.

core/common/test/samples/format/DateTimeComponentsSamples.kt
```diff
@@ -6,6 +6,7 @@
 package kotlinx.datetime.test.samples.format
 
 import kotlinx.datetime.*
+import kotlinx.datetime.TimeZoneContext
```
ilya-g: Looks redundant after a star import (in other sample files as well)

timezones/full/build.gradle.kts
dkhalanskyjb: I couldn't figure out how to do this without copying without also creating a mess in the jar file—if you use the `tzdbDirectory` as a source of JVM resources, its *contents* get included—and we get `GMT+0` right in the jar file's root. We could store `timezones/full/tzdb/tzdb` on the filesystem to work around that—but that's awkward, too.

If there's a way to have both a reasonable repository file hierarchy and reasonable jar file contents, it would of course be nice to avoid copying.

README.md
```diff
@@ -497,7 +498,57 @@ dependencies {
...
+To access it, add this snippet:
+
+```kotlin
+kotlin {
+    sourceSets {
+        commonMain {
+            dependencies {
+                // 2026b is the most recent release of the IANA timezone database,
+                // 0.8.0 is `kotlinx-datetime` version
+                implementation("org.jetbrains.kotlinx:kotlinx-datetime-zoneinfo:2026b-spi.0.8.0")
```
whyoleg:
```suggestion
                implementation("org.jetbrains.kotlinx:kotlinx-datetime-zoneinfo:2026c-spi.0.8.0")
```

As in the example for the wasm-wasi target. ([latest version source](https://www.iana.org/time-zones))

timezones/full/build.gradle.kts
ilya-g: I have not tried, but since `jvmProcessResources` (https://docs.gradle.org/current/javadoc/org/gradle/language/jvm/tasks/ProcessResources.html) task inherits from `Copy` task, you can try to configure additional resources and their destination layout in that task by using `into` and `from` [CopySpec](https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.file/-copy-spec/index.html) DSL, e.g. 
```
into("tzdb_subpath_in_jar") {
   from(tzdbDirectoryOrTaskProducingDirectory)
}
```

dkhalanskyjb: Thanks, this does work!

timezones/full/common/src/BundledTimeZoneContext.kt
```diff
@@ -0,0 +1,106 @@
...
+     * This is optimized shorthand for `try { get(id) } catch (_: IllegalTimeZoneException) { null }`.
+     * Please see the documentation for [get].
+     *
+     * It is guaranteed that this function never throws an exception for [BundledTimeZoneContext].
+     */
+    override fun getOrNull(id: String): TimeZone? = impl.getOrNull(id)
+
+    /**
+     * Returns the set of identifiers of the time zones whose data is bundled in this timezone database
+     * in addition to the fixed-offset time zones.
```
ilya-g: With this wording it's not quite clear whether fixed-offset time zones are included or not included in this set.

timezones/full/common/test/BundledTimeZoneContextTest.kt
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    private fun LocalDateTime(year: Int, month: Int, day: Int) = LocalDateTime(year, month, day, 0, 0)
+
+}
+
+/**
+ * [gapStart] is the first non-existent moment.
+ */
+private fun checkGap(timeZone: TimeZone, gapStart: LocalDateTime) {
```
ilya-g: Can we reuse now these helpers from test-utils?

core/windows/test/TimeZoneRulesCompleteTest.kt
```diff
@@ -32,7 +30,7 @@ class TimeZoneRulesCompleteTest {
 
             dtzi.DynamicDaylightTimeDisabled = 0u
             val (_, tzWithDst) = tzdb.currentSystemDefaultFromDtzi(dtzi)
-            assertTrue(tzWithDst is RegionTimeZone, "Expected RegionTimeZone, got ${tzWithDst::class}")
+            assertTrue(tzWithDst is RuleBasedTimeZone, "Expected RegionTimeZone, got ${tzWithDst::class}")
```
ilya-g: Minor: makes sense to change the assertion message as well

core/common/src/internal/tz/FixedOffsetTimeZoneDatabase.kt
```diff
@@ -0,0 +1,75 @@
...
+ * Copyright 2019-2025 JetBrains s.r.o. and contributors.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.datetime.internal
+
+import kotlinx.datetime.*
+import kotlinx.datetime.format.*
+
+internal class TimeZoneDatabaseWrapperWithFixedOffsetTimeZones(val inner: TimeZoneDatabase): TimeZoneDatabase {
```
ilya-g, line 11: Would you find it helpful for debugging to propagate toString of the inner database implementation through toString of this and other wrappers?

core/common/test/TimeZoneTest.kt
```diff
@@ -121,13 +122,17 @@ class TimeZoneTest {
...
         }
     }
 
     @Test
     fun utcOffsetNormalization() {
-        val sameOffsetTZs = listOf("+04", "+04:00", "UTC+4", "UT+04", "GMT+04:00:00").map { TimeZone.of(it) }
+        val sameOffsetTZs = listOf("+04", "+04:00", "UTC+4", "UT+04", "GMT+04:00:00").map {
+            TimeZoneContext.System.get(
+                it
+            )
```
ilya-g: Minor, formatting: can it be collapsed to single line?

core/darwin/src/Converters.kt
```diff
@@ -76,7 +77,17 @@ public fun TimeZone.toNSTimeZone(): NSTimeZone = if (this is FixedOffsetTimeZone
...
  */
-public fun NSTimeZone.toKotlinTimeZone(): TimeZone = TimeZone.of(name)
+@Deprecated(
+    "Use the overload that specifies the timezone database to use.",
+    ReplaceWith("toKotlinTimeZone(TimeZoneContext.System)"),
+)
+public fun NSTimeZone.toKotlinTimeZone(): TimeZone = toKotlinTimeZone(TimeZoneContext.System)
+
+/**
+ * Converts the [NSTimeZone] to the corresponding [TimeZone].
```
ilya-g: Makes sense to clarify "by looking a time zone with the same name in the provided [timeZoneDatabase]"

timezones/full/common/src/BundledTimeZoneContext.kt
dkhalanskyjb: I'm not sure about the good precise way to formulate this in the current implementation, so I pushed a version I'm not entirely happy with.

In my opinion, this point will become much clearer during the timezone hierarchy restructuring. I believe `FixedOffsetTimeZone` should *only* be returned for the predefined identifiers like `GMT+01:03:15` and not for any timezone whose rules merely happened not to contain transitions—then, the precise statement will be: `This set does not contain the identifiers of [fixed-offset time zones][FixedOffsetTimeZone]`. There's no need to settle this question right here, just bringing in some context.

timezones/full/common/test/BundledTimeZoneContextTest.kt
dkhalanskyjb: Oh, nice, I forgot I also copy-pasted those here.

core/common/src/internal/tz/FixedOffsetTimeZoneDatabase.kt
dkhalanskyjb: Can't hurt.

README.md
```diff
@@ -497,7 +498,57 @@ dependencies {
...
+The `kotlinx-datetime-zoneinfo` artifact bundles the up-to-date version of the timezone database
+and is updated soon after the official announcement from [IANA](https://www.iana.org/time-zones).
+To access it, add this snippet:
+
+```kotlin
+kotlin {
+    sourceSets {
+        commonMain {
+            dependencies {
+                // 2026b is the most recent release of the IANA timezone database,
```
ilya-g: 2026c

core/common/src/internal/tz/FixedOffsetTimeZoneDatabase.kt
ilya-g: I mean so that it eventually gets into toString of TimeZoneContext.System

dkhalanskyjb: Ah, ok. So far, it should be self-evident which timezone database is used, the wrappers aren't that deep. The idea to have `TimeZoneContext.System` in the `toString()` of `TimeZoneContext.System` is good, though, added that.

core/darwin/src/Converters.kt
```diff
@@ -76,7 +80,17 @@ public fun TimeZone.toNSTimeZone(): NSTimeZone = if (this is FixedOffsetTimeZone
...
```
ilya-g:
```suggestion
 * Converts the [NSTimeZone] to the corresponding [TimeZone] by looking a time zone with the same name
 * in the provided [timeZoneDatabase].
```

ilya-g (approved):
Please add this )

core/darwin/src/Converters.kt
dkhalanskyjb: Sorry, done!

ilya-g (approved):

core/jvm/src/Converters.kt
qwwdfsad: Thanks for an elaborate answer.
While there are no user-defined TZ/ZI subclasses, this one definitely makes sense, and there is not much we can do. Later, not so sure

core/common/src/TimeZoneContext.kt
qwwdfsad: Sure, but this one I'd expect to be frequently used/injected with for testing/DI purposes, where a glorified lambda comes in handy

timezones/full/common/src/BundledTimeZoneContext.kt
qwwdfsad: Sure. I didn't mean to be picky about OOM, I was rather thinking about incorrectly proguarded/trimmed resource loading, an error in parsing, anything in this area.
The suggestion would be to just remove this line, "This is optimized shorthand ..." should be enough

timezones/full/common/src/BundledTimeZoneContext.kt
qwwdfsad: >If your intuition tells you it's so obvious we don't even need measurements, I'm ready to listen, but so far, you haven't said that.

Sure, let me elaborate.

>Our general recommendation is to create TimeZone objects once and then keep a reference to them

People will ignore that in the same way they do with `Json` format (https://youtrack.jetbrains.com/issue/KTIJ-18677/). That's a very implicit requirement, and not complying with it usually doesn't punish you in any way.
Reading from disk every time sounds like a time bomb.

>We can't make it unbounded—that's too much data

I am not sure about the whole final cache, but something like reading the whole tzdb into a bytearray at once and caching it is a definitely nice gesture towards more responsive programs (`du -sh timezones/full/tzdb 2.4M` on my machine, a reasonable overhead). Parsing everything to `TimeZoneRulesCommon` -- not sure, worth measuring (also worth replacing `List<Long>` to `LongArray` though) -- probably reasonable to do as well.

I would go with both. The `ByteArray` part, I don't see a reason not to do, the second part -- up to you and the effort required. Feel free to do both in separate PRs to avoid postponing this one further

timezones/full/jvm/src/BundledTimeZoneContextJvm.kt
qwwdfsad: See my other comment.
Sure, not a blocker; not much reason to do a lot of extra disk readings when we can avoid that IMO

core/common/src/TimeZoneContext.kt
qwwdfsad: >The interface is intended to be fully externally implementable eventually

Sorry, I wasn't clear enough. It is extensible, and there are indeed a few use-cases: injecting changes into TZ rules, [white|black]listing of zone ids etc., but not much beyond that. What I wanted to convey is that we probably expect to provide capabilities that imply the instantiation of `TimeZone` -- to be able to do more than just wrap/filter available TZs. For example, to have a few TZDB versions in the same app, to have something similar to Java's `ZoneRulesProvider`, bundle your own, patched tzdb (trimmed or hot-swapped), have named fixed-offset zones etc.

All that likely implies changing `TimeZoneDatabase`, so while this topic ("custom rules/custom time zones") is not decided on, maybe it's better to prohibit this interface implementation

qwwdfsad (approved):
Absolute terrific work! Happy to see it's moving forward.

I have a few minor comments and follow up suggestions, nothing is a blocker, up to you to decide

timezones/full/common/src/BundledTimeZoneContext.kt
dkhalanskyjb: 2.4M is not the right number, your filesystem padded the files to 4K-sized blocks. `find tzdb/ -type f | xargs wc -c | sort -n` is ≈475K.

Still, that's 200x `America/New_York`. I'm wary of the first-access latency reading that much more data would introduce. It's unpleasant if an app reads more data from disk than necessary because of our lack of caching, but it's fixable on the size of the app. Having a hiccup each time any timezone is first accessed would be much worse, because you can do nothing about it if it needs to happen soon after startup.

My point is that without looking at real apps, we can't know whether we're improving or degrading the performance with the added cache. *Intuitively*, based on the usages I've seen, I'd say the solution is to cache only the data for the current system time zone in each timezone database, loading time zones piece by piece. Querying any other timezone is rare, and is mostly done in tests (to see how the code would behave for different values of the *current* one).

core/common/src/TimeZoneContext.kt
dkhalanskyjb: To me, a glorified lambda with that function doesn't look very readable:
```kotlin
printLog(networkError, request) { "Europe/Berlin" }
```

PR #626, opened 2026-05-27
Introduce `LocalDateTimeOffsetInfo` and `TransitionHandler`

### LINKED ISSUE #461: Add unambiguous LocalDateTime.toInstant

dkhalanskyjb:
Fixes #461


core/commonKotlin/src/TimeZone.kt
```diff
@@ -105,17 +112,22 @@ public actual open class TimeZone internal constructor() {
...
 
     internal open fun instantToLocalDateTime(instant: Instant): LocalDateTime = try {
         instant.toLocalDateTimeImpl(offsetAtImpl(instant))
     } catch (e: IllegalArgumentException) {
         throw DateTimeArithmeticException("Instant $instant is not representable as LocalDateTime.", e)
     }
 
+    internal open fun offsetAtImpl(instant: Instant): UtcOffset = error("Should be overridden")
+
+    internal open fun offsetInfoForImpl(dateTime: LocalDateTime): LocalDateTimeOffsetInfo = error("Should be overridden")
```
qwwdfsad, line 127: This method is not overridden in `SystemTimeZone`, and it will break all LD arithmetics when `js-joda/timezone` is not present.

It should at least be `IllegalTimeZoneException("js-joda timezone database is not available")`, though I am not sure because it appraently render the whole mode unusable

core/common/src/Instant.kt
```diff
@@ -63,25 +68,33 @@ public fun Instant.Companion.parse(
...
-        localDateTimeToInstant(ldtPlusDate, timeZone, preferred = initialOffset)
+        localDateTimeToInstantLenient(ldtPlusDate, timeZone, onTransition, preferred = initialOffset)
             .run { if (totalNanoseconds != 0L) plus(totalNanoseconds.nanoseconds).check(timeZone) else this }
     }.check(timeZone)
 } catch (e: ArithmeticException) {
-    throw DateTimeArithmeticException("Arithmetic overflow when adding CalendarPeriod to an Instant", e)
+    throw DateTimeArithmeticException("Arithmetic overflow when adding DateTimePeriod to an Instant", e)
 } catch (e: IllegalArgumentException) {
-    throw DateTimeArithmeticException("Boundaries of Instant exceeded when adding CalendarPeriod", e)
+    throw DateTimeArithmeticException("Boundaries of Instant exceeded when adding DateTimePeriod", e)
```
qwwdfsad: Now, if a handler throws AE or IAE, they will be overridden by DTAE with an irrelevant message

qwwdfsad: On a deeper look:
`LocalDateTime.toInstant` is explicit about `Exceptions thrown from [onTransition] are rethrown untouched.` and it seems to be the case.

`Instant.plus` doesn't state anything about exceptions, and the behaviour is different and a bit surprising:

```
val zone = TimeZone.of("Europe/Berlin")
val start = LocalDateTime(2025, 3, 29, 2, 30).toInstant(zone, TransitionHandler.RE
start.plus(DateTimePeriod(days = 1), zone, TransitionHandler.REJECT_TRANSITIONS)
```
throws `DateTimeArithmeticException: Boundaries of Instant exceeded when adding DateTimePeriod` + `On 2025-03-30T02:30, there is a transition: Gap(2025-03-30T01:00:00Z, offsetBefore=+01:00, offsetAfter=+02:00)` as a cause

core/commonKotlin/src/TimeZone.kt
qwwdfsad: ...not all, because some of it is being tested in `system()` and it passes, but in custom JS sources IDE is beyond being useful to properly pinpoint the root cause

core/commonKotlin/src/TimeZone.kt
```diff
@@ -214,6 +232,5 @@ private val lenientOffsetFormat = UtcOffsetFormat.build {
     }
 }
 
-internal actual fun localDateTimeToInstant(
-    dateTime: LocalDateTime, timeZone: TimeZone, preferred: UtcOffset?
-): Instant = timeZone.localDateTimeToInstant(dateTime, preferred)
+public actual fun TimeZone.offsetInfoFor(dateTime: LocalDateTime): LocalDateTimeOffsetInfo =
+    offsetInfoForImpl(dateTime)
```
qwwdfsad: Newline at EOF was accidentally removed?

core/jvm/src/TimeZoneJvm.kt
```diff
@@ -141,8 +155,16 @@ public actual fun LocalDateTime.toInstant(offset: UtcOffset, youShallNotPass: Ov
...
+    return if (transition == null) {
+        LocalDateTimeOffsetInfo.Regular(rules.getOffset(dateTime.value).let(::UtcOffset))
+    } else {
+        LocalDateTimeOffsetInfo.Transition(
+            transition.instant.toKotlinInstant(),
+            transition.offsetBefore.let(::UtcOffset),
+            transition.offsetAfter.let(::UtcOffset),
+        )
+    }
+}
```
qwwdfsad: Ditto EOF

core/common/src/TransitionHandler.kt
```diff
@@ -0,0 +1,189 @@
...
+ * }
+ * ```
+ *
+ * ## Usage
+ *
+ * Implementations of this interface typically should not be used directly.
+ * Instead, they can be passed as parameters to `kotlinx-datetime` functions such as
+ * [LocalDateTime.toInstant] or calendar-based arithmetic operations on [Instant].
+ */
+public interface TransitionHandler {
```
qwwdfsad, line 53: Probably won't hurt making it more `fun`

core/common/src/LocalDateTimeOffsetInfo.kt
```diff
@@ -0,0 +1,281 @@
...
+     * ```
+     * val transitionDescription = when (transition) {
+     *     is LocalDateTimeOffsetInfo.Gap -> {
+     *         "Gap occurred when transitioning " +
+     *             "from a lower UTC offset (${transition.offsetBefore}) " +
+     *             "to a bigger one (${transition.offsetAfter})."
+     *     }
+     *     is LocalDateTimeOffsetInfo.Overlap -> {
+     *         "Overlap occurred when transitioning " +
+     *             "from a bigger UTC offset (${transition.offsetAfter}) " +
```
qwwdfsad: Isn't it the opposite and `offsetBefore` is bigger?

core/common/src/Instant.kt
```diff
@@ -93,22 +106,28 @@ public fun Instant.plus(period: DateTimePeriod, timeZone: TimeZone): Instant = t
...
     }
 
 /**
  * Returns a [DateTimePeriod] representing the difference between `this` and [other] instants.
  *
- * The components of [DateTimePeriod] are calculated so that adding it to `this` instant results in the [other] instant.
+ * The components of [DateTimePeriod] are calculated so that adding it to `this` instant
+ * with the same [onTransition] handler results in the [other] instant.
+ *
+ * [onTransition] is invoked on several [LocalDateTime] values close to the [other] instant.
```
qwwdfsad: I am not exactly sure what it means semantically or when my handler should be invoked.

From the implementation, it seems like it's invoked on a few intermediate computations, but then the exception will be really hard to follow (because it won't match instants supplied).

Probably worth expanding documentation. Would be nice to change the exception message as well, but I am not sure it is even possible for custom handlers (and it's just unpleasant to fix for default ones)

core/common/src/TimeZone.kt
```diff
@@ -220,8 +220,50 @@ public expect open class TimeZone {
      * @sample kotlinx.datetime.test.samples.TimeZoneSamples.toInstantWithTwoReceivers
      */
     @Suppress("DEPRECATION_ERROR")
+    @Deprecated(
+        "Explicitly pass a TransitionHandler to `toInstant` calls",
+        replaceWith = ReplaceWith("this.toInstant(TransitionHandler.USE_OFFSET_BEFORE)")
+    )
     public fun LocalDateTime.toInstant(youShallNotPass: OverloadMarker = OverloadMarker.INSTANCE): Instant
```
qwwdfsad, line 246: Might be too soon for a patch version?
I.e. even in [icpc-live](https://github.com/icpc/live-v3) it is used ~9 times

qwwdfsad: ...and replace-with there produces mis-indented code, will report separetely 😢

core/common/src/TransitionHandler.kt
dkhalanskyjb: It would prevent us from adding, say, `resolveDateTimeOrNull` later.

core/common/src/TimeZone.kt
dkhalanskyjb: Could you please clarify what you mean? This isn't intended to land in a `0.8.1`, it's part of `0.9.0`.

qwwdfsad: I implicitly assumed it's going to be part of 0.8.1. 
If it's 0.9.0,  this comment is irrelevant (modulo issue I need to report to KTIJ)

core/common/src/Instant.kt
dkhalanskyjb: Which exact datetimes are queried is an implementation detail. It may change if we find it necessary to fix https://github.com/Kotlin/kotlinx-datetime/issues/484#issuecomment-4460157697, for example. For now, we can only promise that we query datetimes not farther than ±2 days from the `other` instant.

I'm curious why you consider it beneficial to document this behavior in more detail. Which situations do you envision where knowing exactly what our implementation does can meaningfully influence how/whether this function gets used? I'd rather not overload the documentation with meaningless precision, especially when sharing those extra details may complicate the maintenance.

qwwdfsad: I was rather surprised when stumbled across this line as our docs are typically quite precise -- I was puzzled enough to look into the implementation (though I cannot say it was necessary).

I'd say something along the lines "[onTransition] is invoked on intermediate computations of date and time components required to compute the period, subject to the implementation" is of the same verbosity but bit less confusing.

Treat this comment as purely optional and feel free to resolve as you see fit.

core/common/src/Instant.kt
```diff
@@ -52,6 +52,11 @@ public fun Instant.Companion.parse(
  * Returns an instant that is the result of adding components of [DateTimePeriod] to this instant. The components are
  * added in the order from the largest units to the smallest, i.e., from years to nanoseconds.
  *
+ * If, after adding the date portion of the [period], the resulting [LocalDateTime]
+ * corresponds to a [time transition][LocalDateTimeOffsetInfo.Transition], [onTransition] is invoked.
+ * The `preferredOffset` passed to [TransitionHandler.resolveDateTime] is the result of calling
+ * [offsetAt] with this [Instant] if it is valid for that datetime.
```
ilya-g:
```suggestion
 * If, after adding the date portion of the [period], the intermediate [LocalDateTime]
 * corresponds to a [time transition][LocalDateTimeOffsetInfo.Transition], [onTransition] is invoked.
 * The `preferredOffset` passed to [TransitionHandler.resolveDateTime] is the result of calling
 * [offsetAt] with this [Instant] if it is valid for the intermediate datetime.
```
I propose slightly different wording here and below to emphasize that the datetime passed to the resolver is not the resulting datetime returned from the operation.

core/common/src/TimeZone.kt
```diff
@@ -220,8 +220,50 @@ public expect open class TimeZone {
...
+     *   [utcOffset] is passed to [TransitionHandler.resolveDateTime] as the `preferredOffset`.
+     *   A non-`null` [utcOffset] must be equal to [LocalDateTimeOffsetInfo.Transition.offsetBefore]
+     *   or [LocalDateTimeOffsetInfo.Transition.offsetAfter], or an [IllegalArgumentException] will be thrown.
+     *   Exceptions thrown from [onTransition] are rethrown untouched.
+     *
+     * @see Instant.toLocalDateTime
+     * @sample kotlinx.datetime.test.samples.TimeZoneSamples.toInstantWithTwoReceivers
+     */
+    // Added after 0.8.0
+    public fun LocalDateTime.toInstant(onTransition: TransitionHandler, utcOffset: UtcOffset? = null): Instant
```
ilya-g, line 284: In the perspective of introducing overloads with context parameters, we may want to introduce these new methods as taking TimeZone as a context (without providing extension members)

core/common/src/TransitionHandler.kt
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Calculate the [Instant] corresponding to the given [dateTime].
```
ilya-g: Calculates

core/common/src/TimeZone.kt
dkhalanskyjb: Yes, agreed. I wanted to add all context parameters in bulk in a separate PR (and deprecate this set of overloads as well).

core/common/test/TimeZoneTest.kt
```diff
@@ -257,64 +258,44 @@ class TimeZoneTest {
...
-        // the offsets after the overlap are equal
-        assertEquals(
-            instantStart.offsetIn(timeZone),
-            instantEnd.offsetIn(timeZone)
-        )
-    } catch (e: Throwable) {
-        throw Exception("Didn't find an overlap at $overlapStart for $timeZone", e)
-    }
+    val after = assertIs<LocalDateTimeOffsetInfo.Regular>(timeZone.offsetInfoFor(overlapStart))
+    val overlap = assertIs<LocalDateTimeOffsetInfo.Overlap>(
```
ilya-g: Can we also check offsetBefore and offsetAfter properties of the overlap? For example, by comparing offsetAfter with the after.offset

integration-testing/js-without-timezones/common/test/TimezonesWithoutDatabaseTest.kt
```diff
@@ -97,3 +115,54 @@ class TimezonesWithoutDatabaseTest {
     }
 
 }
+
+
+/**
+ * [gapStart] is the first non-existent moment.
+ */
+@OptIn(ExperimentalTime::class)
+private fun checkGap(timeZone: TimeZone, gapStart: LocalDateTime) {
```
ilya-g: Can we reuse these helper functions from TimeZoneTest.kt?

core/common/src/TimeZone.kt
ilya-g: Ok, so we could remove them in that PR if they are not released yet?

core/common/test/ReadmeTest.kt
```diff
@@ -33,7 +33,12 @@ class ReadmeTest {
 
         val kotlinReleaseDateTime = LocalDateTime(2016, 2, 15, 16, 57, 0, 0)
 
-        val kotlinReleaseInstant = kotlinReleaseDateTime.toInstant(TimeZone.of("UTC+3"))
+        val kotlinReleaseInstant1 = kotlinReleaseDateTime.toInstant(
+            TimeZone.of("Europe/Moscow"),
+            TransitionHandler.REJECT_TRANSITIONS
+        )
+        val kotlinReleaseInstant2 = kotlinReleaseDateTime.toInstant(UtcOffset(hours = 3).asTimeZone())
+        assertEquals(kotlinReleaseInstant1,kotlinReleaseInstant2)
```
ilya-g: Minor: space

core/common/src/Instant.kt
```diff
@@ -142,6 +180,11 @@ public fun Instant.periodUntil(other: Instant, timeZone: TimeZone): DateTimePeri
  * Returns the whole number of the specified date or time [units][unit] between `this` and [other] instants
  * in the specified [timeZone].
  *
+ * The return value is calculated so that adding it to `this` instant
+ * with the same [unit] and the same [onTransition] handler results in the [other] instant.
+ *
+ * [onTransition] is invoked on a [LocalDateTime] value close to the [other] instant.
```
ilya-g: Currently, it makes an impression that onTransition is always invoked. 
Maybe we can say that it "can be invoked to account for possible transitions on the date of the [other] instant."

core/common/src/Instant.kt
```diff
@@ -447,6 +551,11 @@ public fun Instant.minus(value: Long, unit: DateTimeUnit.TimeBased): Instant =
  * Returns the whole number of the specified date or time [units][unit] between [other] and `this` instants
  * in the specified [timeZone].
  *
+ * The return value is calculated so that adding it to [other] instant
+ * with the same [unit] and the same [onTransition] handler results in `this` instant.
+ *
+ * [onTransition] is invoked on a [LocalDateTime] value close to the [other] instant.
```
ilya-g: Same note as in `until`, but since the arguments are swapped, "on the date of `this` instant"

core/common/src/TransitionHandler.kt
```diff
@@ -0,0 +1,189 @@
...
+     * the ambiguous time-of-day `02:31` will be resolved to the first occurrence.
+     *
+     * On [gaps][LocalDateTimeOffsetInfo.Gap],
+     * it finds the first moment that does not fall into the gap.
+     * For example, if the clocks were moved from `02:00` to `03:00` on a given day,
+     * the non-existent time-of-day `02:31` will be resolved to the moment corresponding to `03:00`.
+     *
+     * Note that the preferred UTC offset is always ignored.
+     */
+    public object FIND_EARLIEST_VALID_TIME : TransitionHandler {
```
ilya-g: I'm not sure if this casing fits for object names. Maybe we should make them companion fields or properties then?

ilya-g (changes requested):
No more suggestions beside these

core/common/src/TimeZone.kt
dkhalanskyjb: Yes, I've been sticking `Added after 0.8.0` to make it easier to remember which overloads haven't been published yet and can be removed.

integration-testing/js-without-timezones/common/test/TimezonesWithoutDatabaseTest.kt
ilya-g: Oh, I see now that they were in different projects, so reusing them is not so straightforward and requires a separate project to enable sharing. 
So now, I'm not insisting on that anymore. But if you see that it can be useful in future, let's have it.

core/common/src/Instant.kt
ilya-g: Even for date-based units it is not always invoked.

dkhalanskyjb: Ah, ok, got your point.

integration-testing/js-without-timezones/common/test/TimezonesWithoutDatabaseTest.kt
dkhalanskyjb: Yeah, I think we'll get some use out of it. In the past, the lack of `test-utils` did lead me to avoid writing a couple of integration tests that would have been nice to have.

ilya-g (approved):

PR #630, opened 2026-07-07
Allow newer version of `@js-joda/core` than 3.2.0

### LINKED ISSUE #627: `@js-joda/timezone` packages with version `>2.23.0` require `@js-joda/core` version `>=5.7.0`
I'm opening this issue mostly to raise awareness to the fact that the latest `@js-joda/timezone` version currently supported by `kotlinx-datetime` seems to be version `2.23.0`.

From version `2.24.0` onwards, the package's peer dependency has changed to `"@js-joda/core": ">=5.7.0"`, while `kotlinx-datetime` is still using version `3.2.0` of `@js-joda/core`.

The [README](https://github.com/Kotlin/kotlinx-datetime#note-about-time-zones-in-js) currently directs users to install the latest available version of `@js-joda/timezone`. This should either be changed so that users are instructed to install version `2.23.0` exactly or, probably better, `kotlinx-datetime` should update its `@js-joda/core` dependency to a more recent one so that users don't have to depend on possibly outdated timezone data.

dkhalanskyjb:
Fixes #627

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -199,7 +199,7 @@ kotlin {
         val commonJsMain by getting {
             dependencies {
                 api("org.jetbrains.kotlinx:kotlinx-serialization-core:$serializationVersion")
-                implementation(npm("@js-joda/core", ">= 3.2.0"))
+                implementation(npm("@js-joda/core", ">= 3.2.0 < 7.0.0"))
             }
         }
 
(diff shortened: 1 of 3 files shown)


YarnSphere:
Note that just doing `>=3.2.0` is a bit dangerous, if `@js-joda/core` releases a v7 with breaking changes that aren't compatible with `kotlinx-datetime`, that would still be the version installed by default since it is the latest available.

It is typically a better practice to depend on known compatible versions, so: `>=3.2.0 <7.0.0` or similar, assuming that `kotlinx-datetime` has been tested with versions 4, 5, and 6.

dkhalanskyjb:
Good idea, thank you!

ilya-g (approved):

PR #633, opened 2026-07-13
Reflect in README that compatibility versions will be published for 0.8.x versions

vsartiushkin:
Reflect in README that compatibility versions related to the deprecation of `Instant` will be published for `0.8.x` versions of the library as well, as version `0.8.0-0.6.x-compat` is already available.


dkhalanskyjb:
Well done!

PR #634, opened 2026-07-16
Update to Gradle 9.6.1

dkhalanskyjb:


core/build.gradle.kts
```diff
@@ -196,26 +194,26 @@ kotlin {
...
 
-        val commonKotlinMain by getting {
+        getByName("commonKotlinMain") {
             dependencies {
                 api("org.jetbrains.kotlinx:kotlinx-serialization-core:$serializationVersion")
             }
         }
 
-        val wasmWasiTest by getting {
+        getByName("wasmWasiTest") {
```
fzhinkin: For brevity, it can be `named("...")`.

core/build.gradle.kts
```diff
@@ -22,12 +22,10 @@ mavenPublicationsPom {
     description.set("Kotlin Datetime Library")
 }
 
-val mainJavaToolchainVersion: String by project
-val modularJavaToolchainVersion: String by project
-val serializationVersion: String by project
+val serializationVersion = project.property("serializationVersion")
```
fzhinkin, line 25: Just curious: why all these changes were required to update the Gradle version?

dkhalanskyjb: The by-delegation is deprecated since 9.5 and emits warnings: https://docs.gradle.org/9.6.0/userguide/upgrading_version_9.html#kotlin_dsl_delegated_properties I upgraded Gradle, saw new warnings, and fixed them as part of the same PR.

fzhinkin: Got you, thanks!

fzhinkin (approved):

core/build.gradle.kts
dkhalanskyjb: Alright. The behavior changes a bit, but it seems okay in this context.

fzhinkin:
^ I reran CI using setting from the branch and the build failed with (and I have the same issue locally when using JDK 17 as a default java version):
```
> Task :kotlinx-datetime-test-with-jpms:compileJava FAILED
/Users/filipp.zhinkin/Development/kotlinx-datetime/integration-testing/jpms-test/src/main/java/my/jpms/test/InitLocalDate.java:11: error: cannot access LocalDate
    LocalDate newLocalDate() {
    ^
  bad class file: /Users/filipp.zhinkin/Development/kotlinx-datetime/core/build/libs/kotlinx-datetime-jvm-0.8.0-SNAPSHOT.jar(/kotlinx/datetime/LocalDate.class)
    class file has wrong version 61.0, should be 55.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.
/Users/filipp.zhinkin/Development/kotlinx-datetime/integration-testing/jpms-test/src/main/java/my/jpms/test/InitLocalDate.java:12: error: cannot find symbol
        return new LocalDate(2025, 3, 31);
                   ^
  symbol:   class LocalDate
  location: class InitLocalDate
2 errors
```

It seems like the build script is missing explicitly specified target/release 8.

dkhalanskyjb:
I fixed the problem differently, by reverting the Java Toolchain choice back to 1.8 for normal compilation and 11 for module-info. This did the trick: the major versions of the classfiles are all 52.

core/build.gradle.kts
```diff
@@ -22,12 +22,10 @@ mavenPublicationsPom {
...
 
 java {
-    toolchain { languageVersion.set(JavaLanguageVersion.of(mainJavaToolchainVersion)) }
+    toolchain { languageVersion.set(JavaLanguageVersion.of(project.findProperty("mainJavaToolchainVersion") as String)) }
```
ilya-g: `findProperty` returns null for the missing property, which would although result in error when cast to String, but it would be harder to diagnose the problem from this error. For properties that are expected to be present it's better to use `property` function.

fzhinkin (approved):

PR #636, opened 2026-07-23
KT-78078: remove watchosArm32

fzhinkin:
Similar to what we did in other kotlinx-libraries, let's remove the watchosArm32 target only from the kc/dev branch.


dkhalanskyjb (approved):
Sure thing, but do first-party commits to `kotlin-community/dev` even require a review?

fzhinkin:
I figured it’s a nice way to let you know that there will be no such target in this branch anymore :)

PR #638, opened 2026-08-06
Add context-parameter-based versions of the existing functions

dkhalanskyjb:


core/commonKotlin/src/TimeZone.kt
```diff
@@ -85,19 +85,22 @@ public actual open class TimeZone internal constructor() {
     public actual open val id: String
         get() = error("Should be overridden")
 
+    @Deprecated(
```
ilya-g, line 54: Maybe we can deprecate it straight with HIDDEN? As `with(timeZone) { }` should continue to work with context parameter overload.

core/common/src/TimeZone.kt
```diff
@@ -347,10 +312,29 @@ public typealias ZoneOffset = FixedOffsetTimeZone
...
+ * **Pitfall**: the offset returned from this function should typically not be used for datetime arithmetics
+ * because the offset can change over time due to daylight-saving-time transitions and other reasons.
+ * Use [TimeZone] directly with arithmetic operations instead.
+ *
+ * @see Instant.toLocalDateTime
+ * @see TimeZone.offsetAt
+ * @see offsetIn
+ * @sample kotlinx.datetime.test.samples.TimeZoneSamples.offsetWithContextParameter
+ */
+context (timeZone: TimeZone)
```
ilya-g: Minor: extra space, here and in other context declarations

core/common/test/samples/TimeZoneSamples.kt
```diff
@@ -61,9 +61,9 @@ class TimeZoneSamples {
         }
         fun logEntry(message: String, now: Instant = Clock.System.now()): String {
             val formattedTime = logTimeFormat.format {
-                with(TimeZone.currentSystemDefault()) {
+                context(TimeZone.currentSystemDefault()) {
                     setDateTime(now.toLocalDateTime())
-                    setOffset(offsetAt(now))
+                    setOffset(contextOf<TimeZone>().offsetAt(now))
```
ilya-g: Maybe `now.offset()`? Or it is important to show how to get a timezone back from context?

PR #647, opened 2026-09-01
Remove usages of deprecated macosX64, tvosX64, watchosX64 targets

Egorand:
These targets are deprecated in Kotlin 2.5.0, see KT-84955.


fzhinkin:
The error will be suppressed, the corresponding change will be merged after a CI's run completion.

PR #649, opened 2026-09-03
Fix byUnicodePattern dropping an escaped quote inside a literal

lenamonj:
`byUnicodePattern` drops the escaped quote inside a quoted literal. The Unicode pattern grammar spells an apostrophe inside a literal as `''`, and java.time formats `'o''clock'` as `o'clock`; here it formats as `oclock` and only parses that spelling. The parser walked the pattern by character, so on a `'` inside a literal it could not see whether the next character was another `'`.

This walks the pattern by index and treats `''` inside a literal as one apostrophe; a lone `'` still closes the literal. Adds a test over six pattern shapes.

Found by an automated audit; the fix and the test were reviewed and verified by a person on a fresh clone with jvmTest, jsTest and wasmJsTest.

### DIFF
--- a/core/common/test/format/DateTimeFormatTest.kt
+++ b/core/common/test/format/DateTimeFormatTest.kt
@@ -114,6 +114,24 @@ class DateTimeFormatTest {
         }
     }
 
+    @OptIn(FormatStringsInDatetimeFormats::class)
+    @Test
+    fun testEscapedQuoteInUnicodePatternLiteral() {
+        val time = LocalTime(23, 53)
+        for ((pattern, expected) in listOf(
+            "'o''clock' HH:mm" to "o'clock 23:53",
+            "'''x' HH:mm" to "'x 23:53",
+            "'x''' HH:mm" to "x' 23:53",
+            "'a''''b' HH:mm" to "a''b 23:53",
+            "'' HH:mm" to "' 23:53",
+            "'''' HH:mm" to "' 23:53",
+        )) {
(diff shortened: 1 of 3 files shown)


core/common/test/format/DateTimeFormatTest.kt
```diff
@@ -114,6 +114,24 @@ class DateTimeFormatTest {
...
+            "'a''''b' HH:mm" to "a''b 23:53",
+            "'' HH:mm" to "' 23:53",
+            "'''' HH:mm" to "' 23:53",
+        )) {
+            val format = LocalTime.Format { byUnicodePattern(pattern) }
+            assertEquals(expected, format.format(time), pattern)
+            assertEquals(time, format.parse(expected), pattern)
+        }
+    }
+
```
dkhalanskyjb: The test location isn't right. At the very least, it's not `DateTimeFormatTest`, as there are no *dates* involved. In my opinion, the best place to put this is `UnicodeFormatTest`. It's JVM-only, which reduces the coverage, but in exchange, we can write tests confirming that Java does exhibit the same behavior as our implementation.

dkhalanskyjb (changes requested):
The fix looks correct, but the tests can be improved.

lenamonj:
Moved it to UnicodeFormatTest as a checkPattern case list, so each pattern is formatted and parsed by both this implementation and java.time and the strings compared. Removed from DateTimeFormatTest. jvmTest and apiCheck green on a fresh clone; the o''clock case fails on master without the fix.

PR #653, opened 2026-09-14
Use IANA tzdb 2026d

dkhalanskyjb:
https://lists.iana.org/hyperkitty/list/tz-announce@iana.org/thread/L4IGHJRBUQR53F7RTEHM5IS7ZMBR3YXP/


fzhinkin:
@dkhalanskyjb, do you plan to wait until the next regular release to release 2026d tzdb, or you will cherry pick changes and release 2026d-0.8.0 spi?

fzhinkin (approved):

dkhalanskyjb:
Since we're already maintaining our own timezone database for the Wasm/WASI version of `kotlinx-datetime`, we have to publish the 2026d release in the next day or two, and the next full release is not going to be ready by then. That means cherry picking.

PR #654, opened 2026-09-16
Move some sources to more appropriate source sets

demiurg906:
These sources referenced some declarations from more specific platform dependency comparing to the source set they were declared, which is incompatible with (new) separate compilation scheme.


timezones/full/wasmWasi/src/TimeZonesInitializer.kt
ilya-g, line 1: It seems that this has already been done in master, does it make sense to rebase kotlin-community/dev on that?

fzhinkin: One can not simply rebase it to master (there's an issue with JPMS we were discussing with `@dkhalanskyjb` elsewhere). And there's still an issue with `windows`.

